PR #15, opened 2016-01-21
Fixed agent config and logging

yujunglo:
@sbaogang @jinruh

### DIFF
--- a/deploy-agent/.bumpversion.cfg
+++ b/deploy-agent/.bumpversion.cfg
@@ -2,5 +2,5 @@
 files = setup.py
 commit = True
 tag = True
-current_version = 1.2.1
+current_version = 1.2.2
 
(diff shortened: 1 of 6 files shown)


mansu:
Not to pick on this specific diff but can we please add more detail to the diff on what the problem being solved is and how this diff fixes it? I think this is a good practice for a few reasons:

a) This is good code development and git hygine. That way looking at the history it would be clear what has changed and why we changed it.
b) A random observer like me would have no context on weather this change is important when I look at the history or the code.
c) Now that we are open source, including this info would also inform external reviewers why this change was made.

sbaogang:
Per discussion, we will treat this as an exception - once we open sourced, we will follow a more strict process.

yujunglo:
This is an urgent fix. I'll provide more context next time.

PR #16, opened 2016-01-22
Remove the PENDING_TERMINATE host state when health check completed.

yujunglo:
Since we have HealthCheckHostTerminator to keep track of health check instances now,  we don't need to set the PENDING_TERMINATE host state (which is used to clean up instances before). And also adjust the time to retain instances before it is terminated.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HealthChecker.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HealthChecker.java
@@ -440,20 +440,18 @@ public class HealthChecker implements Runnable {
             LOG.info("The host {} has been terminated", hostId);
             healthCheckBean.setHost_terminated(true);
         } else {
-            HostBean hostBean = new HostBean();
             if (healthCheckBean.getStatus() == HealthCheckStatus.FAILED || healthCheckBean.getStatus() == HealthCheckStatus.TIMEOUT) {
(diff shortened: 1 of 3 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HealthChecker.java
```diff
@@ -440,20 +440,18 @@ private void processCompletingState(HealthCheckBean healthCheckBean, GroupBean g
...
-                hostBean.setState(HostState.PENDING_TERMINATE);
-                hostBean.setLast_update(System.currentTimeMillis());
                 healthCheckBean.setHost_terminated(false);
             } else {
                 LOG.info(String.format("Health check %s. Termimate the host %s", healthCheckBean.getStatus().toString(), hostId));
                 hostInfoDAO.terminateHost(hostId);
+                HostBean hostBean = new HostBean();
                 hostBean.setState(HostState.TERMINATING);
                 hostBean.setLast_update(System.currentTimeMillis());
+                hostDAO.updateHostById(hostId, hostBean);
```
jinruh: What I am thinking is, we can also move this logic (line 448 to 453) to the HealthCheckHostTerminator.java 
The extra work we need to add in the HealthCheckHostTerminator.java, is for each HealthCheckBean, we check the status:

if (healthCheckBean.getStatus() == HealthCheckStatus.SUCCEEDING || System.currentTimeMillis() - bean.getHost_launch_time() > timeToRetain) {
 // terminate instance.
}

PR #49, opened 2016-02-12
Add percentage based auto-scaling

jinruh:
As discussed in the task:
Percentage based auto-scaling. Right now I (afaik) can only scale up and down using a set number of instances. That turns out to be problematic when the min number of instances is very low (we have some clusters where we do no work at all for periods of time), since the impact of the scale up/down actions ends varying quite a lot depending on how far the cluster has already scaled. Having the ability to do percentage based scaling would help with this.

@sbaogang @yujunglo


jinruh:
screen shot 2016-02-12 at 12 28 45 am

yujunglo:
looks good. thanks

PR #65, opened 2016-02-16
Removed authoriozation check on publish_build and recorded publisher

nickdechant:
-Removed authorization call for publish_builds
-Made publish_info record who published the build instead


sbaogang:
r+

PR #70, opened 2016-02-17
Send maximum 20 records per request

jinruh:
Aws cloud watch can only take up to 20 data points in each request. Send at most 20 in each batch.

@sbaogang @yujunglo


yujunglo:
looks good. thanks.

PR #73, opened 2016-02-18
[pinterest-only] only update deploy priority once in post deploy

jinruh:
this is for pinterest internal ngapp2 deploy only, we use finish_message_flag to check if we have already updated deploy priority or not.

@sbaogang @yujunglo


yujunglo:
looks good to me. thanks

PR #78, opened 2016-03-02
Add sensitivity ratio parameter to the spot auto scaling group

jinruh:
Add a new parameter to the spot auto scaling group. 
@sbaogang


jinruh:
screen shot 2016-03-02 at 12 07 51 pm

PR #79, opened 2016-03-03
Add puppet_file_path configuration

gengmao:
A small change to remove the hardcoded path from source code and make it configurable.

### DIFF
--- a/deploy-agent/deployd/common/utils.py
+++ b/deploy-agent/deployd/common/utils.py
@@ -99,7 +99,10 @@ def run_prereqs(config):
     # check if the puppet has finished or not
     if IS_PINTEREST:
         respect_puppet = config.respect_puppet()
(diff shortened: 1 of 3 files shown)


deploy-agent/deployd/common/utils.py
```diff
@@ -99,8 +99,8 @@ def run_prereqs(config):
     # check if the puppet has finished or not
     if IS_PINTEREST:
         respect_puppet = config.respect_puppet()
-        if respect_puppet and not os.path.exists("/var/lib/puppet/state/state.yaml"):
-            log.info("Waiting for first successful puppet run.")
+        if respect_puppet and not os.path.exists(config.get_puppet_file_path()):
```
jinruh: To make it more defensive. You probably want to check if config.get_puppet_file_path()

jinruh:
Please also bump the deploy agent version in setup.py

PR #83, opened 2016-03-10
File mirroring

nickdechant:
Added a function in Deploy-Agent stager.py that mirrors files in the teletraan/teletraan/mirror directory for a service.


sbaogang:
Sorry my bad - I think at this point we better do not check in this feature - mainly because this feature could be replaced by telefig templater and user customized copy or rsync themselves. Add more ways to accomplish the similar function is hard to maintain and confuse people. I think we should keep most of the features in telefig.

PR #89, opened 2016-03-21
fix slack mention and update cloud-init

sbaogang:
1. Added send message to user feature and implemented for Slack
2. Updated cloud-init to work with latest docker-ami

@nickdechant


nickdechant:
Looks great! r+

PR #90, opened 2016-03-22
Write config change log when enabling/disabling health check

jinruh:
as title. Also send message to the slack channel when failed to enable/disable health check.

@sbaogang


sbaogang:
r+

PR #91, opened 2016-03-22
[pinterest-only] use facter to read host information

yujunglo:
@sbaogang 
Allow to use facter to read host information.

### DIFF
--- a/deploy-agent/deployd/agent.py
+++ b/deploy-agent/deployd/agent.py
@@ -350,6 +350,8 @@ def main():
                         help="Group name being used when interact with Teletraan service. "
                              "This is optional. By default the group name defined in host-info "
                              "file will be used")
+    parser.add_argument('--use-facter', dest='use_facter', action='store_true', default=False)
+
     args = parser.parse_args()
     config = Config(args.config_file)
     utils.run_prereqs(config)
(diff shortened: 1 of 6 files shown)


deploy-agent/deployd/agent.py
```diff
@@ -350,6 +350,9 @@ def main():
                         help="Group name being used when interact with Teletraan service. "
                              "This is optional. By default the group name defined in host-info "
                              "file will be used")
+    parser.add_argument('-u', '--use-facter', dest='facter', required=False, default=False,
```
sbaogang: small change: dest='facter' => dest='use_facter', so naming is more consistent

deploy-agent/deployd/client/client.py
```diff
@@ -28,14 +29,33 @@
...
+            if not self._hostname and self._config.get_facter_name_key():
+                self._hostname = utils.get_info_from_facter(self._config.get_facter_name_key())
+
+            if not self._ip and self._config.get_facter_ip_key():
+                self._ip = utils.get_info_from_facter(self._config.get_facter_ip_key())
+
+            if not self._id and self._config.get_facter_id_key():
+                output = utils.get_info_from_facter(self._config.get_facter_id_key())
+                if output:
+                    self._id = output.split(",")
```
sbaogang: probably do not need to split

deploy-agent/deployd/client/client.py
```diff
@@ -28,14 +29,33 @@
...
+
+            if not self._hostgroup and self._config.get_facter_group_key():
+                output = utils.get_info_from_facter(self._config.get_facter_group_key())
+                if output:
+                    self._hostgroup = output.split(",")
+
+        # read host_info file
```
sbaogang: Let us make the host_info inside an else block - if we use factor, we should not parse host_info anymore

sbaogang:
r+

PR #92, opened 2016-03-22
Only check unused reserved instance for specific instance type

jinruh:
This is to make sure we won't hit AWS API rate limit
Also fixed some bugs in launch event collector. 

@sbaogang


yujunglo:
looks good. r+

PR #93, opened 2016-03-22
Fixed missing host details for capacity hosts and also fixed env name

yujunglo:
@sbaogang @jinruh 
(1) Fixed missing host details for capacity hosts.
(2) Fixed wrong env name in hosts details.


sbaogang:
r+

jinruh:
It would be nice to put this filtering on the client side to make the API more general.

yujunglo:
Updated w/ @jinruh comment.

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
```diff
@@ -209,6 +211,6 @@ public HostBean getByEnvIdAndHostId(String envId, String hostId) throws Exceptio
     @Override
     public HostBean getByEnvIdAndHostName(String envId, String hostName) throws Exception {
```
jinruh: Do we still need this function call?

yujunglo: Yes, this is used in EnvironHandler.

PR #95, opened 2016-03-24
Make user data and other aws setting configurable

yujunglo:
@sbaogang 
1. Make user data and other aws settings configurable
2. Fix a bug when pushing change feed


sbaogang:
r+

PR #97, opened 2016-03-24
Make cluster url public and add image name options

yujunglo:
@sbaogang 
1. Make cluster config url public
2. Add image name options in config
screen shot 2016-03-24 at 3 56 27 pm


yujunglo:
update ui
screen shot 2016-03-24 at 4 46 24 pm

sbaogang:
r+

PR #99, opened 2016-03-25
Fix InvalidInstanceID.Malformed regex pattern bug

jinruh:
Should have made it match all characters

@sbaogang @yujunglo


sbaogang:
r+

yujunglo:
Thanks r+!!

PR #100, opened 2016-03-25
fix the path of test scripts for ngapp2

sbaogang:
as title
@jinruh @yongwen


jinruh:
Thanks for the fix!
r+

PR #103, opened 2016-03-25
Remove redundant lifecycle call

yujunglo:
@sbaogang @jinruh 
If the host has been removed, no need to do lifecycle complete action.


jinruh:
thanks for the quick fix! r+

PR #104, opened 2016-03-25
Make Ec2 instance Ids checks more defensive

jinruh:
As suggested by @sbaogang , do id check one by one would be safer.

@sbaogang @yujunglo

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/arcee/aws/EC2HostInfoDAOImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/arcee/aws/EC2HostInfoDAOImpl.java
@@ -26,7 +26,6 @@ import com.pinterest.arcee.dao.HostInfoDAO;
 import com.pinterest.deployservice.common.DeployInternalException;
 
 import com.pinterest.deployservice.bean.HostState;
-import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


deploy-service/common/src/main/java/com/pinterest/arcee/aws/EC2HostInfoDAOImpl.java
```diff
@@ -212,4 +187,9 @@ public void terminateHost(String hostId) throws Exception {
         }
         return instanceIds;
     }
+
+
+    public static void main(String[] args) {
```
jinruh: oops, will remove this.

deploy-service/common/src/main/java/com/pinterest/arcee/aws/EC2HostInfoDAOImpl.java
```diff
@@ -58,48 +57,35 @@ public EC2HostInfoDAOImpl(AmazonEC2Client client) {
...
-                LOG.debug("Cannot find the following ids in AWS:", ids);
-                terminatedHosts.addAll(ids);
-                return terminatedHosts;
+                }
             } catch (AmazonServiceException ex) {
-                Collection<String> invalidHostIds = handleInvalidInstanceId(ex);
-                ids.removeAll(invalidHostIds);
-                // add invalid host ids to the terminated host list.
-                terminatedHosts.addAll(invalidHostIds);
+                Collection<String> invalidIds = handleInvalidInstanceId(ex);
```
sbaogang: in handleInvalidInstanceId(), we still parse the id in the exception, could we just assume the id in the exception is the staleId here? The idea of handling one instance a time is so that we could get rid of the potentially fragile parsing code; otherwise, I would think describe instances in a batch makes more sense

yujunglo:
Looks good! thanks

PR #105, opened 2016-03-26
use different s3 bucket for ngapp integ and prod, fix rollback histor…

yongwen:
…y display with existing deploy.


yongwen:
close this one. will create a new one after resolving conflict.

PR #107, opened 2016-03-28
Prevent overwriting terminating host state

yujunglo:
@sbaogang @jinruh 
Prevent overwriting the terminating host state to ACTIVE when agent pings back.


yujunglo:
Updated with @jinruh's comment.

jinruh:
Thanks for the quick fix! r+

PR #108, opened 2016-03-29
Use cloudwatch alarm to enable/disable scale up/down activities

jinruh:
The spot auto scaling group would work as follows:
1. When there are enough reserved instance: disable scale up, enable scale down -> peak hours, scale up on demand group;  off-peak hours, scale down spot first, then on demand instance.
2. When the reserve instance inventory is low: disable scale down, enable scale up -> off-peak hours, scale down on demand first, then spot instance; peak hours: scale up spot instance first, then on demand instances.

@yujunglo


yujunglo:
looks good. r+!

PR #110, opened 2016-03-31
Remove UNREACHABLE host state

jinruh:
UNREACHABLE host state is redundant. If agent failed to ping server, setting agent state to UNREACHABLE would be enough. 

@sbaogang @yujunglo @nickdechant


yujunglo:
Could you also update this line to use hostTip instead of hostStateTip?
https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/templates/hosts/host_details.html#L63

And remove the hostStateTip
https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/webapp/templatetags/utils.py#L413

Thanks!

jinruh:
update PR per @yujunglo

yujunglo:
Looks good! Thanks!

PR #111, opened 2016-04-01
Use spot auto scaling name instead of cluster name for spot auto scaling

jinruh:
to make spot launch config more unique

@yujunglo


yujunglo:
looks good to me. r+!

PR #119, opened 2016-04-13
Don't send instance warning message if the instance has already

jinruh:
So we won't get false warning messages.

@yujunglo


yujunglo:
looks good. Thanks!

PR #120, opened 2016-04-13
fix date issue

yujunglo:
@jinruh


jinruh:
thanks for the quick fix!

PR #121, opened 2016-04-14
Fix some UI bugs for auto scaling landing page

jinruh:
screen shot 2016-04-14 at 12 58 44 pm

@sbaogang @yujunglo


yujunglo:
thanks for fixing!

PR #123, opened 2016-04-15
fix the s3 bucket name used for rollback history in prod

yongwen:


michaelfu:
Was the bug because that the s3 bucket name didn't match the actual bucket name in S3?

jinruh:
looks good.

jinruh:
from the stack trace, it's likely previously it's due to the bucket name mismatch.

yongwen:
@michaelfu the new deploy history code was looking for rollback history from s3 bucket with the prefix "ngapp-rollback-history" for prod, which would list the history from "ngapp-rollback-history-integ". we should use "ngapp-rollback-history-prod" instead. It was caused by a merge mistake from my side.

PR #124, opened 2016-04-18
Refactor CMP

yujunglo:
@sbaogang @jinruh 
ClusterManager code refactoring


deploy-service/common/src/main/java/com/pinterest/clusterservice/bean/ClusterBean.java
```diff
@@ -44,6 +45,7 @@
     private String host_type_id;
     private String security_zone_id;
     private String placement_id;
+    private String configs;
```
sbaogang: I think we need config_id

sbaogang:
A small suggestion, otherwise looks good, thanks! r+

PR #126, opened 2016-04-21
Fixed wrong conf path for downloader

yujunglo:
@sbaogang 
Fixed config file path for s3 downloader


sbaogang:
r+, thanks for the quick fix, can you also do a new release?

PR #128, opened 2016-04-25
Added isDocker to ping response

yujunglo:
@sbaogang

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_creation.tmpl
+++ b/deploy-board/deploy_board/templates/clusters/cluster_creation.tmpl
@@ -24,8 +24,23 @@
                         title="Number of hosts for this service">
                         Capacity
                     </label>
-                    <div class="col-xs-10">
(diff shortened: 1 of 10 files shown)


deploy-agent/deployd/common/utils.py
```diff
@@ -104,7 +104,7 @@ def run_prereqs(config):
         if respect_puppet and \
            puppet_file_path is not None and \
            not os.path.exists(puppet_file_path):
-            log.info("Waiting for first puppet run.")
+            print("Waiting for first puppet run.")
```
jinruh: forgot to change it back after debugging?

yujunglo: This line is called before the log is initialized, so we need to use print for this.

sbaogang:
Looks good! Now it is easier to switch back and forth for docker run, need to come up with safe solution for the switch before we commit this code.

PR #129, opened 2016-04-25
Fix crash on downloading multipart S3 uploads.

brennentsmith:
As S3 stores multipart upload md5’s in a special format within the
etag, standard md5sum’ing doesn’t work and causes the download to fail.
See:
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders
.html and


sbaogang:
r+, Thanks for the fix!

PR #131, opened 2016-04-26
added cluster migration worker

yujunglo:
@sbaogang 
Added 1 worker and 2 boolean in hostDAO
1. can_retire (default=0) indicated this host is scheduled for replacement/migration
2. can_replace (default=1); if can_replace=1, terminate host directly; otherwise, terminate and then reduce the cluster capacity.

Replace Cluster button:
screen shot 2016-04-26 at 2 00 45 pm

screen shot 2016-04-26 at 2 11 55 pm

If found other non-cmp groups,
screen shot 2016-04-26 at 2 02 05 pm

Show messages in env_landing page if env is scheduled for cluster replacement.
screen shot 2016-04-26 at 2 02 36 pm

PR #132, opened 2016-04-27
Allowing set a percentage number for maximum number of hosts deployed

lilida:
Allowing set a percentage number for maximum number of hosts deployed  in parallel. Originally, only a fixed number of hosts can be specified. This is not ideal for an environment with dynamically scaling. Now, we support set a percentage number. Note: The host number is still there. They are set not he same UI. When the input contains a %, it is treated a percentage. At UI and Bean level, we have tried to clear the other one when one is set.

Tests:
Unit tests coverage for algorithm to pick the right value.
Test set different values in the UI and validate the values have been used.

### DIFF
--- a/deploy-board/deploy_board/templates/configs/env_config.tmpl
+++ b/deploy-board/deploy_board/templates/configs/env_config.tmpl
@@ -8,14 +8,28 @@
                     <label for="maxParallelHosts" class="deployToolTip control-label col-xs-2"
                         data-toggle="tooltip"
                         title="Maximum number (or percentage) of hosts allowed to deploy at the same time">
-                        Maximum Deploys In Parallel
+                        Maximum Parallel Number
                     </label>
 
-                    <div class="col-xs-4">
+                    <div class="col-xs-2">                      
                         <div class="input-group">
-                            <input class="form-control" name="maxParallelHosts" required="true"
-                               type="text" value="{{ env.finalParallel }}"/>
-                            <span class="input-group-addon">eg. 5 or 20%</span>
+                            <span class="input-group-addon">
+                                <input type="radio" name="maxParallel" value="Number"
+                                    {% ifequal env.showNumber True%}checked{% endifequal %} >
+                            </span>
+                                <input type="text" id="maxParallelValue" name="maxParallelValue" class="form-control" 
+                                    value="{% ifequal env.showNumber True%}{{env.maxParallel}}{% endifequal %}">
+                        </div>
+                    </div>
+                    <div class="col-xs-2">
+                        <div class="input-group">
+                            <span class="input-group-addon">
+                                <input type="radio" name="maxParallel" value="Percentage"
+                                    {% ifequal env.showNumber False%}checked{% endifequal %}>
+                            </span>
+                            <input type="text" id="maxParallelPctValue" name="maxParallelPctValue" class="form-control" 
+                                    value="{% ifequal env.showNumber False %}{{ env.maxParallelPct }}{% endifequal %}">
+                            <span class="input-group-addon">%</span>
                         </div>
                     </div>
                     <label for="successThreshold" class="deployToolTip control-label col-xs-2"
@@ -269,6 +283,15 @@
                     $("#envConfigId").parent().html(data.html);
                 }
             });
+        });
+
+       $('input:radio[name=maxParallel]').change(function() {
+            if (this.value == 'Number') {
+                $('#maxParallelPctValue').attr('value','');   
+            }
+            else if (this.value == 'Percentage') {
+                 $('#maxParallelValue').attr('value', '');   
+            }
         });
     });
 </script>
--- a/deploy-board/deploy_board/templates/configs/env_config.tmpl
+++ b/deploy-board/deploy_board/templates/configs/env_config.tmpl
@@ -7,13 +7,16 @@
                 <div class="form-group">
                     <label for="maxParallelHosts" class="deployToolTip control-label col-xs-2"
                         data-toggle="tooltip"
-                        title="Maximum number of hosts allowed to deploy at the same time">
-                        Maximum Hosts
+                        title="Maximum number (or percentage) of hosts allowed to deploy at the same time">
+                        Maximum Deploys In Parallel
                     </label>
 
                     <div class="col-xs-4">
-                        <input class="form-control" name="maxParallelHosts" required="true"
-                               type="text" value="{{ env.maxParallel }}"/>
+                        <div class="input-group">
+                            <input class="form-control" name="maxParallelHosts" required="true"
+                               type="text" value="{{ env.finalParallel }}"/>
+                            <span class="input-group-addon">eg. 5 or 20%</span>
+                        </div>
                     </div>
                     <label for="successThreshold" class="deployToolTip control-label col-xs-2"
                         data-toggle="tooltip"
@@ -24,7 +27,7 @@
                     <div class="col-xs-4">
                         <div class="input-group">
                         <input class="form-control" name="successThreshold" required="true"
-                               type="text" value="{{ env.successThreshold|convertSuccThreshold }}"/>
+                               type="text" value="{{ env.successThreshold|convertSuccThreshold }}"/> 
                         <span class="input-group-addon">%</span>
                         </div>
                     </div>
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Pinterest, Inc.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *  
- *     http://www.apache.org/licenses/LICENSE-2.0
- *    
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -16,8 +16,10 @@
 package com.pinterest.deployservice.bean;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.pinterest.deployservice.common.Constants;
 import org.apache.commons.lang.builder.ReflectionToStringBuilder;
 import org.hibernate.validator.constraints.NotEmpty;
+import org.hibernate.validator.constraints.Range;
 
 import java.io.Serializable;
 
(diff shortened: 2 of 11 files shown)


michaelfu:
Instead of only allowing an absolute number or a percentage, why don't we allow people to set both in the UI and let them cap each other?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -517,4 +519,52 @@ PingResponseBean generateDeleteResponse(GoalAnalyst.UninstallCandidate candidate
...
+
+        if (!numIsApplicable && percentageIsApplicable) {
+            ret = (int) (totalHosts * environBean.getMax_parallel_deploy_percentage() / 100);
+            LOG.info("Max parallel count {} is decided by percentage solely. Percentage is {}",
+                    ret,
+                    environBean.getMax_parallel_deploy_percentage());
+        } else if (!percentageIsApplicable && numIsApplicable) {
+            ret = environBean.getMax_parallel();
+            LOG.info("Max parallel count {} is decided by number of host solely.", ret);
+        } else if (numIsApplicable && percentageIsApplicable) {
```
michaelfu: When could this happen? I thought only it is either number or percentage but not both?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -195,7 +195,9 @@ boolean canDeploy(EnvironBean envBean, String host, AgentBean agentBean) throws
         // TODO use ecache to optimize
         // Make sure we do not exceed allowed number of concurrent active deploying agent
         String envId = envBean.getEnv_id();
-        long parallelThreshold = envBean.getMax_parallel();
+        long totalHosts = environDAO.countTotalCapacity(envBean.getEnv_id(), envBean.getEnv_name(),
```
michaelfu: Does the totalHosts contain newly provisioned hosts? If yes then it may be dangerous. If suddenly a lot of new hosts are being launched, totalHosts could be bigger than # of hosts that can serve traffic and possibly allow more hosts to deploy.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
lilida: There is no guard in DB level for directly DB update. This is just added for safety

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
lilida: Good point. I believe it contains newly provisioned hosts as there is no filter on hosts state. Will further look into this to confirm. The logic right now is following what exists.

lilida:
Add a screen shot

screen shot 2016-04-26 at 6 28 18 pm

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
michaelfu: Sounds good.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
yujunglo: The countTotalCapacity currently contains newly provisioned hosts, missing hosts and active hosts. I modified the logic in #131 to only contain newly provisioned hosts and active hosts. Let me know if this does not make sense to you.

lilida: Thanks. In this particular scenario, I think only active hosts should be counted in order to avoiding the risk bringing too many active hosts while adding large number of new provisioned hosts.

yujunglo: If you only want the active hosts, you can use this query: agentDAO.getAllByEnv(envBean.getEnv_id()).size().

deploy-board/deploy_board/templates/configs/env_config.tmpl
```diff
@@ -7,13 +7,16 @@
                 <div class="form-group">
                     <label for="maxParallelHosts" class="deployToolTip control-label col-xs-2"
                         data-toggle="tooltip"
-                        title="Maximum number of hosts allowed to deploy at the same time">
-                        Maximum Hosts
+                        title="Maximum number (or percentage) of hosts allowed to deploy at the same time">
+                        Maximum Deploys In Parallel
```
sbaogang: Deploy has been accepted as a higher level concept, usually they are in the context of environments. "Deploys in parallel" sounds like there are multiple deploys going on for a service.

deploy-board/deploy_board/templates/configs/env_config.tmpl
```diff
@@ -7,13 +7,16 @@
...
                     </label>
 
                     <div class="col-xs-4">
-                        <input class="form-control" name="maxParallelHosts" required="true"
-                               type="text" value="{{ env.maxParallel }}"/>
+                        <div class="input-group">
+                            <input class="form-control" name="maxParallelHosts" required="true"
+                               type="text" value="{{ env.finalParallel }}"/>
+                            <span class="input-group-addon">eg. 5 or 20%</span>
```
sbaogang: can we ask add a radiobutton, or dropdown, for users to explicitly choose percentage or absolute numbers? I feel user have to type % in the input box not very elegant

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -52,6 +54,7 @@
  * max_deploy_num      INT           NOT NULL,
  * max_deploy_day      INT           NOT NULL,
  * is_docker           TINYINT(1)    DEFAULT 0,
+ * max_parallel_deploy_percentage TINYINT(1)
```
sbaogang: to be consistent, can we make it shorter? we have been user short names. something like max_parallel_pct

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -517,4 +519,52 @@ PingResponseBean generateDeleteResponse(GoalAnalyst.UninstallCandidate candidate
...
+     * 2. If both of them are set, use the minimum value
+     * @param environBean
+     * @param totalHosts
+     * @return The maximum parallel count the system respects
+     * @throws Exception
+     */
+    public final static int getFinalMaxParallelCount(EnvironBean environBean, long totalHosts) throws Exception {
+
+        int ret = Constants.DEFAULT_MAX_PARALLEL_HOSTS;
+        LOG.info("Get final maximum parallel count for total capactiy {}", totalHosts);
```
sbaogang: can we change info->debug?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -517,4 +519,52 @@ PingResponseBean generateDeleteResponse(GoalAnalyst.UninstallCandidate candidate
...
+        } else if (numIsApplicable && percentageIsApplicable) {
+            ret = Math.min(environBean.getMax_parallel(),
+                    (int) (totalHosts * environBean.getMax_parallel_deploy_percentage() / 100));
+            LOG.info("Max parallel count {} is decided by combing host {} and percentage {}.",
+                    ret, environBean.getMax_parallel(), environBean.getMax_parallel_deploy_percentage());
+        }
+
+        //Ensure the value falls into the range making sense
+        if (ret <= 0) {
+            ret = 1;
```
sbaogang: can we log a warning here?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -517,4 +519,52 @@ PingResponseBean generateDeleteResponse(GoalAnalyst.UninstallCandidate candidate
...
+        } else if (ret > totalHosts) {
+            ret = (int) totalHosts;
+        }
+        return ret;
```
sbaogang: a suggestion, regardless of applicable or not, we could just compute both the totals and return the min(). I think it will make the code much simpler and easy to read.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
lilida: Use all agents-first deploy now

deploy-board/deploy_board/templates/configs/env_config.tmpl
lilida: Change to maximum parallel number

deploy-board/deploy_board/templates/configs/env_config.tmpl
lilida: Here you are.

screen shot 2016-04-27 at 2 59 59 pm

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
lilida: Done

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
lilida: Done

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
lilida: I actually like to keep it in the current format. We still need to exclude 0 in the min computation that it doesn't look like simpler to me

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
lilida: Done

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
sbaogang: We do not need to recompute this multiple times:
(int) (totalHosts * environBean.getMax_parallel_pct() / 100)

Also, take a look at the following code, does that make sense? Either way, if you just remove the dup code, that would be just fine. Thanks!

=========
        int maxHostsPerPCT = (int) (totalHosts * environBean.getMax_parallel_pct() / 100);
        int maxHostsPerNumber = environBean.getMax_parallel();
        int ret = Math.min(maxHostsPerNumber, maxHostsPerPCT);
        LOG.debug("Max parallel count {} is decided by getting the min of host {} and percentage {}.", ret, maxHostsPerNumber, maxHostsPerPCT);
=========

lilida: It does not work this way as the current system uses 0 to indicate the option (number or percentage) is not used. The math.min will get 0 in this case. The things behind this is we want user only to use one of them in most scenario instead of always set them both which I believe bringing confusion and extra burden to them. The min should be an exceptional case only. Also the current code makes it be safer to handle invalid values (like negative number etc.) There is no extra computation cost at runtime as only one branch is entered.

sbaogang: Yep, you are right!

sbaogang:
r+, thanks for the new feature!

PR #134, opened 2016-04-28
Auto check radio button when a user types in the input box

lilida:
When a user types in the input. Radio button should be checked automatically. Also fix format issue


sbaogang:
Thanks for the quick fix, r+

PR #135, opened 2016-04-29
Read aws launch config directly instead of reading from db

yujunglo:
@jinruh @sbaogang 
encoding user data and transforming role name into instance profile will be taken care by createLaunchConfigInternal (in AwsAutoScalingManager) and launchEC2Instances (in EC2HostInfoDAOImpl). 

decoding user data and transforming instance profile to role name will be take care by getLaunchConfigInfo (in AwsAutoScalingManager).

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/arcee/autoscaling/AwsAutoScalingManager.java
+++ b/deploy-service/common/src/main/java/com/pinterest/arcee/autoscaling/AwsAutoScalingManager.java
@@ -120,11 +120,10 @@ public class AwsAutoScalingManager implements AutoScalingManager {
         monitoring.setEnabled(false);
         configurationRequest.setInstanceMonitoring(monitoring);
         if (StringUtils.isEmpty(request.getRawUserDataString())) {
-            String userData =
-                transformUserDataConfigToString(groupName, request.getUserDataConfigs());
+            String userData = transformUserDataConfigToString(groupName, request.getUserDataConfigs());
             configurationRequest.setUserData(Base64.encodeBase64String(userData.getBytes()));
         } else {
-            configurationRequest.setUserData(request.getRawUserDataString());
+            configurationRequest.setUserData(Base64.encodeBase64String(request.getRawUserDataString().getBytes()));
         }
 
         aasClient.createLaunchConfiguration(configurationRequest);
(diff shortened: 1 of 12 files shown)


deploy-service/common/src/main/java/com/pinterest/arcee/aws/EC2HostInfoDAOImpl.java
```diff
@@ -99,12 +106,14 @@ public void terminateHost(String hostId) throws Exception {
         RunInstancesRequest request = new RunInstancesRequest();
         request.setImageId(groupBean.getImage_id());
         request.setInstanceType(groupBean.getInstance_type());
-        request.setKeyName("ops");
+        request.setKeyName(vmKeyName);
         request.setSecurityGroupIds(Arrays.asList(groupBean.getSecurity_group()));
         request.setSubnetId(subnet);
-        request.setUserData(groupBean.getUser_data());
+        String userData = Base64.encodeBase64String(groupBean.getUser_data().getBytes());
```
jinruh: Just to double check: the userdata we read from groupbean is always decoded string, right?

yujunglo: Yes, groupBean always stores decoding data.

deploy-service/common/src/main/java/com/pinterest/arcee/autoscaling/AwsAutoScalingManager.java
```diff
@@ -548,25 +547,43 @@ public void enableScalingDownEvent(String groupName) throws Exception {
...
-        if (group == null) {
-            LOG.warn(String.format("Failed to get cluster %s: auto scaling group %s does not exist", groupName, groupName));
+    public AwsVmBean getAutoScalingGroupInfo(String clusterName) throws Exception {
+        AutoScalingGroup autoScalingGroup = getAutoScalingGroup(clusterName);
+        if (autoScalingGroup == null) {
+            LOG.warn(String.format("Failed to get cluster %s: auto scaling group %s does not exist", clusterName, clusterName));
             return null;
         }
 
+        AwsVmBean launchConfigInfo = getLaunchConfigInfo(clusterName, autoScalingGroup.getLaunchConfigurationName());
```
jinruh: In theory, you probably don't need the clusterName as parameter to the API.

yujunglo:
updaed

jinruh:
r+ thanks! looks good!

PR #138, opened 2016-05-03
Added terminated host check in worker

yujunglo:
@jinruh


jinruh:
Thanks for the quick fix. r+

PR #139, opened 2016-05-03
Remove hardcoded recent environment

brennentsmith:
Either this existed for a demo, dev purposes, or something Pinterest Internal related. When clicked, it causes essentially a 404 error to occur.


sbaogang:
Thanks for bring this up, that item should be guarded by an if statement. and it needs a change in the view as well, I will create PR for that

sbaogang:
https://github.com/pinterest/teletraan/pull/141 should fix this. Thanks again for bringing this up, closing this for now...

brennentsmith:
Excellent - thanks for the update and I appreciate the fix!

PR #143, opened 2016-05-05
Deprecating detach instance from asg function

jinruh:
We can do instance protection in auto scaling group now.
@yujunglo @sbaogang 

screen shot 2016-05-04 at 6 21 56 pm
screen shot 2016-05-04 at 6 22 32 pm


deploy-board/deploy_board/templates/host_side_panel.html
```diff
@@ -72,34 +72,66 @@ <h4 class="modal-title">Host Termination Confirm</h4>
 {% endif %}
 
 
-<!--- detach instances -->
 {% if asg_group %}
```
yujunglo: Could you also add "if pinterest" here?

yujunglo:
looks good, thanks.

PR #144, opened 2016-05-05
Added a worker for cluster rolling upgrade

yujunglo:
@sbaogang 
1. Replace cluster button:
screen shot 2016-05-05 at 3 23 44 pm
screen shot 2016-05-05 at 3 23 54 pm

2. warning message and pause/resume/cancel button
screen shot 2016-05-05 at 3 18 24 pm
screen shot 2016-05-05 at 3 22 20 pm
screen shot 2016-05-05 at 3 22 39 pm

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/clusters.html
+++ b/deploy-board/deploy_board/templates/clusters/clusters.html
@@ -30,11 +30,31 @@
             data-toggle="modal" title="Delete Cluster">
             <span class="glyphicon glyphicon-remove-sign"></span> Delete cluster
             </button>
-            <!--- TODO To be implemented
-            <button class="deployToolTip btn btn-default btn-block" data-target="#replaceFleet"
-            data-toggle="modal" title="Replace fleet">
+            <!-- disable for now
+            {% if basic_cluster_info|canReplaceCluster %}
+            <button class="deployToolTip btn btn-default btn-block" data-target="#enableReplacement"
+            data-toggle="modal" title="Click to Schedule cluster replacement">
             <span class="glyphicon glyphicon-random"></span> Replace cluster
             </button>
+            {% elif basic_cluster_info.state == "REPLACE" %}
+                <button class="deployToolTip btn btn-default btn-block" data-target="#pauseReplacement"
+                data-toggle="modal" title="Click to pause cluster replacement">
+                <span class="glyphicon glyphicon-pause"></span> PAUSE
+                </button>
+                <button class="deployToolTip btn btn-default btn-block" data-target="#cancelReplacement"
+                data-toggle="modal" title="Click to cancel cluster replacement">
+                <span class="glyphicon glyphicon-step-backward"></span> Cancel
+                </button>
+            {% elif basic_cluster_info.state == "PAUSE" %}
+                <button class="deployToolTip btn btn-default btn-block" data-target="#resumeReplacement"
+                data-toggle="modal" title="Click to resume cluster replacement">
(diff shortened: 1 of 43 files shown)


deploy-board/deploy_board/templates/clusters/clusters.html
```diff
@@ -47,6 +65,13 @@ <h4 class="panel-title pull-left">Cluster Manager</h4>
 {% endblock %}
 
 {% block main %}
+{% if not basic_cluster_info|canReplaceCluster %}
+    <div class="alert alert-warning" role="alert">
+        <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <strong>Warning!</strong> This environment stage has scheduled cluster rolling upgrade.
```
sbaogang: Since we think the cluster replace should be be interfering with regular deploy - I would suggest change this to info ( light blue)

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Clusters.java
```diff
@@ -124,6 +133,38 @@ public void deleteCluster(@Context SecurityContext sc,
     }
 
     @PUT
+    @Path("/actions")
+    public void replaceCluster(@Context SecurityContext sc,
+                               @PathParam("envName") String envName,
+                               @PathParam("stageName") String stageName,
+                               @QueryParam("type") String type) throws Exception {
```
sbaogang: We should use ActionType directly - the framework will be responsible to convert string to enum automatically for us, and throw exceptions if string is illegal.
@QueryParam("type") ActionType type

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Clusters.java
```diff
@@ -124,6 +133,38 @@ public void deleteCluster(@Context SecurityContext sc,
...
+                clusterHandler.replaceCluster(envName, stageName);
+                break;
+            case PAUSE:
+                clusterHandler.pause(envName, stageName);
+                break;
+            case RESUME:
+                clusterHandler.resume(envName, stageName);
+                break;
+            case CANCEL:
+                clusterHandler.complete(envName, stageName);
```
sbaogang: suggest to use the following names:
replace/pauseReplace/resumeReplace/cancelReplace

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/ClusterReplacer.java
```diff
@@ -0,0 +1,151 @@
...
+        String clusterName = clusterBean.getCluster_name();
+        List<EnvironBean> envs = environDAO.getEnvsByGroups(Collections.singletonList(clusterName));
+        EnvironBean envBean = null;
+        for (EnvironBean env : envs) {
+            // Find the env/stage
+            if (clusterName.equals(clusterHandler.getClusterName(env.getEnv_name(), env.getStage_name()))) {
+                envBean = env;
+                break;
+            }
+        }
```
sbaogang: If we think there is a one-to-one mapping, we could just use cluste rname and derive the env/stage name.

yujunglo:
updated api and messages.

yujunglo:
Updated to make rolling upgrade process more robust and able to be timeout and abort.

PR #149, opened 2016-05-16
add system priority to env, we will use it for sidecar deploys

sbaogang:
This change will not break users. I would think this is long term solution as well. I am not convinced regular deploys need such finer level of control on deploy orders, it might just complicate Teletraan and confuse users. Plan to keep this five higher level priority, and keep this system_priority hidden from user for now. (we could expose this through system level UI or API in the future).

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/DeployPriority.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/DeployPriority.java
@@ -16,7 +16,7 @@
 package com.pinterest.deployservice.bean;
 
 public enum DeployPriority {
-    NORMAL(30), HIGH(20), LOW(40), HIGHER(10), LOWER(60);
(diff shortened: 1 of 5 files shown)


sbaogang:
@jinruh @yujunglo @lilida

sbaogang:
more background - so that we could use this finer level priorities to keep the sidecar services start in predefined order

yujunglo:
Looks good to me. r+

lilida:
What is the expecting value in System Priority as an integer?

sbaogang:
Integers, in general it is not good idea to use integers for priorities, since it is hard to coordinate with different service (owners), but since this is global/system services, and mostly owned by the same person, it makes sense. To start, we could have 10/20/30/... system priority up to 99

lilida:
I see. Looks good. r+

jinruh:
looks good. r+

PR #153, opened 2016-05-18
Added cluster name to environs table

yujunglo:
@sbaogang 
when creating and removing clusters, also update the cluster name in environs table


sbaogang:
Looks good, thanks! r+

PR #159, opened 2016-05-25
use embeded DB and the mysql scheme inside the release tar

sbaogang:
so our demo run will break due to DB changes.


sbaogang:
@yujunglo

yujunglo:
Looks good to me! r+

PR #165, opened 2016-06-01
Fix OAUTH access token variable name

brennentsmith:
The variable is currently misnamed. As a result, it doesn't properly override the default access token URL, causing it to call against Pinterest's internal system rather than the specified endpoint. 

See: https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/settings.py#L55 for instantiation location.


nickdechant:
Hey there, nice catch! Thanks for the fix. 👍 💯

brennentsmith:
No problem :)

Issue #161, opened 2016-05-29
username and password for teletraan UI 


sergeyr87:
hi 
how can i create a username access to the webui?

sbaogang:
Teletraan does not manage user identities, but rather it integrates with OAuth provider, such as google openid, to authenticate user. To enable authentication/authorization, you need to configure them in both the deploy-board and deploy-service. see both manage.py and this page https://github.com/pinterest/teletraan/wiki/Setting-up-user-roles for more details, let us know if you have any further questions

brennentsmith:
We were able to get Google OAuth working with Teletraan, but it did require some minor modifications to deploy-board and deploy-service due to how keys were extracted. If I have time, I'll put together a PR and tutorial.

sergeyr87:
great thanks :)

sbaogang:
Close this for now, let know if there is any question setting up OAuth integration. thanks!

PR #169, opened 2016-06-03
Google OAUTH2 Support

brennentsmith:
This provides the ability to specify Google’s OAUTH2 service and have
Teletraan auth against it. If the option is not explicitly set, the default
is to fall back to Pinadmin as the OAUTH2 provider. 

Fixed spellings, formatting (quotes), and added SECRET_KEY environment variable to default settings file (since it is a required variable for OAUTH2).

While I have tested this against Google's endpoints, I have not been able to (for obvious reasons) test against pinadmin. 

This should help solve questions like: https://github.com/pinterest/teletraan/issues/161

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/UserDataHelper.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/UserDataHelper.java
@@ -84,15 +84,15 @@ public class UserDataHelper {
         Map<String, String> params = ImmutableMap.of("access_token", token);
         String jsonPayload = httpClient.get(userDataUrl, params, null, 3);
         JsonObject jsonObject = new JsonParser().parse(jsonPayload).getAsJsonObject();
-        final String user;
-        if (oauthProvider.equals("googlev2")){
-            String email = jsonObject.getAsJsonPrimitive("email").getAsString();
-            user = email.split("@")[0];
-        } else { 
-            // fall back to pinadmin
-            jsonObject = jsonObject.getAsJsonObject("user");
-            user = jsonObject.get("username").getAsString();
-        } 
+        String user;
+        switch(oauthProvider) {
+            case "pinadmin": {
+                user = jsonObject.getAsJsonObject("user").get("username").getAsString();
+            }
+            default: {
+                user = jsonObject.getAsJsonPrimitive("email").getAsString().split("@")[0];
+            }
+        }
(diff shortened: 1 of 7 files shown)


nickdechant:
Hey! Cool fix! Thanks for the pull request! We're going to take another look at this and experiment with our internal use vs OpenID vs other providers and how we want to handle defining providers and their response formats. Will put a hold on merging this until that is done.

brennentsmith:
👍 Sounds good to me! Looking forward to see what your team comes up with

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/UserDataHelper.java
```diff
@@ -82,8 +84,15 @@ public String getUsername(String token) throws Exception {
         Map<String, String> params = ImmutableMap.of("access_token", token);
         String jsonPayload = httpClient.get(userDataUrl, params, null, 3);
         JsonObject jsonObject = new JsonParser().parse(jsonPayload).getAsJsonObject();
-        jsonObject = jsonObject.getAsJsonObject("user");
-        String user = jsonObject.get("username").getAsString();
+        final String user;
+        if (oauthProvider.equals("googlev2")){
```
nickdechant: I'm thinking this should be the default. No need to check if "googlev2" or any other string at this time. Google follows the OpenID Connect format for UserInfo responses so this code would apply to other sources of providers.

Im summary:
if pinadmin: get username
else: get by email

brennentsmith: I agree that we could use that as the default, but it may be good to retain the oauthProvider variable for quirk handling.

sbaogang: I would agree - let us keep the oauthProvider directive, but treat Pinterest Provider a special case. Any other providers will fall back to default behavior

brennentsmith:
Moved to a switch rather than conditional blocks for cleanliness and made OpenID the default operation.

Arch88:
Hi Teletraan team, 

Would you please merge this branch.

sbaogang:
will double check the fix and update

sbaogang:
could you take a look at https://github.com/pinterest/teletraan/pull/285, I believe it handles the openid ( and internal pinterest case ) easier.

tylerwowen:
Close as PR is too old.

PR #174, opened 2016-06-07
Make auto scaling backward compatible

jinruh:
So service owners  can still use launch config without enabling auto scaling to launch new instances.
Also add batch attach instance function.

@yujunglo @nickdechant @sbaogang


deploy-board/deploy_board/templates/hosts/hosts.tmpl
```diff
@@ -53,7 +53,7 @@
   </div>
 
 <div class="panel-footer clearfix">
-    {% if is_retryable %}
+    {% if is_retryable host_ids %}
```
yujunglo: Where is the host_ids from?

jinruh: I don't know. Must be a typo, I guess :)

yujunglo:
r+!

PR #176, opened 2016-06-07
Add a link in Host Details for more information on host + Set override policy for environment

sonia-y:


sbaogang:
Look good to me, thanks! r+

sbaogang:
@jinruh it would be great if we can also render the content natively in Teletraan, along with the link ( so people could follow the link to CMDB)

sbaogang:
if this is not trivial, forget about it, this is good enough! Thanks!

jinruh:
Thanks baogang, I would think to check in this first, and we can add more stuff later. 
r+

deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -88,6 +88,7 @@ def get(self, request, groupname, hostname):
             'asg_group': asg,
             'is_unreachable': is_unreachable,
             'pinterest': IS_PINTEREST,
+            'host_information_url': TELETRAN_HOST_INFORMATION_URL,
```
yujunglo: missing TELETRAN_HOST_INFORMATION_URL import from deploy_board.settings

PR #177, opened 2016-06-07
Fixed missing launch config bug

yujunglo:
@jinruh


jinruh:
r+
Thanks for the quick fix!

PR #181, opened 2016-06-08
Fix a bug for launching instance when auto scaling group is not available

jinruh:
as long as we've stored the launch config in the database, even the asg is not enabled, we should be able to launch instance.

@yujunglo


yujunglo:
thanks for fixing. r+!

PR #182, opened 2016-06-09
Fix update error for pas config

nickdechant:


jinruh:
looks good. Thanks! r+

PR #183, opened 2016-06-09
Skip the env installation if system env installation failed

yujunglo:
@sbaogang 
If the system env installation failed, skip the rest of the env installations


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -92,17 +92,17 @@
 
         int getDeployPriority(EnvironBean env) {
             if (env.getDeploy_type() == null) {
-              return DeployPriority.NORMAL.getValue();
+                return DeployPriority.NORMAL.getValue();
```
sbaogang: can we remove the extra space?

tylerwowen:
Close as PR is too old.

PR #185, opened 2016-06-10
Fix for new envs with same name and updating roles table

nickdechant:


sbaogang:
r+

yujunglo:
r+ thanks for fixing!

PR #188, opened 2016-06-14
Moved PasConfigBean over to deploy-service

nickdechant:


deploy-service/common/src/main/java/com/pinterest/arcee/bean/PasConfigBean.java
```diff
@@ -25,8 +25,43 @@
     @JsonProperty("pas_state")
     private PasState pas_state;
 
+    @JsonProperty("min_size")
```
jinruh: What are these used for?

nickdechant: Ah I suppose nothing on Arcee side of things. I'll remove them.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/AutoScalingGroups.java
```diff
@@ -329,9 +336,22 @@ public void updatePasConfig(@Context SecurityContext sc,
...
+
+        // Combine PasConfigBean data and data from AWS
+        pasConfig.put("group_name", groupName);
+        pasConfig.put("throughput", config.getThroughput());
+        pasConfig.put("metric", config.getMetric());
+        pasConfig.put("pas_state", config.getPas_state());
+        pasConfig.put("last_updated", config.getLast_updated());
+        pasConfig.put("max_size", group.getMaxSize());
+        pasConfig.put("min_size", group.getMinSize());
+        pasConfig.put("cooldown", group.getDefaultCooldown());
```
jinruh: You need to read this information from ScalingPolicy.

PR #200, opened 2016-06-23
Delete SQS message if the group does not exist in the database

jinruh:
as title. if the group does not exist, we need to delete SQS message to avoid infinite loop

@yujunglo @sbaogang


yujunglo:
Looks good. r+

PR #202, opened 2016-06-23
Quick stage removal typo fix

eweizman:
Stage name wasn't displaying in remove stage modal because of a typo. Fixed that and added notice to the user if they are about to delete the environment (by removing the last stage).


nickdechant:
Thanks for the improvement! r+

PR #209, opened 2016-06-28
Spot Call Additions

nickdechant:


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/AutoScalingGroups.java
```diff
@@ -429,4 +432,24 @@ public void deleteScheduledActionFromAutoScalingGroup(@Context SecurityContext s
...
+    @GET
+    @Path("/spot_instances/details")
+    public SpotAutoScalingBean getSpotAutoScalingDetails(@Context SecurityContext sc,
+                                                         @PathParam("groupName") String groupName) throws Exception {
+        return spotAutoScalingDAO.getAutoScalingGroupsByCluster(groupName);
+    }
+
+    @GET
+    @Path("/spot_instances/spot_name")
+    public String getSpotNameByClusterName(@Context SecurityContext sc,
```
yujunglo: If you want to return a bare string, you need to add "\". e.g. return "\"foo\"" instead of "foo".

PR #213, opened 2016-07-01
fix the deployagent detect running pattern issue due to the recent pa…

sbaogang:
…ttern changed

fix the deployagent detect running pattern issue due to the recent pattern changed

@yujunglo 
@haom-pinterest


yujunglo:
r+

PR #214, opened 2016-07-07
Save Min and Max AS Group Values to Database

nickdechant:


deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -268,6 +268,14 @@ def create_asg(request, group_name):
 
 def get_asg_config(request, group_name):
     asg_summary = groups_helper.get_autoscaling_summary(request, group_name)
+    pas_config = groups_helper.get_pas_config(request, group_name)
+
+    # Check saved min and max values for autoscaling group
+    if "desired_min_size" in pas_config and pas_config.get("desired_min_size"):
+        asg_summary["minSize"] = pas_config.get("desired_min_size")
+    if "desired_max_size" in pas_config and pas_config.get("desired_max_size"):
+        asg_summary["maxSize"] = pas_config.get("desired_max_size")
```
jinruh: I think in asg_summary,  we should still show the current asg min size to avoid confusing people.

jinruh:
You probably also want to update the getPasConfig API to load min/max size from database instead of aws.

PR #219, opened 2016-07-20
Add exception handler

yujunglo:
@sbaogang 
Display full stack trace to user. 
This can be disable by setting disableStackTrace: true in yaml file.


sbaogang:
r+, thanks for the fix!

PR #220, opened 2016-07-21
Remove cluster/asg related code

yujunglo:
@sbaogang @jinruh @nickdechant 
1. Remove cluster related code and tables.
2. Add rodimus manager to get host state and terminate hosts


sbaogang:
Thanks for the major clean up, R+!

PR #221, opened 2016-07-25
[pinterest only] move cluster manager to capacity page

yujunglo:
@sbaogang 
1. Create button
screen shot 2016-07-25 at 12 02 32 pm

2. Create page
screen shot 2016-07-25 at 12 02 47 pm

3. After creating
screen shot 2016-07-25 at 12 03 04 pm


sbaogang:
r+, thanks!

PR #229, opened 2016-08-02
Fix new deploy creation

yujunglo:
@sbaogang @nickdechant


sbaogang:
r+, thanks for the fix!

PR #232, opened 2016-08-06
add db schema upgrade script

sbaogang:
Create schema-versions table which will track DB schema versions and use upgrade.sh to upgrade it. For example, assume current schema version is 2, the upgrade script will first read the current schema version, and run only schema-upgrade-3.sql,  schema-upgrade-4.sql... until the latest version.

From this version on, when we change schema, we need to create the schema-upgrade-N+1.sql, assume current version is N

This is believed to be much effective and easier to track and upgrade than the upgrade text file we kept before. Let us use this new approach from now on.


sbaogang:
@haom-pinterest @jinruh @yujunglo @nickdechant

yujunglo:
looks good to me. Thanks.

PR #234, opened 2016-08-09
Put extra host details inside host page

sonia-y:
Reviewers: 
@jinruh

### DIFF
--- a/deploy-board/deploy_board/webapp/host_views.py
+++ b/deploy-board/deploy_board/webapp/host_views.py
@@ -19,6 +19,9 @@ import logging
 from helpers import environs_helper, agents_helper, autoscaling_groups_helper
 from helpers import environ_hosts_helper, hosts_helper
 from deploy_board.settings import IS_PINTEREST, TELETRAAN_HOST_INFORMATION_URL
+import brood
+from datetime import datetime
+import pytz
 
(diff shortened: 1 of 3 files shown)


sonia-y:
UI: 
image

jinruh:
Looks good. It would be nice to also add ami id to the host information.

jinruh:
looks good. r+

deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -19,6 +19,9 @@
 from helpers import environs_helper, agents_helper, autoscaling_groups_helper
 from helpers import environ_hosts_helper, hosts_helper
 from deploy_board.settings import IS_PINTEREST, TELETRAAN_HOST_INFORMATION_URL
+import brood
```
jinruh: Didn't notice this at first place. This is a Pinterest specific library. I would rather write a simple http client to query cmdb instead of adding a new library dependency.

sbaogang: for a simpler fix, I would suggest just add IS_PINTEREST around it, and hide the part uses brood function as well ( probably the part show host details, the opensource version does not have this feature any ways)

PR #240, opened 2016-08-10
Fix private build notice

sonia-y:
Reviewers:
@jinruh 

Fix private build notice and make sure it only appears when it is actually a private build


jinruh:
r+

PR #241, opened 2016-08-10
Fix config history changes styling

sonia-y:
Reviewers:
@yujunglo 

Fix overlapping rows in changes column in config history


yujunglo:
Thanks for the fix. r+

PR #251, opened 2016-09-03
fix the host retry button not working in multiple env case (#249)

sbaogang:
drop schema_versions table as well in cleanup.sql

@haom-pinterest


haom-pinterest:
LGTM

PR #260, opened 2016-10-21
Add docker compose for teletraan.

lilida:
@sbaogang @haom-pinterest 
This is docker-compose for teletraan dev. Now the expect steps for starting should be just
mvn for deploy-service and docker-compose up.
(The first time still needs mysql provision. That will come later through a docker run on the mysql image)


deploy-board/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:2.7
+ENV PYTHONUNBUFFERED 1 
+
+RUN mkdir /opt/deploy-board
+WORKDIR /opt/deploy-board
```
sjoshi6: We can refactor the /opt/deploy-board to be referred via a single 
ENV PROJECT_DIR = /opt/deploy-board

Then say
RUN mkdir $PROJECT_DIR

deploy-board/manage_docker_compose.py
```diff
@@ -0,0 +1,57 @@
...
+    #
+    # Backend Teletraan service url and settings. Change it accordingly in prod
+    #
+    os.environ.setdefault("TELETRAAN_SERVICE_URL", "http://deploy-service:8080")
+    os.environ.setdefault("TELETRAAN_SERVICE_VERSION", "v1")
+
+    #
+    # Logging
+    #
+    os.environ.setdefault("LOG_DIR", "/tmp/deploy_board")
```
sjoshi6: Is the /tmp/deploy_board directory created anywhere ?

sbaogang:
what is the difference on manage.py and manage_docker_compose.py, can we share so we do not end up with all the dup code? same true for service.yaml file

sbaogang:
can we provide instruction in wiki on how to run in docker?

lilida:
@sbaogang The main difference is the service endpoint. In docker compose, localhost won't work and the container names are used. I have removed the python one and used environment variables. For server.yaml, it is a bit complicated. We have a hierarchy structure here that doesn't fit the environment variable preferred for docker/docker-compose world that is flat. I need to think more about it. I would prefer keep that as a separate file right now.

@sjoshi6 Goog suggestion for defining variables. I have updated that. For log dir, I don't mount it right now. I feel for a dev environment, console output or console redirection are good enough.

lilida:
@sbaogang Updated wiki

sbaogang:
r+

PR #262, opened 2016-10-24
Better handling of DELETE ops on the agent side

sbaogang:
This fix will actually delete the "deleted env" from agent side. This is consider a mid-term solution until we fully implement the DELETE or SHUTDOWN feature

@haom-pinterest @lilida


lilida:
r+

PR #269, opened 2016-11-08
add Dockerfile for deployagent, will update docs once ready

sbaogang:
@lilida @haom-pinterest @yongwen


sbaogang:
the plan is for developer on their local box, execute the following command:

docker run deployagent -e FOO:BAR --mode serverless --build {...}

will run their service (in docker) very similar to production

lilida:
r+

PR #270, opened 2016-11-09
fix the wrong help on DOW

sbaogang:
The previous help message is not accurate, we use quartz and the detailed doc could be found here: http://www.quartz-scheduler.org/documentation/quartz-2.x/tutorials/crontrigger.html, especially:
Day of week YES 1-7 or SUN-SAT

@haom-pinterest @lilida

### DIFF
--- a/deploy-board/deploy_board/templates/configs/promote_config.tmpl
+++ b/deploy-board/deploy_board/templates/configs/promote_config.tmpl
@@ -144,6 +144,7 @@
   <tr><td>DOW</td><td>The day of the week (1-7 or SUN-SAT)</td></tr>
   </table>
   <ul>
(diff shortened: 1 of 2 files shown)


haom-pinterest:
Could you also make it clear the time is UTC based if it is not there?

haom-pinterest:
I also feel that we do not have validation logic on schedule either on the UI and on the server side when it is changed because I found there were a bunch of errors in the Teletraan worker logs saying invalid schedule. This should be avoided during the interaction with users. Can be in another PR though.

sbaogang:
right, could add validation. it is not easy to get it right.

haom-pinterest:
r+

PR #273, opened 2016-11-16
Add PUT and DELETE methods to webhooks

brennentsmith:


sbaogang:
r+, thanks for the fix!

PR #275, opened 2016-11-22
Gracefully handle 401 code from oauth

sjoshi6:
@sbaogang @lilida


sbaogang:
r+, thanks for the fix!

PR #278, opened 2016-11-28
specified absolute path (/var/teletraan/tools) for files referenced i…

ruthgrace:
…n upgrade.sh, to prevent file not found errors

This error occured when I followed the instructions here: https://github.com/pinterest/teletraan/wiki/Installation-for-developers

### DIFF
--- a/tools/mysql/upgrade.sh
+++ b/tools/mysql/upgrade.sh
@@ -91,7 +91,7 @@ done
 [ ! -z "$PORT" ] && PORT="--port=$PORT" 
 [ ! -z "$PASSWORD" ] && PASSWORD="--password=$PASSWORD" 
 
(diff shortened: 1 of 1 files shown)


sbaogang:
Thanks Ruth for the PR. Ideally we do not assume check_version.sql has to be in a hardcoded location, in this case we do not know if the script is indeed under /var/teletraan (I guess you are using docker, which is only one of the use cases). I believe the upgrade.sh should be able to run under the same dir where check_version.sql resides. If we would refactoring the script to use the dir of upgrade.sh as the root dir of check_version, that would be considered better change ( since we know these two files are always at the same dir).

sbaogang:
Thanks, r+

PR #279, opened 2016-12-02
add logout feature

sbaogang:
added logout feature, also fix some minor issues, added docker run without DB. 

@lilida @yongwen @haom-pinterest

### DIFF
--- a/deploy-board/deploy_board/webapp/security.py
+++ b/deploy-board/deploy_board/webapp/security.py
@@ -74,15 +74,6 @@ class DelegatedOAuthMiddleware(object):
             logger.debug("Redirect oauth for authentication!, url = " + url)
             return HttpResponseRedirect(url)
 
-    # TODO not currently used, need to add logout button on the UI and call this
-    def logout(self, request):
-        self.oauth.logout(session=request.session)
-
-        if 'teletraan_user' in request.session:
-            del request.session['teletraan_user']
-
-        return HttpResponseRedirect('/')
-
 
 class FixedOAuthMiddleware(object):
     """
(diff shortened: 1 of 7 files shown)


deploy-service/teletraanservice/bin/server-no-db.yaml
```diff
@@ -0,0 +1,92 @@
...
+      target: stdout
+
+
+#
+# Mysql connection config. By default, we use the real mysql as the backend.
+# Note: the embeded mysql is for demo or test purpose only. Use the official Mysql
+# installs for production usage.
+#
+db:
+  type: embedded
```
sbaogang: This is really what I want to change - there is no better way I could find to not copy the whole server.yaml

deploy-service/teletraanservice/bin/server.yaml
```diff
@@ -51,7 +51,7 @@ db:
   type: mysql
 
   # Mysql instance configurations
-  host: ${TELETRAAN_DB_HOST:-default localhost}
+  host: ${TELETRAAN_DB_HOST:-localhost}
```
sbaogang: I did some simple test, seems this is correct syntax

docker-compose-no-db.yml
```diff
@@ -0,0 +1,28 @@
+version: '2'
```
sbaogang: again, I could not find a better way to not copy paste the docker-compose file

sbaogang:
add also @sjoshi6

haom-pinterest:
LGTM

deploy-board/deploy_board/webapp/security.py
```diff
@@ -151,3 +142,25 @@ def login_authorized(request):
         return HttpResponseRedirect(data['origin_path'])
 
     return HttpResponseRedirect('/')
+
+def logout(request):
+    logger.debug("Logout!")
```
sbaogang: sounds good, will add it in my next PR.

PR #283, opened 2016-12-06
added link to phobos to host details

ruthgrace:
@sjoshi6

### DIFF
--- a/deploy-board/deploy_board/webapp/host_views.py
+++ b/deploy-board/deploy_board/webapp/host_views.py
@@ -18,7 +18,7 @@ from django.views.generic import View
 import logging
 from helpers import environs_helper, agents_helper, autoscaling_groups_helper
 from helpers import environ_hosts_helper, hosts_helper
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -103,6 +103,13 @@ def get_host_details(host_id):
      'Launch Time': launch_time,
      'AMI Id': ami_id,
     }
+    if IS_PINTEREST:
+        from brood.client import Brood
+        cmdb = Brood()
+        host_ip = cmdb.get_query(query="id:" + ami_id,
```
sjoshi6: id: ami-id ? Is this what we want ?  AMI-ID would be same for many hosts ?

sjoshi6: You can just use the above instance object. It will have the config.internal_address you are looking for . It is queried from cmdb.

PR #285, opened 2016-12-07
Refactoring oauth integration to make openid easier

sbaogang:
1. Provided a google openid integration example
2. OAUTH_USERNAME_INFO_KEY to define which field (usually email) to use for username. Allow nested field ( Pinterest return nested userinfo map )
3. provide OAUTH_EXTRACT_USERNAME_FROM_EMAIL to allow extract username from email

@lilida @sjoshi6 @haom-pinterest @yongwen


deploy-board/manage.py
```diff
@@ -39,20 +39,36 @@
...
-    #os.environ.setdefault("OAUTH_CLIENT_TYPE", <CLIENT_TYPE>)
-    #os.environ.setdefault("OAUTH_USER_INFO_URI", <USER_INFO_URL>)
-    #os.environ.setdefault("OAUTH_ACCESS_TOKEN_URL", <ACCESS_TOKEN_URL>)
-    #os.environ.setdefault("OAUTH_AUTHORIZE_URL", <OAUTH_AUTH_URL>)
-    #os.environ.setdefault("OAUTH_DEFAULT_SCOPE", <DEFAULT_SCOP>)
-    # Key for user response, which key is associated with the username in the response. Emails are parsed for prefix.
-    #os.environ.setdefault("OAUTH_USERNAME_INFO_KEY", <USERNAME_KEY>)
+
+    # The client ID that you obtain from the provider
+    #os.environ.setdefault("OAUTH_CLIENT_ID", 'REPLACEME.apps.googleusercontent.com')
```
lilida: It may be more convenient to define a dict like Google_OAUTH_Settings grouped all Google related settings there and set it. Later each OAuth provider have a similar grouped configs

lilida:
r+ with a few small comments

brennentsmith:
If only this was created/merged before we made our own patches on Teletraan to support Google Auth, it would have saved us a ton of time...

I know a lot of people will appreciate having this merged, thank you for putting it together! Super excited to have official support for non-pinadmin auth!

PR #288, opened 2016-12-10
Cluster configuration should use the current cluster image setting in…

lilida:
Fix the bug in cluster configuration the image selector still uses the default one instead of the current cluster one

@haom-pinterest @sbaogang @sjoshi6


haom-pinterest:
R+

PR #291, opened 2016-12-12
use bare phobos link instead of html href

ruthgrace:


deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -106,7 +106,7 @@ def get_host_details(host_id):
     if IS_PINTEREST and PHOBOS_URL:
         host_ip = instance['config']['internal_address']
         phobos_link = PHOBOS_URL + host_ip
```
sjoshi6: Could you add a check of if host_ip is not None then do this. Else display unknown ?

PR #292, opened 2016-12-13
Redirect to environment when cluster is created and fix the bug that …

lilida:
1. After cluster created successfully. Redirect to environment page
2. Remove django template tag in javascript. Use js object to get value.
3. In advanced settings, we don't show cmp_group in UI. We need to add it back when user tries to create a cluster with cmp_base image. Otherwise it will report an error saying cmp group is not set


lilida:
@haom-pinterest @sjoshi6 @sbaogang

haom-pinterest:
LGTM

PR #293, opened 2016-12-13
Jscache

lilida:
deployboard.js is cached in user browser. That causes some caching issue when some change is made and user didn't close browser and refresh. Add  a change time stamp to force js reload. Whenever we have a change in deployboard.js we can update the timestamp. Have tested loading

@haom-pinterest @sbaogang


haom-pinterest:
LGTM.

Thanks for doing this proactively.

PR #296, opened 2016-12-13
Remove isDocker update when updating cluster configuration

lilida:
IsDocker in environ will be deprecated. Changing cluster configuration will update it. Get rid of this code.

@haom-pinterest


haom-pinterest:
LGTM

PR #297, opened 2016-12-14
Realpinleaf

sjoshi6:
@lilida updated UI to launch hosts in PG and terminate all hosts in a group


tylerwowen:
Close as PR is too old.

PR #299, opened 2016-12-15
Use flock in Pinterest as well.

haom-pinterest:
This is to address the issues we ran into where > 1 deploy-agent instances ran after we used pex version of deploy-agent which cmd does not match pgrep.

Have tested on cmp-amp-validator-canary-0a011bca and it worked as expected.


sbaogang:
r+

PR #304, opened 2016-12-16
Append cmp_group when cluster config is changed

lilida:
@haom-pinterest @sjoshi6  Can you take a look at this urgent fix. Right now, if user change cluster configuration, we will remove the cmp_group from user data


sjoshi6:
r+

PR #306, opened 2016-12-19
used proper django templating to check if phobos link exists before u…

ruthgrace:
…sing


ruthgrace:
I merged without review because I figured that the original one is broken, so if this one is broken too it's ok.

PR #309, opened 2016-12-21
Refactor placement selection in cluster creation

lilida:
@haom-pinterest @sbaogang


lilida:
Create cluster CMP mode

screen shot 2016-12-21 at 10 19 28 am


Create cluster Advanced mode

screen shot 2016-12-21 at 10 19 37 am

Edit configuration basic mode

screen shot 2016-12-21 at 10 19 28 am


Edit configuration advanced mode

screen shot 2016-12-21 at 10 19 37 am

haom-pinterest:
Nice. Also, have come up with  a better name about default for placement or we just let users choose since placement is already simple? I am fine with latter choice as well.

lilida:
@haom-pinterest The current decision is just showing our concrete default selection in the select box. The first two screenshots are actually what users will see if they don't set any placements

haom-pinterest:
Just a minor concern is that we need to shuffle concrete default selection otherwise, one of Zone gets less utilized. I guess it is not a big issue.

haom-pinterest:
LGTM!

lilida:
Underlying, it picks the subnet with the max capacity in an AZ every time. We should get a rough balance result

PR #310, opened 2016-12-21
Delete cluster should set env cluster_name to null

lilida:
Right now. Delete cluster only deletes group and remove association with cluster. It doesn't set cluster_name of environment to be null. This is the fix for that change. It also renames 
EnvCapacitys to EnvCapacities


haom-pinterest:
LGTM

PR #313, opened 2016-12-23
try deleting inadvanced class that's only on the select fields withou…

ruthgrace:
…t a help button


lilida:
I don't quite get why you need this change? It serves for purpose of the component

lilida:
@ruthgrace Please revert this change. It will break our existing design

PR #317, opened 2017-01-04
Fix the UI issue that checkbox doesn't work in adv creation page

lilida:
Missing one bind value caused check box doesn't work in new_capacity_adv page

@haom-pinterest


haom-pinterest:
LGTM

PR #318, opened 2017-01-04
Better cluster deterministic logic

lilida:
Use basic_cluster_info to detect if there is a cmp cluster instead of using cluster name in case the cluster has been removed but the environment association still exists.


lilida:
@haom-pinterest

PR #326, opened 2017-01-08
Hide deploy/rollback/hotfix action from the special ngapp2-A and ngapp2-B

yongwen:
This is specific to Pinterest. 
The special env ngapp2-A and B should never able to do deploy etc actions on them because of the A/B deploy with the need to do special handling on varnish. "ngapp2" env should be used to go ngapp deploy/rollback/hotfix.

### DIFF
--- a/deploy-board/deploy_board/templates/environs/env_landing.html
+++ b/deploy-board/deploy_board/templates/environs/env_landing.html
@@ -24,6 +24,7 @@
         <h4 class="panel-title pull-left">Current Deploy</h4>
     </div>
     {% if env|isEnvEnabled %}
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/templates/environs/env_landing.html
```diff
@@ -24,22 +24,26 @@
         <h4 class="panel-title pull-left">Current Deploy</h4>
     </div>
     {% if env|isEnvEnabled %}
+    {% if env.envName != "ngapp2-A" and env.envName != "ngapp2-B" %}
```
lilida: May also add isPinterest check.

PR #329, opened 2017-01-10
Remove image, aws_role from basic cmp settings and change aws_role 

lilida:
AMI Image and aws_role are removed from basic settings.
aws is renamed to iam_role
Fix an issue in adv creation mode, placements was not set if user doesn't click the placement input

Tested with both basic and adv creation page and cluster configuration in both basic and advanced mode


sbaogang:
r+

PR #331, opened 2017-02-06
add remote_trigger for authentication enabled jenkins build triggering

yongwen:
this will prepare for jenkins enabling authentication.  the extra token also works for non authenticated jenkins. When the jenkins auth is enabled, this is required to trigger a job from teletraan.


yongwen:
close this PR, will use teletraan config for this purpose.

PR #346, opened 2017-03-07
fix committer mention in slack message

yongwen:
this is to fix the deploy slack notification message having "..." if the committer id does not exists in slack. such as "@build", or a non pinterest email id.  if the id not exist in slack, this fix will still show the plain id without the mention link.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
@@ -217,7 +217,7 @@ public class DeployHandler {
                     continue;
                 }
                 // TODO hipchat is different, screw it for now
-                authors.add(String.format("<@%s>", commit.getAuthor()));
+                authors.add(String.format("<@%s|%s>", commit.getAuthor(), commit.getAuthor()));
             }
 
             String mentions = Joiner.on(",").join(authors);


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -217,7 +217,7 @@ private String generateMentions(EnvironBean envBean, DeployBean newDeployBean, D
                     continue;
                 }
                 // TODO hipchat is different, screw it for now
-                authors.add(String.format("<@%s>", commit.getAuthor()));
+                authors.add(String.format("<@%s|%s>", commit.getAuthor(), commit.getAuthor()));
```
lilida: Can you give me a bit more info on why we call commit.getAuthor twice here

yongwen:
the format <@comitter_id> will show up in slack as "..." if the commit_id not an existing id in slack. 
the format <@comitter_id|comitter_id> will show up in slack as "@commit_id" without the mention link if the commit_id not exist. 
One example is the auto-commit SHA, the author of the auto-commit is "build@example.com", we extract the author from the email as "build", which is not an existing id in slack. The new format will display this id at least. see https://api.slack.com/docs/message-formatting

lilida:
Get it.

lilida:
LGTM

PR #348, opened 2017-03-08
Double check cmp_group and pinfo_role 

lilida:
when user set different AMI name. Remove cmp_group if users change to a different AMI


lilida:
This address the issue that users switch AMI and we don't auto remove cmp_group or validate if pinfo_role is set

PR #349, opened 2017-03-08
Fixes instance launch failure for disabled ASG on groups UI

sjoshi6:
@lilida

### DIFF
--- a/deploy-board/deploy_board/webapp/group_view.py
+++ b/deploy-board/deploy_board/webapp/group_view.py
@@ -960,6 +960,7 @@ def add_instance(request, group_name):
     placement_group = None
     asg_status = params['asgStatus']
     launch_in_asg = False
+    use_placement_group = False
 
     try:
         # Configure subnet to launch hosts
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -959,30 +959,39 @@ def add_instance(request, group_name):
...
+            # check if custom subnet is specified and custom placement group is in it.
+            if 'customSubnet' in params and 'customPlacementGroup' in params and 'placementGroup' in params:
                 # The check box is ticked and placement group is entered.
                 placement_group = params['placementGroup']
 
-    try:
-        if not launch_in_asg:
-            # Launch static hosts
+        # Check if asg is enabled. Then launch in asg
+        if str(asg_status).upper() == "ENABLED":
```
lilida: If user set a custom subnet and placementgroup, even we have ASG enabled, I think we should still launch outside of the ASG.

sjoshi6: Done

PR #351, opened 2017-03-13
add NEW_AND_FAILED option

sulixu:
this option is for replace cluster feature.

we did have an option to get can_retire=1, however I want to have can_retire!=1 hosts. instead of use <get_all> - <can_retire=1> = <can_retire!=1>. maybe its easier to just provide this new option.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Groups.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Groups.java
@@ -41,7 +41,8 @@ public class Groups {
         ALL,
         RETIRED,
         FAILED,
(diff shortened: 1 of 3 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Groups.java
```diff
@@ -41,7 +41,8 @@
         ALL,
         RETIRED,
         FAILED,
-        RETIRED_AND_FAILED
+        RETIRED_AND_FAILED,
+        NEW_AND_FAILED
```
lilida: For consistency can we name it CANNOT_RETIRED_AND_FAILED?

PR #355, opened 2017-03-15
new api to get serving and normal agent

sulixu:
in the event of replace cluster, I marked the can_retire host's agent to be STOP.  but this agent's deploy_stage is still SERVING_BUILD

next time, when I query for all agent status, I dont want to include deploy_stage = 'SERVING_BUILD' and state = 'STOP' agents.

I purely wants agents whose  deploy_stage = 'SERVING_BUILD' and state = 'NORMAL'  means currently healthy and serving build agent.


lilida:
LGTM

jinruh:
👍

Issue #354, opened 2017-03-14
Feature request: ability to stop/start environments


nathantypanski:
I'm evaluating Teletraan as a replacement for Capistrano. One important feature in our Capistrano deployment system is the ability to run our init scripts to stop/start all hosts in what Teletraan calls an "environment".

It looks like I can roughly emulate this with the following workflow, when hosts are enumerated explicitly in the "Capacity" page:

1. Go to the environment -> all hosts page.
2. Terminate the desired hosts.
3. Do whatever workflow tasks are required while the hosts are down.
4. Go to the environment -> all hosts page.
5. Perform the RESET action on the previously stopped hosts.

This is pretty clunky, however: it downloads a new build and goes through the complete lifecycle for a host (from `PRE_DOWNLOAD` to `POST_RESTART`). Really, all we need to run is `RESTARTING` and `POST_RESTART` to resume a host from the `STOPPED` state. It would be nice if there were a way to do this directly.

lilida:
Have you tried the restart button on the environment page? It should restart services on all hosts

nathantypanski:
Yes, I've tried the restart button - that does not allow me to take an environment offline for a controlled period of time, which is the main thing I'm trying to do here. e.g., bring down an environment while I run migrations. Does that make sense?

lilida:
I see. So the real ask here is a STOP script support as existing PRE_RESTART or RESTART?

nathantypanski:
Sure. Akin to clicking the `RESTART` button, you click a `STOP` button that runs a `STOP` script. Then, maybe reusing the `RESUME` button starts the environment again.

nathantypanski:
One thing I noticed: the service will actually try to use `STOPPING` and `STOPPED` scripts for these states already when they are encountered. In my `deploy-agent.log`, I noticed:

```
INFO:deployd.common.executor:script: /tmp/simplehttpserver/teletraan/STOPPING does not exist.
```

To test this out, I created a `STOPPING` script, and sure enough, it worked! But it doesn't perform interpolation of environment variables like the normal script execution stages. Output from my application's log shows the following, even though I have `PORT` set to `8000` in my **Script Config**:

```
Running STOPPING script.
PORT:
ENV_NAME           simplehttpserver
STAGE_NAME         canary
DEPLOY_ID          M-hE1FdZQVug0BvrwyFroQ
DEPLOY_STE         STOPPING
OPCODE             STOP
DEPLOY_TYP         STOP
[...]

PR #356, opened 2017-03-16
optionally display STOPPING and STOPPED hosts

nathantypanski:
Previously, the environment overview page only showed hosts in the deploy
status that were not STOPPING or STOPPED. The only way to look at them is to
look in "Total" hosts.

First of all, it doesn't make sense to me to to have an agent state that shows
up in the TOTAL but isn't available in the board.

This adds an extra configuration parameter to display STOPPING and STOPPED
hosts. It is disabled by default, in case you have some reason you do not
display those states.

Related: #354

Current behavior:

screen shot 2017-03-15 at 9 24 35 pm

Notice the TOTAL for the *Previous(a2f66f1)* build is `1`, even though none of the state columns actually display 1 host!

With `os.environ.setdefault("DISPLAY_STOPPING_HOSTS", "true")` in `manage.py`, we can see even the stopped hosts:

screen shot 2017-03-15 at 9 24 59 pm

### DIFF
--- a/deploy-board/deploy_board/settings.py
+++ b/deploy-board/deploy_board/settings.py
@@ -194,7 +194,7 @@ SITE_METRICS_CONFIGS = []
 TELETRAAN_SERVICE_HEALTHCHECK_URL = os.getenv("TELETRAAN_SERVICE_HEALTHCHECK_URL", None)
 
 # Show hosts that are STOPPING or STOPPED in the environments page
-DISPLAY_STOPPING_HOSTS = os.getenv("DISPLAY_STOPPING_HOSTS", "false")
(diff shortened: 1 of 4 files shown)


lilida:
+@sbaogang for comment on STOPPING and STOPPED.

sbaogang:
I am all for it - STOPPING/STOPPED was added as new features to gracefully terminate apps, apparently the UI part was not updated.

nathantypanski:
Perhaps we should default it to enabled, then? Is there a reason to hide this?

lilida:
I think it should be enabled by default, Nathan, can you make it be enabled?  Also, I'd like to have your agreement on distributing the code in Apache 2.0 license for me to merge it.  You can reply your agreement just here.

nathantypanski:
I agree to have my code released under the Apache 2.0 license.

PR #358, opened 2017-03-17
get not serving agent count should include those in STOP state

sulixu:
Host can be marked as PENDING_TERMINATING, then its agent state will be immediate marked as STOP. (the whole process takes a few mins)  but during this few mins, we should include those in STOP state as part of `totalDeployingAgents` count.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/dao/AgentDAO.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/dao/AgentDAO.java
@@ -49,7 +49,7 @@ public interface AgentDAO {
 
     void deleteAllById(String hostId) throws Exception;
 
-    // return how many agents are deploying for this env, regardless of deployId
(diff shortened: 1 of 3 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/dao/AgentDAO.java
```diff
@@ -49,7 +49,7 @@
 
     void deleteAllById(String hostId) throws Exception;
 
-    // return how many agents are deploying for this env, regardless of deployId
+    // return how many agents are deploying and agent state is STOP (because agent with STOP state won't deploy in future) for this env, regardless of deployId
```
sbaogang: can we change the comments to something like:

// return how many agents are deploying for this env, including agents whose state is STOP

sbaogang:
r+, thanks for the change.

sulixu:
Thanks for the feedback! I've updated the comment on `countDeployingAgent` API.

PR #361, opened 2017-03-30
cluster replace progress bar

sulixu:
screen shot 2017-03-29 at 8 56 00 pm
The Web UI part will look like the screenshot

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/replace_progress.tmpl
+++ b/deploy-board/deploy_board/templates/clusters/replace_progress.tmpl
@@ -8,35 +8,35 @@
         <th class="col-lg-1">Elapsed</th>
         <th class="col-lg-1">Operator</th>
    </tr>
-   {% if progresses %}
-        {% for replaceProgress in progresses %}
-           <tr>
-                <td>
-                    <span class="replaceToolTip pointer-cursor {{ replaceProgress.state }}"
-                          data-toggle="tooltip" title="replaceProgress.state">
-                    </span>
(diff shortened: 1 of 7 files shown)


lilida:
Will this be always shown on environment page? I suggest moving to cluster capacity page to not bring extra view for non cluster users

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -131,6 +131,44 @@ def update_deploy_progress(request, name, stage):
...
+        host_ids = progress.get('host_ids')
+        succeeded = len(host_ids.split(',')) if host_ids else 0
+        progress_rate = succeeded * 100 / total
+        progress_report = {
+            'state': progress.get('state'),
+            'status': progress.get('status'),
+            'startDate': progress.get('start_time'),
+            'lastUpdateDate': progress.get('last_worked_on'),
+            'progressType': 'success',
+            'progressTip': 'Among total {} hosts, {} successfully replaced and {} are stuck'.format(
```
lilida: Change stuck to pending? I think stuck means it cannot progress

deploy-board/deploy_board/templates/clusters/replace_progress.tmpl
```diff
@@ -0,0 +1,61 @@
...
+           </tr>
+        {% endfor %}
+   {% endif %}
+
+</table>
+</div>
+
+<script>
+
+    $(function () {
```
lilida: Any chance we can consider not mixed template tags in javascript?

PR #367, opened 2017-04-05
replacement progress, replacement history, details, and schedule pages

sulixu:
moved 'replacement history' under 'cluster', same level as 'capacity'

screen shot 2017-04-05 at 4 03 40 pm
screen shot 2017-04-05 at 4 03 53 pm
screen shot 2017-04-05 at 4 04 02 pm


deploy-board/deploy_board/templates/clusters/replace_histories.tmpl
```diff
@@ -0,0 +1,37 @@
+{% load utils %}
+{% load static %}
+<table id="deployHistoryTableId" class="table table-condensed table-striped table-hover">
+    <tr>
+        <th>Start(-08:00)</th>
```
lilida: Change -8:00 to the real time diff?

lilida: Do you think it'd be better to put Active Replacement in capacity page instead of cluster configuration page?

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -503,7 +546,33 @@ <h4 class="panel-title pull-left">Capacity</h4>
...
+                var replaceStartTime = new Date().getTime();
+                var replaceProgressInterval = setIntervalAndExecute(function() {
+                    if(new Date().getTime() - replaceStartTime > 1800000) {  // if more than 30 mins, clear it
+                        console.log('cleared time interval');
+                        clearInterval(replaceProgressInterval);
+                    }
+                    $('#activeClusterReplaceId').load(
+                            '/env/{{ env.envName }}/{{ env.stageName }}/get_replacement_progress/',
+                            function(response, status, xhr) {
+                                if(response == 'There is no on-going replacement.') {
```
lilida: Check content body seems quite hacky. Any way we can make this as an Rest API call and do the rendering on client side?

sulixu: yes you are right, I need to remove this part, maybe dont need to clear the timeout event since after 30 mins, it will be cleared anyway.

deploy-board/deploy_board/templates/clusters/replace_histories.tmpl
sulixu: yea somehow the default page when user clicks on `capacity` is this `cluster configuration page`. (currently pause/cancel/replace cluster buttons are all inside `capacity`
Not sure if I should just move all the actions related with `replacement` outside of `capacity` ?  and add a `replacement` button below `capacity` for all replacements related actions?

PR #369, opened 2017-04-10
when involving the agent reports, we need to consider the environment id

sulixu:
I have 2 hosts (whose agent report as `failed`)

screen shot 2017-04-10 at 4 37 01 pm

current `get successfully serving build hosts` SQL is:
SELECT DISTINCT h.host_id FROM hosts h INNER JOIN agents a ON a.host_id=h.host_id WHERE h.can_retire!=1 AND h.group_name='suli-test-integ' AND h.state not in ('PENDING_TERMINATE','TERMINATING') AND a.deploy_stage = 'SERVING_BUILD' AND a.state = 'NORMAL';
+---------------------+
| host_id             |
+---------------------+
| i-048777bfa7ce30647 |
| i-0538b5a40cb3afb1f |
| i-05a0da0fb8d43847a |
| i-0620aaa554525ecf7 |
| i-08a212050d7841540 |
| i-09200b6976f6b571b |
| i-099f66ae6c3bd8cca |
| i-0a81682a831fac6d0 |
| i-0c2e1c5214cd1e2f6 |
| i-0e894c4610b420a43 |
+---------------------+
10 rows in set (0.01 sec)

which returns 10, but actually it should be 8, like below:

 SELECT DISTINCT h.host_id FROM hosts h INNER JOIN agents a ON a.host_id=h.host_id WHERE h.can_retire!=1 AND h.group_name='suli-test-integ' AND a.env_id = 'WRqlXgiXTnuk3aoBQXDTKQ' AND h.state not in ('PENDING_TERMINATE','TERMINATING') AND a.deploy_stage = 'SERVING_BUILD' AND a.state = 'NORMAL';
+---------------------+
| host_id             |
+---------------------+
| i-09200b6976f6b571b |
| i-0c2e1c5214cd1e2f6 |
| i-048777bfa7ce30647 |
| i-05a0da0fb8d43847a |
| i-0a81682a831fac6d0 |
| i-0620aaa554525ecf7 |
| i-08a212050d7841540 |
| i-0e894c4610b420a43 |
+---------------------+
8 rows in set (0.00 sec)

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/dao/HostDAO.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/dao/HostDAO.java
@@ -65,9 +65,9 @@ public interface HostDAO {
 
     Collection<String> getRetiredHostIdsByGroup(String groupName) throws Exception;
 
-    Collection<String> getRetiredAndFailedHostIdsByGroup(String groupName) throws Exception;
+    Collection<String> getRetiredAndFailedHostIdsByGroup(String groupName, String envId) throws Exception;
 
-    Collection<String> getCanNotRetireAndServingBuildHostIdsByGroup(String groupName) throws Exception;
+    Collection<String> getCanNotRetireAndServingBuildHostIdsByGroup(String groupName, String envId) throws Exception;
 
-    Collection<String> getFailedHostIdsByGroup(String groupName) throws Exception;
+    Collection<String> getFailedHostIdsByGroup(String groupName, String envId) throws Exception;
 }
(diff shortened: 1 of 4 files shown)


lilida:
Agent state is below, is NORMAL the only one we should consider good? Who is using this API? I wonder if the logic change will cause our calculation on deploy goals


```
public enum AgentState {
    NORMAL,
    PAUSED_BY_SYSTEM,
    PAUSED_BY_USER,
    RESET,
    DELETE,
    UNREACHABLE,
    STOP
}
```

sulixu:
yes NORMAL is the only one we consider good in replacement situation.
getFailedHostIdsByGroup
getRetiredAndFailedHostIdsByGroup

are added by original owner for replace cluster feature only.

then I added `getCanNotRetireAndServingBuildHostIdsByGroup` (this one is new) follow the similar format.  so it is safe to say these 3 APIs are only used in replace  cluster.

lilida:
Sounds good

PR #387, opened 2017-04-26
get current build name and commit id by group name

sulixu:


deploy-service/common/src/main/java/com/pinterest/deployservice/dao/BuildDAO.java
```diff
@@ -63,4 +63,6 @@
     List<BuildBean> get(String scmCommit, String buildName, String scmBranch,
         Optional<Integer> pageIndex, Optional<Integer> pageSize, Long before, Long after)
         throws Exception;
+
+    List<BuildBean> getByGroupName(String groupName) throws Exception;
```
lilida: can we rename to getBuildsByGroupName

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBBuildDAOImpl.java
```diff
@@ -94,6 +94,12 @@
             + "AND NOT EXISTS (SELECT 1 FROM deploys WHERE deploys.build_id = builds.build_id) "
             + "ORDER BY publish_date ASC LIMIT ?";
 
+    private static final String GET_BY_GROUP_NAME = "SELECT * FROM builds WHERE build_id IN " +
+        "(SELECT build_id FROM deploys WHERE deploy_id IN " +
+        "(SELECT deploy_id FROM environs WHERE env_id IN" +
```
lilida: So why not using directly build_name and branch in environs table to join with builds name or you are trying to track the case environment builds and branch has been changed? This query may miss the builds that haven't been deployed yet. Does this satisfy your requirement?

PR #392, opened 2017-04-29
Reimplement autopromote

lilida:
The previous causes a few regressions. I feel it has some flaws in the  way of implementation and it is very hard to understand and maintain. So I rewrite another implementation.

Basically, the workflow changes from the old way
1. Find out the right time window
2. Find all builds and deploys in the window

TO
1. Find out all builds and deploys since last build/deploy to now (or now - required delay)
2. For each of candidates, starting from the latest one and check the due on cron based on build and deploy time. If the due time is before now. Deploy it

The difference is how to check cron. To check cron, we need to provide a time to check against.
In the old way, we tried hard to last deploy, last schedule update etc. But it turns out to be very hard to cover all cases. Right now, we directly check the candidate build and deploy against cron. I personally think it is clearer and easier to understand. This gives me more confidence on avoiding regressions


lilida:
One is that now, there may be many candidates to check against. But in practice, I think the number should be small. Also I limit to check only the latest 100 builds or deploys

PR #410, opened 2017-05-15
Add default access_role and make cluster_view fit PEP8

lilida:


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -493,6 +512,7 @@ def get_default_cmp_configs(name, stage):
     config_map['pinfo_environment'] = DEFAULT_CMP_PINFO_ENVIRON
     config_map['pinfo_team'] = 'cloudeng'
     config_map['pinfo_role'] = 'cmp_base'
+    config_map['access_role'] = '{}-{}'.format(name, stage)
```
lilida: This is the only meaningful change

PR #491, opened 2017-09-01
deploy constraint by tag name

sulixu:
A deploy constraint can now be added to a teletraan deploy environment. 

teletraan has a max_parallel number (or percentage) of hosts for the user to set for deployment, to restrict how many hosts can get upgraded with new software concurrently.

However, hosts are chosen randomly, and that might not be desirable (for example, all of the east coast servers are upgraded first). We now introduce a deploy constraint to let user exert more control how hosts are selected.

Currently, we implemented a simple constraint that for a matching host tag, (for example, a host has a tag that represents AWS's autoscaling group name), you can specify a max number that represents how many hosts with the same tag will be deployed concurrently.

Using the scenario mentioned above as an example:

Step 1. First, view all the host tags in this deploy environment , go to:
 /env/<env_name>/<stage_name>/constraint/

You can filter the host tags by tag_name, in our case there are total of 5 tag_names: Autoscaling, aws:autoscaling:groupName, Name, service_mapping and usage_tag.

screen shot 2017-09-27 at 9 19 30 pm

Step 2. Next, create a deploy constraint for a tag_name (in this example, use aws:autoscaling:groupName ), here we specify that teletraan can only deploy a single instance at a time among all host instances with the same autoscaling group tag value.

screen shot 2017-09-27 at 9 19 50 pm

Once this constraint is in place, in our example for host instances with the same autoscaling tag `teletraan-prod-docker`, no more than one host will get deployed concurrently.

Implementation details:

Since querying ec2 instance tags takes time and cannot be done en mass (rate limited by AWS), a teletraan host tag sync worker will be working in the background to sync the tags from ec2 with teletraan database, which also takes into consideration of corner cases like host terminated or getting replaced, pause while tag update etc.

and after tag sync:
screen shot 2017-09-28 at 12 02 52 pm

To monitor tag sync progress, click "View Current Constraint" (under  /env/<env_name>/<stage_name>/constraint)

screen shot 2017-09-27 at 9 31 28 pm
 
Step 3. Once you have defined the constraint, you can start deployment at any time 

under  /env/<env_name>/<stage_name>/host_tags/<tag_name>/, click "Current deployment" or go directly to  /env/<env_name>/<stage_name>/?sortByTag=[tag_name]

then click "Deploy" button at the top right corner.

screen shot 2017-09-18 at 2 41 41 pm

The deployment will start as soon as tag sync is completed.


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -16,33 +16,11 @@
 package com.pinterest.deployservice.handler;
```
lilida: We need to be very careful about the changes here. We may sync next week on the purpose here. We have about 6 DB reads and 1 DB updates for processing every ping request. Not sure how many new DB operations are introduced. But I think we will hit scale limit if we 4x our current load.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -259,6 +278,29 @@ boolean canDeploy(EnvironBean envBean, String host, AgentBean agentBean, String
...
+        try {
+            LOG.debug("verify active agents for host {} rule {}", hostId, ruleId);
+            DeployRuleBean deployRuleBean = deployRuleDAO.getById(ruleId);
+            String tagValue = deployRuleBean.getCondition_key();
+            String tagName = deployRuleBean.getCondition_value();
+
+            long maxParallelWithHostTag = deployRuleBean.getMax_parallel();
+            long totalActiveAgentsWithHostTag = agentDAO.countDeployingAgentWithHostTag(envBean.getEnv_id(), tagName, tagValue);
+            if(totalActiveAgentsWithHostTag >= maxParallelWithHostTag) {
+                LOG.info("already exceed {} for env = {}, host = {}, rule = {}, return false", maxParallelWithHostTag, envId, hostId, deployRuleBean.toString());
```
mingzhaodotname: seems that this line is to long. I also notice that there is no style check in pom file, and maybe we should add that (not necessary in this PR) :-)

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -144,6 +126,27 @@ void updateHosts(PingRequestBean pingRequest) throws Exception {
         }
     }
 
+    void updateHostTag(PingRequestBean pingRequestBean) throws Exception {
+        String hostId = pingRequestBean.getHostId();
+        String hostName = pingRequestBean.getHostName();
+        List<Map<String, String>> tags = pingRequestBean.getTags();
+        if(tags != null) {
```
romans1212notes: maybe better to use (by adding a space):
if (tags != null) {

I also notice similar issues in this PR. I understand that there is no style check in the pom. I'm sorry that it just that my coding habit  prompts me to feel uncomfortable about this :-)

deploy-service/common/src/test/java/com/pinterest/deployservice/db/DBDAOTest.java
```diff
@@ -531,7 +531,7 @@ public void testAgentQueries() throws Exception {
     @Test
     public void testEnvDAO() throws Exception {
 
-        // Test insert
```
mingzhaodotname: Would you want to add some tests for DeployRuleDAO and HostTagDao?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -259,6 +278,29 @@ boolean canDeploy(EnvironBean envBean, String host, AgentBean agentBean, String
...
 
+    boolean canDeployWithRule(String hostId, EnvironBean envBean) throws Exception{
+        String envId = envBean.getEnv_id();
+        String ruleId = envBean.getDeploy_rule_id();
+
+        try {
+            LOG.debug("verify active agents for host {} rule {}", hostId, ruleId);
+            DeployRuleBean deployRuleBean = deployRuleDAO.getById(ruleId);
+            String tagValue = deployRuleBean.getCondition_key();
+            String tagName = deployRuleBean.getCondition_value();
```
mingzhaodotname: Seems that these two lines should be switched:
            String tagName = deployRuleBean.getCondition_key();
            String tagValue = deployRuleBean.getCondition_value();

Is this a sign for a test :-)?

sulixu: thanks for looking at it, the purpose is to add a `constraint` deployment feature to our teletraan deploy. and this PR is to give others an idea of it, after the team approves with the design idea, will get to the test phrase.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
mingzhaodotname: Would https://github.com/pinterest/teletraan/issues/500 be helpful for this case? :-)

deploy-board/deploy_board/webapp/helpers/environ_hosts_helper.py
```diff
@@ -30,3 +30,36 @@ def get_host_by_env_and_hostname(request, env_name, stage_name, host_name):
...
+    return deploy_client.get("/envs/%s/%s/host_tags/%s" % (env_name, stage_name, tag_name),
+                             request.teletraan_user_id.token)
+
+
+def remove_host_tags(request, env_name, stage_name, tag_name):
+    return deploy_client.delete("/envs/%s/%s/host_tags/%s" % (env_name, stage_name, tag_name),
+                                request.teletraan_user_id.token)
+
+
+def get_ec2_host_tags(request, env_name, stage_name, tag_name):
```
lilida: It sounds a little bit weird this is called get_ec2_host_tags with a reload=true parameter

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -259,6 +249,53 @@ boolean canDeploy(EnvironBean envBean, String host, AgentBean agentBean, String
...
+        String envId = envBean.getEnv_id();
+        String constraintId = envBean.getDeploy_constraint_id();
+
+        try {
+            LOG.info("DeployWithConstraint env {}: verify active agents for host {} constraint {}", envId, hostId, constraintId);
+
+            DeployConstraintBean deployConstraintBean = deployConstraintDAO.getById(constraintId);
+            String tagName = deployConstraintBean.getConstraint_key();
+            HostTagBean hostTagBean = hostTagDAO.get(hostId, tagName);
+            if (deployConstraintBean.getState() != TagSyncState.FINISHED) {
```
sbaogang: I think canDeployWithConstraint function should not care or know about the sync process at all. simply look at the host tags:
if no such tag or tag value is undefined, return false
otherwise, check the tag value.

sbaogang: this also implies, deploy does not have to wait for sync explicitly, both deploy and sync could be on going at the same time, concurrently

sulixu: make sense, just removed the update tag sync state logic from pingHandler, which should only happen in tag sync worker

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+    }
+
+    private void processBatch() throws Exception {
+        List<DeployConstraintBean> jobs = deployConstraintDAO.getAllActiveDeployConstraint();
+        if (CollectionUtils.isNotEmpty(jobs)) {
+            Collections.shuffle(jobs);
+            for (DeployConstraintBean job : jobs) {
```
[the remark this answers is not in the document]
sulixu: yes.

PR #502, opened 2017-09-22
Improve exception handing logic of deploy-agent cleanup

brennentsmith:
Currently, the operations to delete the actual build directory and the tarball are considered within a single exception handler. 

We have seen cases on very large directories where `shutil` will throw an error and thus exit the exception statement. As a result, the archive is left behind as well, which is not optimal. This has also caused issue as the error thrown is not always an `OSError` exception, so the entire application halts. 

This PR ensures that each operation is performed in an isolated exception handler to prevent the `shutil` operation from skipping the the gzip removal.

Apologies for the git mess - we were up all night working on another project and happened to detect this issue. 

Thank you @ellezv for the help as well!


deploy-agent/deployd/common/helper.py
```diff
@@ -65,17 +65,26 @@ def get_stale_builds(build_timestamps, num_builds_to_retain=2):
...
             log.error(e)
 
         try:
+            # Remove build directory from disk
             shutil.rmtree(os.path.join(base_dir, builds))
+        except BaseException as e::
+            # Catch base exception class, as there's a mulitude of reasons a rmtree can fail
+            log.error(e)
+
+        try:
```
mingzhaodotname: Since we only log exceptions in this function, how would you think of merging all the try's into one try, and simplify the exception catching process?

deploy-agent/deployd/common/helper.py
```diff
@@ -65,17 +65,26 @@ def get_stale_builds(build_timestamps, num_builds_to_retain=2):
...
                 os.remove(staged_file)
         except OSError as e:
             log.error(e)
 
         try:
+            # Remove build directory from disk
             shutil.rmtree(os.path.join(base_dir, builds))
+        except BaseException as e::
+            # Catch base exception class, as there's a mulitude of reasons a rmtree can fail
+            log.error(e)
```
mingzhaodotname: How about using log.exception, see:
https://stackoverflow.com/questions/5191830/best-way-to-log-a-python-exception

deploy-agent/deployd/common/helper.py
brennentsmith: Correct me if I'm wrong, but if we put all of the logic into a single exception handler, if the first function fails, the rest will not be executed - thus if trying to remove the `<id>.staging` file fails, none of the other cleanup functions will execute. 

Yes, having a bunch of exception handlers isn't pretty, but it ensures that an early failure won't leave even more artifacts laying around.

deploy-agent/deployd/common/helper.py
brennentsmith: 👍 Will update PR

deploy-agent/deployd/common/helper.py
ellezv: I would also be in favor of catching earlier OSErrors and then any rmtree errors separately as to actually execute the remove if that if statement is reached.

If casting such a wide net, being redundant/explicit is not necessarily a bad thing.

deploy-agent/deployd/common/helper.py
ellezv: typo in comment line 83: multitude

brennentsmith: 2 hours of sleep will do that - fixing

deploy-agent/deployd/common/helper.py
romans1212notes: You're right that we want to catch all the errors. One option to make it clean is to define a list of functions, and then iterate on those function with one try-and-except, e.g.
for func in func_list:
  try: 
     func()
  except 
    log.exception

lilida:
LGTM

deploy-agent/deployd/common/helper.py
mingzhaodotname: Yes, you're right that we want to catch all the exceptions and give all the functions a chance to run.

One way to make it work is using a list of functions, e.g.
func_list = [func1, func2, func3]
for func in func_list:
    try: 
        func
    except IOError:
        xxx
    except Exception:
        xxx

Not sure whether it's worth it :-)

deploy-agent/deployd/common/helper.py
```diff
@@ -65,17 +65,26 @@ def get_stale_builds(build_timestamps, num_builds_to_retain=2):
...
             staged_file = os.path.join(base_dir, '{}.staged'.format(builds))
             if os.path.exists(staged_file):
                 os.remove(staged_file)
         except OSError as e:
             log.error(e)
 
         try:
+            # Remove build directory from disk
             shutil.rmtree(os.path.join(base_dir, builds))
+        except BaseException as e::
```
mingzhaodotname: BaseException seems too wide since it also include SystemExit, KeyboardInterrupt. Otherwise, it's hard to stop exit the code. A better option might be just Exception.

PR #509, opened 2017-10-02
Builds clean up now is done per environment

lilida:
The build clean now looks at all envname-xxx.yy files and sort them by mtime on the file. It then tries to extract build from it and follow the old clean up code


haom-pinterest:
Nice!

PR #511, opened 2017-10-04
add mark build as bad in rollback dialogue and never autopromote buil…

sulixu:
…d if its bad

### DIFF
--- a/deploy-board/deploy_board/templates/deploys/confirm_rollback.tmpl
+++ b/deploy-board/deploy_board/templates/deploys/confirm_rollback.tmpl
@@ -26,7 +26,7 @@
                     <input type="radio" name="mark_build_as_bad" value="on" checked>On
                 </label>
                 <input id="confirmCurrentBuildId" type="hidden" name="toBeMarkedBuildId" value="" />
-                <label class="radio-inline">Mark Build As bad</label>
+                <label class="radio-inline">Mark Build As bad (bad build will not be auto-deployed) </label>
             </div>
         </div>
 
--- a/deploy-board/deploy_board/templates/deploys/confirm_rollback.tmpl
+++ b/deploy-board/deploy_board/templates/deploys/confirm_rollback.tmpl
@@ -12,11 +12,24 @@
         Are you sure to rollback to <a id="confirmCommitLinkId" href="#REPLACE">REPLACE</a>?
         {% csrf_token %}
         <div class="row">
-        <div class="form-group">
-          <input class="form-control" name="description" type="text" required="true"
-                 placeholder="Rollback reason (required)" value=""/>
+            <div class="form-group">
+              <input class="form-control" name="description" type="text" required="true"
+                     placeholder="Rollback reason (required)" value=""/>
+            </div>
         </div>
+        <div class="row">
+            <div class="form-group" align="center">
+                <label class="radio-inline">
+                    <input type="radio" name="mark_build_as_bad" value="off" >Off
+                </label>
+                <label class="radio-inline">
+                    <input type="radio" name="mark_build_as_bad" value="on" checked>On
+                </label>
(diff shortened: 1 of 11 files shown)


deploy-board/deploy_board/templates/deploys/confirm_rollback.tmpl
```diff
@@ -12,11 +12,24 @@
...
+        <div class="row">
+            <div class="form-group" align="center">
+                <label class="radio-inline">
+                    <input type="radio" name="mark_build_as_bad" value="off" >Off
+                </label>
+                <label class="radio-inline">
+                    <input type="radio" name="mark_build_as_bad" value="on" checked>On
+                </label>
+                <input id="confirmCurrentBuildId" type="hidden" name="toBeMarkedBuildId" value="" />
+                <label class="radio-inline">Mark Build As bad</label>
```
lilida: Can we add more description that bad build will not be auto-deployed for giving more context to users

deploy-board/deploy_board/webapp/common.py
```diff
@@ -135,6 +135,17 @@ def restart(request, name, stage):
 def rollback_to(request, name, stage, deploy_id):
     query_dict = request.POST
     desc = query_dict.get('description', None)
+    mark_build_as_bad = query_dict.get('mark_build_as_bad', True)
```
lilida: suggest mark_build_as_bad = true if query_dict.get('mark_build_as_bad', 'on') == 'on else False and removing the  if statements below. It is not ideal to mix the type 'off' and True on the same variable.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AutoPromoter.java
```diff
@@ -278,8 +269,14 @@ public void promoteBuild(EnvironBean currEnvBean, DeployBean currDeployBean, int
             currEnvBean.getEnv_name());
         if (result.getResult() == PromoteResult.ResultCode.PromoteBuild &&
             StringUtils.isNotEmpty(result.getPromotedBuild())) {
-            safePromote(null, result.getPromotedBuild(), Constants.BUILD_STAGE, currDeployBean,
-                currEnvBean);
+            TagBean buildTagBean = getBuildTag(result.getPromotedBuild());
```
lilida: But it can be tricky here. Say we have
build1, bad
build2, good
build3, current
We still want to promote to build2 as we are running on build3.  But if we change the code here, it will not promote to build2 and the candidate is always build1. I think it'd be better to change the SQL select for candidates build.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AutoPromoter.java
```diff
@@ -481,29 +470,75 @@ public Date getScheduledCheckDueTime(long start_date, String cronExpressionStrin
...
+        List<DeployBean> availableDeploys = deployDAO.getAcceptedDeploys(envId, interval, size);
+        if(!availableDeploys.isEmpty()) {
+            List<String> availableBuildIds = new ArrayList<String>();
+            Map<String, DeployBean> buildId2DeployBean = new HashMap<String, DeployBean>();
+
+            for(DeployBean deployBean: availableDeploys) {
+                String buildId = deployBean.getBuild_id();
+                if(StringUtils.isNotEmpty(buildId)) {
+                    buildId2DeployBean.put(buildId, deployBean);
+                    availableBuildIds.add(deployBean.getBuild_id());
```
lilida: What if we have multiple deploys with the same build? We may add duplicated build id here and some accept deploy won't be int he list. Maybe we should just create a build_id hashmap, then get tags and mark the bad. Then loop through the available deploys and only copy good build to another list.

sulixu: good point, just updated

PR #512, opened 2017-10-06
some weird case would cause build id to be empty, and its not a valid…

sulixu:
… build
screen shot 2017-10-06 at 4 09 36 pm

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
@@ -30,6 +30,7 @@ import com.pinterest.teletraan.security.Authorizer;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -189,6 +190,9 @@ public Response create(
         EnvironBean envBean = Utils.getEnvStage(environDAO, envName, stageName);
         authorizer.authorize(sc, new Resource(envBean.getEnv_name(), Resource.Type.ENV), Role.OPERATOR);
         String operator = sc.getUserPrincipal().getName();
+        if(StringUtils.isEmpty(buildId)) {
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, "Build id can not be empty.");
```
jinruh: Will deploy-board catch this exception?

sulixu: i removed the `submit` button from the UI if the build id is empty, so at least from deploy-board, this API wont be triggered. adding this exception is to make sure when clients call `create deploy` api, the buildId is set appropriately.

PR #515, opened 2017-10-18
Make sure curl returns on non-0 exit code on non-200 responses. Prior…

smeder:
… to this change a 404 would be counted a success.


deploy-agent/deployd/download/http_download_helper.py
```diff
@@ -25,7 +25,7 @@
 class HTTPDownloadHelper(DownloadHelper):
 
     def _download_files(self, local_full_fn):
-        download_cmd = ['curl', '-o', local_full_fn, '-ks', self._url]
+        download_cmd = ['curl', '-o', local_full_fn, '-fks', self._url]
```
lilida: Please also bump the verison in setup.py

smeder: Done

PR #519, opened 2017-10-25
Adding timestamp logging to deploy scripts.

TheLat:
We currently don't log timestamps on the deploy scripts and that makes it very difficult to identify pain points.  This change causes all logs to be prepended with the current time.

### DIFF
--- a/deploy-agent/deployd/common/caller.py
+++ b/deploy-agent/deployd/common/caller.py
@@ -27,13 +27,14 @@ class Caller(object):
     @staticmethod
     def call_and_log(cmd, **kwargs):
         output = ""
+        start = time.time()
         try:
             process = subprocess.Popen(cmd, stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


deploy-agent/deployd/common/caller.py
```diff
@@ -25,10 +26,15 @@ def __init__(self):
...
     def call_and_log(cmd, **kwargs):
+        output = ""
         try:
             process = subprocess.Popen(cmd, stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE, **kwargs)
-            output, error = process.communicate()
+            while process.poll() is None:
+                line = process.stdout.readline()
+                if line:
+                    output = output + "[%s]" % time.time() + line
```
lilida: With this change, the time is not the actual time that the log was created but the current time. I think the right approach is to let the subprocess write timestamp to its stdout

Issue #517, opened 2017-10-18
The post-deploy webhook is currently only called on successful deploys


smeder:
Right now the post-deploy webhook is only called on successful deploys and the deploy state is not updated before calling the webhook. This causes the deploy state to always be RUNNING when the webhook is being called, so the deploy state variable substitution is currently meaningless.

lilida:
@sbaogang While looking at the code, I think there are some inconsistencies here on how we handle the deploy event and webhooks.


1. We notify webhook when succeeding the deploy

`
   if (succeeded * 10000 >= sucThreshold * total) {
            LOG.debug("Propose deploy {} as SUCCEEDING since {} agents are succeeded.", deployId, succeeded);
            if (deployBean.getSuc_date() == null) {
                newDeployBean.setSuc_date(System.currentTimeMillis());

                // Submit post-webhooks, if exists
                EnvWebHookBean webhooks = dataHandler.getDataById(envBean.getWebhooks_config_id(), WebhookDataFactory.class);
   
`

2.  In the outside function internalTransition, we trigger
`
        if (shouldSendFinishMessage(state, newPartialDeployBean.getState(), deployBean.getSuc_date())) {
            jobPool.submit(new FinishNotifyJob(envBean, deployBean, newPartialDeployBean, deployBean.getBuild_id()));
        }
`

Should these two be in the same place? Should we send the notificiations for both succeeding and failing state?

PR #520, opened 2017-10-26
Postdeploywebhook moves the ShouldSendFinishMessage

lilida:
This is a fix trying to address the issue below:
https://github.com/pinterest/teletraan/issues/517

It tries to consolidate when to send a deploy finish message. FinishNotifyJob was used to send slack notification today

### DIFF
--- a/deploy-agent/deployd/common/helper.py
+++ b/deploy-agent/deployd/common/helper.py
@@ -45,7 +45,6 @@ class Helper(object):
         finally:
             return builds
 
-
     @staticmethod
     def get_build_id(filename, env_name):
         """
--- a/deploy-agent/deployd/common/helper.py
+++ b/deploy-agent/deployd/common/helper.py
@@ -45,6 +45,7 @@ class Helper(object):
         finally:
             return builds
 
+
     @staticmethod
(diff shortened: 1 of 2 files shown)


lilida:
There are five users and I will follow up

sbaogang:
Sorry for the late reply - reading the current code made me believe the followings:

- Invoking webhook is one-time only, right after deploy succeeds the first time; sending notification on the other hand is a repeating work happens for every "transition" call. 

For example, in the auto-scaling case where hosts keep getting terminated and new machine keep getting added, let us assume also all the newly added hosts having issues with deploy in the beginning (maybe can not warm up in time) but later on succeed. In this case, we would send notify in the pattern of FAILING->SUCCEEDING->FAILING->SUCCEEDING... If we change the behavior, we would call the hook over and over again, I do not think this is most people anticipated.

webhook checking is closely related to the check of (success/total>threshold && sucdate==null); and send notify is related to STATE changes, not care about success, fail etc. That might be the reason why the code was written in the current way.

But regardless where these logic live, I think calling webhook once makes more sense to most people, in most case. I would not recommend to change such behavior.
-

PR #529, opened 2017-11-13
Alert webhook

lilida:
This is for adding a web API to allow external alerting system setting up a webhook to trigger auto rollback or mark build as bad. Still polishing and testing.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
@@ -1,0 +1,16 @@
+package com.pinterest.deployservice.alerts;
+
+import com.pinterest.deployservice.bean.DeployBean;
+import com.pinterest.deployservice.bean.EnvironBean;
+
+
+/**
+ * AlertAction is the base class for an action for an alert
+ */
+public abstract class AlertAction {
+  public abstract Object perform(EnvironBean environ,
+                                 DeployBean lastDeploy,
+                                 int actionWindowInSeconds,
+                                 AlertContext context,
+                                 String operator) throws Exception;
+}
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
@@ -8,9 +8,10 @@ import com.pinterest.deployservice.bean.EnvironBean;
  * AlertAction is the base class for an action for an alert
  */
 public abstract class AlertAction {
-  public abstract Object perform(EnvironBean environ,
+
+  public abstract Object perform(AlertContext context,
+                                 EnvironBean environ,
                                  DeployBean lastDeploy,
                                  int actionWindowInSeconds,
-                                 AlertContext context,
                                  String operator) throws Exception;
 }
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AutoRollbackAction.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AutoRollbackAction.java
@@ -18,6 +18,8 @@ public class AutoRollbackAction extends AlertAction {
 
   public final static Logger LOG = LoggerFactory.getLogger(AutoRollbackAction.class);
 
+  //Maximum time looks back. May make it pass in from webhook later
+  private final static int MaxLookbackDays = 30;
 
   @Override
   public Object perform(EnvironBean environ, DeployBean lastDeploy, int actionWindowInSeconds,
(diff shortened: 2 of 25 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
```diff
@@ -0,0 +1,16 @@
...
+
+
+/**
+ * AlertAction is the base class for an action for an alert
+ */
+public abstract class AlertAction {
+  public abstract Object perform(EnvironBean environ,
+                                 DeployBean lastDeploy,
+                                 int actionWindowInSeconds,
+                                 AlertContext context,
```
smeder: Nit: can we put context first? I find it easier to have the more generic arguments first

deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
```diff
@@ -0,0 +1,16 @@
...
+import com.pinterest.deployservice.bean.DeployBean;
+import com.pinterest.deployservice.bean.EnvironBean;
+
+
+/**
+ * AlertAction is the base class for an action for an alert
+ */
+public abstract class AlertAction {
+  public abstract Object perform(EnvironBean environ,
+                                 DeployBean lastDeploy,
```
smeder: What happens if there were multiple deploys in the window?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -448,6 +454,20 @@ void internalTransition(String deployId, EnvironBean envBean) throws Exception {
...
         if (shouldSendFinishMessage(state, newPartialDeployBean.getState(), deployBean.getSuc_date())) {
             jobPool.submit(new FinishNotifyJob(envBean, deployBean, newPartialDeployBean, deployBean.getBuild_id()));
+
+            if (newPartialDeployBean.getState() == DeployState.SUCCEEDING) {
+                //Set SucDate to avoid resent SUCCEEDING
+                newPartialDeployBean.setSuc_date(System.currentTimeMillis());
+            }
+
+            // Submit post-webhooks, if exists
+            EnvWebHookBean
```
smeder: Did you mean to include this change in this PR?

deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
lilida: Rollback goes to the last known good deploy before the window so all the deploys in the window are rolled back. However, markbadbuild only marks the build of the latest deploy. It is a bit tricky

deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
lilida: Done

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
lilida: Removed

deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AutoRollbackAction.java
```diff
@@ -0,0 +1,53 @@
...
+    // the now-window
+    List<DeployBean>
+        candidates =
+        context.getDeployDAO().getAcceptedDeploys(environ.getEnv_id(),
+            new Interval(DateTime.now().minusDays(MaxLookbackDays),
+                DateTime.now().minusSeconds(actionWindowInSeconds)),
+            1);
+    if (candidates.size() > 0) {
+      try {
+        LOG.info("AutoRollback to {}", candidates.get(0).getDeploy_id());
```
sulixu: Could we add env_id in log as well,  easier to debug

sulixu: do we need to filter out deploy with `bad build` from the db result of `getAcceptedDeploys`? Even though `getAcceptedDeploys` only includes acc_status='ACCEPTED' deploys.  I am not sure if there is any possibility that a build is marked as bad, but acc_status is ACCEPTED.

lilida: This is a good point. I plan to move most of logic from AutoPromoter to DeployHandler. This can ensure we share the same logic on handling deployment selections

deploy-service/common/src/main/java/com/pinterest/deployservice/alerts/AlertAction.java
smeder: I guess the problem I see is with auto deploy. If I want to combine rollback & mark build as bad I would really want to mark all builds deployed in the window as bad. Wouldn't auto-deploy otherwise pick a build in the window if there happen to be multiple?

lilida: @smeder I see. This is an issue if (waiting to promote period <window size). I can make markbadbuild to label the build of the first deploy in the window.  In general, I think people configure a workflow like (deploy a build, waiting x minutes with no alerts, promote)  and x is larger than the window size in general.

smeder: Ok - I think I can live within those constraints.

PR #544, opened 2017-12-13
Autopromoter may be stuck forever if trying to rollback in an environment with no successful deploys 

lilida:
Fix the bug that autopromoter entering the infinite loop if there are no succeeded deployments. The old code has a bug that not set the page index which is pretty bad. Surprisingly, we just saw it now as it looks like we always found one successful deploy in the first 10. Also adding some guardian for maximum check latest 10000 deploys.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
@@ -395,7 +395,8 @@ public class DeployHandler implements DeployHandlerInterface{
         filterBean.setPageIndex(index);
         filterBean.setPageSize(size);
         int maxPages = 50; //This makes us check at most 5000 deploys
-        while (maxPages-- > 0) {
+        int tocheckPages = maxPages;
+        while (tocheckPages-- > 0) {
             DeployQueryFilter filter = new DeployQueryFilter(filterBean);
             DeployQueryResultBean resultBean = deployDAO.getAllDeploys(filter);
             if (resultBean.getTotal() < 1) {
@@ -411,7 +412,7 @@ public class DeployHandler implements DeployHandlerInterface{
             index += 1;
             filterBean.setPageIndex(index);
         }
-        LOG.warn("Latest 100000 deploys are all failed for {}. Give up", envBean.getEnv_id());
+        LOG.warn("Latest {} deploys are all failed for {}. Give up", size*maxPages, envBean.getEnv_id());
         return null;
 
     }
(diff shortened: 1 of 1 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -389,12 +389,13 @@ public String promote(EnvironBean envBean, String fromDeployId, String descripti
...
         int index = 1;
-        int size = 10;
+        int size = 100;
         DeployFilterBean filterBean = new DeployFilterBean();
         filterBean.setEnvIds(Arrays.asList(envBean.getEnv_id()));
         filterBean.setPageIndex(index);
         filterBean.setPageSize(size);
-        while (true) {
+        int maxPages = 100; //This makes us check at most 10000 deploys
+        while (maxPages-->0) {
```
smeder: Nit: Add spaces around '>' - currently not very readable.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -389,12 +389,13 @@ public String promote(EnvironBean envBean, String fromDeployId, String descripti
...
     DeployBean getLastSucceededDeploy(EnvironBean envBean) throws Exception {
         int index = 1;
-        int size = 10;
+        int size = 100;
         DeployFilterBean filterBean = new DeployFilterBean();
         filterBean.setEnvIds(Arrays.asList(envBean.getEnv_id()));
         filterBean.setPageIndex(index);
         filterBean.setPageSize(size);
-        while (true) {
+        int maxPages = 100; //This makes us check at most 10000 deploys
```
smeder: 10000 seems really large. Do we actually need that or can we do say 1000?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -407,7 +408,11 @@ DeployBean getLastSucceededDeploy(EnvironBean envBean) throws Exception {
                 }
             }
             index += 1;
+            filterBean.setPageIndex(index);
         }
+        LOG.warn("Latest 100000 deploys are all failed for {}. Give up", envBean.getEnv_id());
```
smeder: Maybe change to 

LOG.warn("Latest {} deploys are all failed for {}. Give up", maxPages * size, envBean.getEnv_id());

smeder:
LGTM

PR #545, opened 2017-12-13
Handle large number of build tags more efficient

lilida:
In some scenario, we have users apply good tags on every build and it may add bad tag to the same build later. The change tries to make this handling be more sane.
1. When computing a build is good or bad, only look at the most recent 1000 tags
2. If one commit have multiple tags, use the latest created one


lilida:
The test changes include a lot of reformatting. Only meaningful is changing the mock method as we change the call TagDAO

PR #552, opened 2018-01-24
change HostBean can_retire type (db schema is already TINYINT(1)), an…

sulixu:
1. change HostBean can_retire type (db schema is already TINYINT(1))
2. the UI can not rely `cluster capacity count` on capacity field of clusterBean, need AWS query for ASG summary.


sbaogang:
LGTM

PR #564, opened 2018-02-23
availability_zone aware deploy

sulixu:
This commit extends the deploy constraint functionality from tag-aware to availability_zone-aware deploy, though different in:
 tag-aware choose X number of hosts from each tag groups, and deploy in parallel.
 az-aware choose X number of hosts from one group only, and deploy group by group.

screen shot 2018-02-22 at 4 37 12 pm
 
in this example deploy in order of:  us-east-1c -> us-east-1a -> us-east-1d


jinruh:
likestamp

PR #567, opened 2018-02-28
availability_zone aware deploy

sulixu:
group deploy used to be:
 choose X number of hosts from each tag groups, and deploy them in parallel with X number of hosts at a time. (by default inclusive is false)

now group deploy can be:
choose X number of hosts from one group only, and deploy group by group. (can achieve this by set inclusive to true)

1. the availability_zone group names are already integrated with host tags. (provided by our CMDB)
2. group will get deployed in alphabetical order, for example:
us-east-1a -> us-east-1b -> us-east-1c -> us-east-1d


deploy-service/common/src/main/java/com/pinterest/deployservice/bean/DeployConstraintBean.java
```diff
@@ -9,6 +9,7 @@
  * constraint_id     VARCHAR(22)         NOT NULL,
  * constraint_key    VARCHAR(64)         NOT NULL
  * max_parallel      BIGINT              NOT NULL,
+ * inclusive         TINYINT(1)          NOT NULL DEFAULT 0,
```
sbaogang: instead of support yet another special constraint, I wonder if we can come up with a more generic solution.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -311,6 +311,22 @@ boolean canDeployWithConstraint(String hostId, EnvironBean envBean) throws Excep
...
             }
             String tagValue = hostTagBean.getTag_value();
+            if(deployConstraintBean.getInclusive()) {
+                // if it is inclusive deploy, needs to check pre-requisite tags when pre-requisite tag hosts are all done, then it becomes its turn
+                List<String> prerequisiteTagValues = hostTagDAO.getAllPrerequisiteTagValuesByEnvIdAndTagName(envBean.getEnv_id(), tagName, tagValue);
+                if(prerequisiteTagValues!= null && prerequisiteTagValues.size() > 0) {
+                    // if there is any pre-requisite tag, the hosts tagged by this pre-requisite tag should deploy first
+                    long totalExistingHostsWithPrerequisiteTags = hostTagDAO.countHostsByEnvIdAndTags(envBean.getEnv_id(), tagName, prerequisiteTagValues);
+                    long totalFinishedAgentsWithPrerequisiteTags = agentDAO.countFinishedAgentsByDeployWithHostTags(envBean.getEnv_id(), envBean.getDeploy_id(), tagName, prerequisiteTagValues);
+                    if(totalFinishedAgentsWithPrerequisiteTags < totalExistingHostsWithPrerequisiteTags) {
```
sbaogang: have we considered failed hosts in this case?

sbaogang:
can we find a time to discuss this change in details?

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/DeployConstraintBean.java
sulixu: good point, just updated into an enum field

PR #573, opened 2018-03-23
host can be chose to another stage, exclude such host for current stage deployment

sulixu:
we found a bug in AZ aware deploy that prod stage deployment waits for the host that belong to canary stage. (waiting forever for a host that doesnt belong to current deployment stage)

this fix excludes such hosts, and sync host tags with the filtered result

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/dao/HostDAO.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/dao/HostDAO.java
@@ -72,4 +72,6 @@ public interface HostDAO {
     Collection<String> getNewHostIdsByGroup(String groupName) throws Exception;
 
     Collection<String> getFailedHostIdsByGroup(String groupName) throws Exception;
+
+    Collection<String> getExcludedHostsFromEnvByEnvId(String envId) throws Exception;
 }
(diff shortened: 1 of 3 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
```diff
@@ -42,6 +42,19 @@ private void processEachEnvironConstraint(DeployConstraintBean bean) throws Exce
         String tagName = bean.getConstraint_key();
         String envId = environBean.getEnv_id();
         Collection<HostBean> hostBeans = hostDAO.getHostsByEnvId(envId);
+        Collection<String> excludedHostNames = hostDAO.getExcludedHostsFromEnvByEnvId(envId);
```
lilida: Could you give a bit more description here? Are we just excluding syncing tags? How do we handle the case that host has been added and removed? Do we remove the previous tags?

deploy-service/common/src/main/java/com/pinterest/deployservice/dao/HostDAO.java
```diff
@@ -72,4 +72,6 @@
     Collection<String> getNewHostIdsByGroup(String groupName) throws Exception;
 
     Collection<String> getFailedHostIdsByGroup(String groupName) throws Exception;
+
+    Collection<String> getExcludedHostsFromEnvByEnvId(String envId) throws Exception;
```
lilida: I wonder if we should follow the pattern how we do deploy. Which means we rely on deployagent report groups instead of changing the logic here.

PR #596, opened 2018-06-05
Revert "Show provider name for instances"

tobicodes:
Reverts pinterest/teletraan#595


yongwen:
Tobi, do we still need this revert? what was the issue?

PR #598, opened 2018-07-09
Monorepo paths

chrislloyd:
Teletraan pulls lots of information from Phabricator to make deploys more useful (branch, commit range etc.). I'd like to add a "path" property to environments so that we can effectively get this information for builds in a monorepo.

The goal here is to be able to scope a build to a directory so that calls to "contributors" doesn't return _all_ contributors but just the subset that touched the files in the build.

**Test plan:**

screen shot 2018-07-16 at 3 10 48 pm

```bash
#!/usr/bin/env bash
REPO="P"
START_SHA="f2be1448a5c4294b4dbcc20377f1c33e410138b7"
END_SHA="90be8b304f9719edb15a8d3d2155457461b2becf"
SIZE="500"
REPO_PATH="webapp/*"

curl -s "localhost:8080/v1/commits?repo=$REPO&startSha=$START_SHA&endSha=$END_SHA&size=$SIZE&path=$PATH" \
  --header "Accept: application/json"
```


sbaogang:
LGTM, thanks for the changes!

PR #599, opened 2018-08-13
extend the metrics config to define a dashboard url for a env/stage

yongwen:
current metrics config only support Gauge metrics. There is often need to supply an (external) dashboard url for a service (stage) that service owner can be easily click to. Once the user-defined dashboard url is specified in the metrics config, the "Metrics" section on the main stage page will show the "Dashboard" button along side with other gauge metrics defined.

### DIFF
--- a/deploy-board/deploy_board/templates/configs/metrics_config.tmpl
+++ b/deploy-board/deploy_board/templates/configs/metrics_config.tmpl
@@ -27,6 +27,8 @@
         {% csrf_token %}
 
         <div id="newMetricSpot_{{ metric.title|slugify }}" class="panel-default"></div>
+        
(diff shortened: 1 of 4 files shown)


deploy-board/deploy_board/templates/configs/metrics_config.tmpl
```diff
@@ -143,6 +163,11 @@
     var validUrl = false;
     var newEntry;
     $(function () {
+        $("select[name='metricsType']").change(function () {
+                document.getElementById("newMetricLabel").style.visibility= this.value=='Gauge' ? 'visible' : 'hidden';
```
lilida: Use jquery selector rather than document getElementById for consistency. Or even better, use vuejs to binding the field

PR #600, opened 2018-08-13
Adding timeout to prevent webhook failures

TheLat:
If a webhook never returns, it will hang forever and all subsequent webhooks will not be called.  This resolves the issue.


deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -73,7 +73,7 @@ private String internalCall(String url, String method, String payload, Map<Strin
                 conn = (HttpURLConnection) urlObj.openConnection();
                 conn.setRequestMethod(method);
                 conn.setRequestProperty("Accept-Charset", "UTF-8");
-                conn.setConnectTimeout(15000)
+                conn.setConnectTimeout(15000);
```
yongwen: please make it as a constant, even better, a configurable variable in config file.  and specify the unit of this number in comment. I assume this amount is in milli-second.

Please also do some manual testing to see what will actually happen if this timeout is reached.

TheLat:
Manually tested and verified to work.

PR #610, opened 2018-09-27
Revert "Create env/consolev2"

tobicodes:
Reverts pinterest/teletraan#609


tobicodes:
Commit #609 works as expected so no need to revert cc @yongwen

PR #613, opened 2018-10-08
Add gpg support to deploy-agent

brennentsmith:
Currently, build artifacts are stored raw and do not have any signing methods. This means that if a nefarious third party or evil employee gained access to our S3 bucket, they would have the following: 

1. They could view all source code in one place, bypassing any source control ACL groupings that exist. We don't store secrets in our codebase, but it would still be a very bad day/week/month if our source code was leaked. 
2. More importantly, if the third party gained write access, they would be able to modify the artifact. This would be a direct injection point to add code to grab session identities, credit card details, or inject any code (client/serverside) into our stack. This would be a maximum severity exploit and theoretically could take down any service. 

The solution to this issue is to encrypt and sign our code artifacts. One industry standard method is via GPG. 

This PR adds the ability to detect if the archive is contained within a GPG container, and decrypt it before decompressing the archive. This would ensure that:

1. If a user gained access to our archive storage location, they would not be able to view the code unless they had the "deploy-agent side" GPG private key. 
2. If a user gained write access to our archive storage location, there would be no way for them to modify the archives unless they are able to gain a copy of the "build service" GPG private key. 

Overall, this PR ensures that code cannot be tampered with after the build process and is encrypted end to end from build container to end destination.

The PR does not interact with GPG beyond the decrypting process - to maximize portability, it is up to the surrounding orchestration and secret handlers to set up the GPG keychains correctly.

### DIFF
--- a/deploy-agent/deployd/download/downloader.py
+++ b/deploy-agent/deployd/download/downloader.py
@@ -82,7 +82,7 @@ class Downloader(object):
                 dest_full_fn = local_full_fn[:(len(extension) * -1)] + innerExtension
 
                 # decrypt gpg archive
-                gpgHelper.decryptFile(local_full_fn, dest_full_fn)
+                status = gpgHelper.decryptFile(local_full_fn, dest_full_fn)
                 
                 if status != Status.SUCCEEDED:
                     # die if we hit a decryption or signing error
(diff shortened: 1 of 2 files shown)


deploy-agent/deployd/download/downloader.py
```diff
@@ -62,9 +68,39 @@ def download(self):
...
+                # GPG decryption complete, move to normal logic
+            except Exception:
+                status = Status.FAILED
+                log.error(traceback.format_exc())
+                return status
+
         curr_working_dir = os.getcwd()
         os.chdir(working_dir)
         try:
+
```
lilida: Remove this extra line?

brennentsmith: 👍

PR #614, opened 2018-10-19
Adding integration with Nimbus (for add new stage workflow)

tobicodes:
Summary: 1) Create NimbusClient 2) Use Client to support workflow of adding new Teletraan stage


yongwen:
r+

PR #624, opened 2018-11-16
Making requesting permission to deploy not do unsafe arbitrary javascript execution.

TheLat:


deploy-board/deploy_board/templates/deploys/warning_no_deploy.tmpl
```diff
@@ -4,22 +4,10 @@
     <div id="noDeployAlert" class="alert alert-danger" role="alert" align="center" style="{{ hidden }}">
       <a href="#" class="close" data-dismiss="alert">&times;</a>
       <i><b style="font-size:24px;">Change freeze in effect !</b></i>
-      <p>You can create a change ticket <button id="changefreezeRequest" type="button" class="btn btn-default btn-sm" >here</button> to request approval.</p>
+      <p>You can create a change ticket <a href="{{jira_source_url}}">here</a>.</p>
```
rowoot: can you add `target='_blank'` so the link opens up on a new window?

yongwen:
r+
maybe would be nice to open a new window.

deploy-board/deploy_board/templates/deploys/warning_no_deploy.tmpl
```diff
@@ -4,22 +4,10 @@
...
       <p>Visit <a href="{{code_freeze_url}}">pinch/changefreeze</a> for more info</p>
```
rowoot: can you add target='_blank' so the link opens up on a new window?

PR #626, opened 2018-12-11
Rearranging priority order

edenjnbaptiste-pinterest:
Now the priority drop down will show in an order that goes from lowest to highest. This will be less confusing.


yongwen:
Please check how the DB is storing this priority. If the DB stores the
index, there might be a problem as the index has changed. If DB stores the
string, should be ok.

PR #628, opened 2019-01-11
Add nimbus integration 

tobicodes:
1) Add NimbusClient 
2) Create Identifiers when creating new stage (including clone stage)


yongwen:
you can rebase and squash the commit so the PR could only show one commit.

deploy-board/deploy_board/webapp/helpers/environs_helper.py
```diff
@@ -49,8 +50,19 @@
 DEPLOY_CONSTRAINT_TYPES = ['GROUP_BY_GROUP', 'ALL_GROUPS_IN_PARALLEL']
 
 deployclient = DeployClient()
+nimbusclient = NimbusClient()
```
yongwen: pinterest only?

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -788,22 +788,75 @@ def post(self, request, name, stage):
...
-    else:
-        data = {}
-        data['envName'] = name
-        data['stageName'] = stage
-        data['description'] = description
-        environs_helper.create_env(request, data)
-    return redirect('/env/' + name + '/' + stage + '/config/')
+    
+    if not IS_PINTEREST and not from_stage:
+        create_simple_stage(request, name, stage, description, external_id=None)
```
yongwen: would be nice to put "create_simple_stage" into common module similar to common.clone_from_stage_name, so looks more consistent.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -788,22 +788,75 @@ def post(self, request, name, stage):
...
+    
+    # Pinterest methods
+    if IS_PINTEREST:
```
yongwen: seems like all these "IF" blocks probably could be re-organized/optimized a bit for readability. something like:
if is_pinterest:
   create_identifer
if from_stage:
   clone
else:
   simple

deploy-board/deploy_board/settings.py
```diff
@@ -64,6 +64,10 @@
 TELETRAAN_SERVICE_FIXED_OAUTH_TOKEN = os.getenv("TELETRAAN_SERVICE_FIXED_OAUTH_TOKEN", None)
 TELETRAAN_HOST_INFORMATION_URL = os.getenv("HOST_INFORMATION_URL")
 
+# Nimbus service url
+NIMBUS_SERVICE_URL = os.getenv("NIMBUS_SERVICE_URL", "https://nimbus-beta.pinadmin.com")
```
yongwen: should not expose anything "pinadmin.com"?

PR #634, opened 2019-02-06
Modify python import and templates for meet new Django version. (1.11.18)

lixmgl:
We need upgrade Django from 1.6 to 1.11 for security vulnerability.
This PR's changes are needed to meet new Django version.

### DIFF
--- a/deploy-board/deploy_board/settings.py
+++ b/deploy-board/deploy_board/settings.py
@@ -16,10 +16,10 @@
 Django settings for deploy_board project.
 
 For more information on this file, see
-https://docs.djangoproject.com/en/1.6/topics/settings/
+https://docs.djangoproject.com/en/1.11/topics/settings/
 
 For the full list of settings and their values, see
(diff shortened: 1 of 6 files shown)


yongwen:
please provide description on this PR. e.g, security

deploy-board/deploy_board/settings.py
```diff
@@ -32,9 +32,21 @@
...
-    os.path.join(BASE_DIR, 'templates'),
-)
+TEMPLATES = [
+    {
+        'BACKEND': 'django.template.backends.django.DjangoTemplates',
+        'DIRS' : [os.path.join(BASE_DIR, 'templates')],
+        'APP_DIRS': True,
+        'OPTIONS': {
+            'context_processors': [
+                'django.template.context_processors.debug',
```
yongwen: do we really need debug? what is the impact on performance?

lixmgl: We do not need debug, only auth is required. I removed the other optional ones. Thanks.

PR #636, opened 2019-02-11
display warning message when find project name is default from nimbus

lixmgl:
image


deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -324,6 +324,19 @@ def get(self, request, name, stage=None):
         env_tag = tags_helper.get_latest_by_targe_id(request, env['id'])
         basic_cluster_info = None
         capacity_info = {'groups': groups}
+
+        stage_with_external_id= None
+        for env_stage in envs:
+            if env_stage['externalId'] is not None:
+                stage_with_external_id = env_stage
+                break
```
yongwen: It is possible that one of the stage in the environment  has a nimbus id that is not default project but the other stages in the environment still are default project. I think it is better to find out the stage this page is displaying and use that stage.

lixmgl: Thanks for pointing out, updated.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -324,6 +324,18 @@ def get(self, request, name, stage=None):
...
+        project_name_is_default = False
+        stage_with_external_id = None
+        for env_stage in envs:
+            if env_stage['externalId'] is not None and env_stage['stageName'] == stage:
+                stage_with_external_id = env_stage
+                break
+
+        if stage_with_external_id is not None and stage_with_external_id['externalId'] is not None:
+            existing_stage_identifier = environs_helper.get_nimbus_identifier(stage_with_external_id['externalId'])
+            project_name_is_default = True if existing_stage_identifier['projectName'] == "default" else False
```
yongwen: please add a None check to existing_stage_identifier

lixmgl: Updated, thanks for reminding.

PR #637, opened 2019-02-12
fix for cluster capacity button disappear after Django upgrade

lixmgl:


deploy-board/deploy_board/templates/environs/env_landing.html
```diff
@@ -129,7 +129,7 @@ <h4 class="panel-title pull-left">Current Deploy</h4>
 </div>
 
 {% if pinterest %}
-<script type="text/javascript" src="{% static "js/components/sharedcomponents.js?changedate=2017.1.4.140000"%}"></script>
```
yongwen: thanks. please do a gobal search on "static" reference

lixmgl: Thanks for reminding, updated all places.

PR #646, opened 2019-02-21
add wiki link for user data config settings in teletraan UI page

lixmgl:


deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -70,6 +70,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
                         <div class="form-group"></div>
                         <div class="form-group">
                             <label class="deployToolTip control-label col-xs-2" data-toggle="tooltip"  title="Customize user data">User data configs:</label>
+                            <a href="https://w.pinadmin.com/pages/viewpage.action?spaceKey=IN&title=Launching+a+new+service#Launchinganewservice-DeploytheservicetoCMP/Dockercluster">wikiHelp</a>
```
yongwen: as it is open source, please add pinterest specific link to env variable, same as other .pinadmin.com url.

lixmgl: Thanks for reminding, updated.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -56,6 +56,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
                         <div class="form-group"></div>
                         <div class="form-group">
                             <label class="deployToolTip control-label col-xs-2" data-toggle="tooltip" title="Customize user data">User data configs:</label>
+                            <a href="{{user_data_config_settings_wiki}}">wikiHelp</a>
```
yongwen: what if the env var is not set in non pinterest case? should we not show the wikihelp text?

yongwen: maybe use the "?" icon instead to be consistent with other ui element

lixmgl: Thanks for pointing out, updated.

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -70,6 +70,9 @@ <h4 class="panel-title pull-left">Capacity</h4>
                         <div class="form-group"></div>
                         <div class="form-group">
                             <label class="deployToolTip control-label col-xs-2" data-toggle="tooltip"  title="Customize user data">User data configs:</label>
+                            {% if is_pinterest %}
+                                <a href={{user_data_config_settings_wiki}}><span class="glyphicon glyphicon-question-sign"></span></a>
```
yongwen: can you add target="_blank" after href so the wiki will show up in a new tab?

lixmgl: good point, thanks, updated.

PR #648, opened 2019-03-01
update default placements list from 1c to 1d

lixmgl:


deploy-board/deploy_board/static/js/deploy-board.js
```diff
@@ -66,6 +66,8 @@ function getDefaultPlacement(capacityCreationInfo) {
     //The abstract name of group can be default assigned
     var defaultAssignedGroup = new Set(['us-east-1a','us-east-1c', 'us-east-1d',
         'us-east-1e', 'us-east-2a', 'us-east-2b', 'us-west-2b', 'us-west-2a']);
+    
+    var defaultAssginedGroupForC5Intance = new Set(['us-east-1a', 'us-east-1d', 'us-east-1e']);
```
yongwen: should we add us-east-2*

yongwen: i think we should keep all the 1a, 1d, 1e in both cell

deploy-board/deploy_board/static/js/deploy-board.js
```diff
@@ -100,13 +102,13 @@ function getDefaultPlacement(capacityCreationInfo) {
     $.each(this.capacityCreationInfo.placements, function (index, item) {
         if (item.assign_public_ip) {
             allPublicIPPlacements.push(item)
-            if (defaultAssignedGroup.has(item.abstract_name)){
+            if (defaultAssginedGroupForC5Intance.has(item.abstract_name)){
```
yongwen: what happen to non-c5 instances

deploy-board/deploy_board/static/js/deploy-board.js
lixmgl: update, thanks.

PR #657, opened 2019-03-20
add phobos link for each env in agent

lixmgl:
Screen Shot 2019-03-20 at 1 04 00 AM
Once you click the link, it will redirect you to this page:
Screen Shot 2019-03-20 at 1 04 31 AM

### DIFF
--- a/deploy-board/deploy_board/templates/hosts/host_details.html
+++ b/deploy-board/deploy_board/templates/hosts/host_details.html
@@ -41,18 +41,16 @@
 {% endif %}
 
 {% if agent_wrappers %}
-    <div class="panel panel-default">
+<div class="panel panel-default">
     {% include "panel_heading.tmpl" with panel_title="Agent Details" panel_body_id="agentDetailsId" direction="down" copy_host_name_button=True %}
     <div id="agentDetailsId" class="collapse in panel-body">
-
-    {% for agent_wrapper in agent_wrappers %}
-        {% with env=agent_wrapper.env agent=agent_wrapper.agent %}
-        {% include "hosts/host_details.tmpl" with agent=agent env=env%}
+        {% for agent_wrapper in agent_wrappers %}
+        {% with env=agent_wrapper.env agent=agent_wrapper.agent hostdetailsvalue=host_details.items|lookup:"Phobos Link" %}
+        {% include "hosts/host_details.tmpl" with agent=agent env=env hostdetailsvalue=hostdetailsvalue%}
         {% endwith %}
-    {% endfor %}
-
+        {% endfor %}
     </div>
-    </div>
+</div>
 {% endif %}
 
 <div class="panel panel-default">
--- a/deploy-board/deploy_board/templates/hosts/host_details.tmpl
+++ b/deploy-board/deploy_board/templates/hosts/host_details.tmpl
@@ -39,7 +39,11 @@
                 <td>{{ agent.failCount }}</td>
             </tr>
             <tr>
-                <td>Error Message</td>
+                <td>Error Message
+                    {% if pinterest %}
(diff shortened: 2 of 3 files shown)


deploy-board/deploy_board/templates/hosts/host_details.html
```diff
@@ -41,18 +41,24 @@
...
 {% if agent_wrappers %}
-    <div class="panel panel-default">
-    {% include "panel_heading.tmpl" with panel_title="Agent Details" panel_body_id="agentDetailsId" direction="down" copy_host_name_button=True %}
+<div class="panel panel-default">
+    {% include "panel_heading.tmpl" with panel_title="Agent Details" panel_body_id="agentDetailsId" direction="down"
+    copy_host_name_button=True %}
     <div id="agentDetailsId" class="collapse in panel-body">
-
-    {% for agent_wrapper in agent_wrappers %}
+        {% for key, value in host_details.items %}
```
lixmgl: This code works now, but I need help here to avoid extra for loop to get ip value from host_details map.

lixmgl: Or any idea for pass in host ip variable in a better way?

yongwen: the key 'phobos link' might not exists for non pinterest(for example), we still want to show the agent_wrapper for loop. so it is better to use custom filter see the end of https://stackoverflow.com/questions/2970244/django-templates-value-of-dictionary-key-with-a-space-in-it
add the custom filter in deploy-board/deploy_board/webapp/templatetags/utils.py

lixmgl: Thanks for the info, updated.

deploy-board/deploy_board/templates/hosts/host_details.html
```diff
@@ -41,18 +41,19 @@
...
-    {% include "panel_heading.tmpl" with panel_title="Agent Details" panel_body_id="agentDetailsId" direction="down" copy_host_name_button=True %}
+<div class="panel panel-default">
+    {% include "panel_heading.tmpl" with panel_title="Agent Details" panel_body_id="agentDetailsId" direction="down"
+    copy_host_name_button=True %}
     <div id="agentDetailsId" class="collapse in panel-body">
-
-    {% for agent_wrapper in agent_wrappers %}
+        {% with hostdetailsvalue=host_details.items|lookup:"Phobos Link" %}
+        {% for agent_wrapper in agent_wrappers %}
         {% with env=agent_wrapper.env agent=agent_wrapper.agent %}
```
yongwen: can you move hostdetailsvalue=host_details.items|lookup:"Phobos Link" into here to have one less {% with ... %}

deploy-board/deploy_board/templates/hosts/host_details.tmpl
```diff
@@ -39,7 +39,9 @@
                 <td>{{ agent.failCount }}</td>
             </tr>
             <tr>
-                <td>Error Message</td>
+                <td>Error Message
+                    <a href="{{ hostdetailsvalue }}/log/deploylog?file={{env.envName}}.log">Phobos Link for {{env.envName}}</a>
```
yongwen: please add a {% if hostdetailsvalue %} for non pinterest.
and say "Full deploy log for {{env.envName}}"

deploy-board/deploy_board/templates/hosts/host_details.html
lixmgl: updated, thanks for reminding.

deploy-board/deploy_board/templates/hosts/host_details.tmpl
lixmgl: Thanks for reminding, do you mean add {% if pinterest %} for displaying phobos  links?

PR #665, opened 2019-04-10
add a service add-on for console project

euccas:
This change adds a service add-on for console project. A service add-on button will be displayed when the project exists in console.


deploy-board/deploy_board/settings.py
```diff
@@ -86,6 +86,8 @@
 ENABLING_SERVICE_RATELIMIT_URL = os.getenv("ENABLING_SERVICE_RATELIMIT_URL", "OFF")
 KAFKA_MSGS_DELIVERED_METRIC = os.getenv("KAFKA_MSGS_DELIVERED_METRIC", "OFF")
 DASHBOARD_URL_ENDPOINT_FORMAT = os.getenv("DASHBOARD_URL_ENDPOINT_FORMAT","OFF")
+CONSOLE_PROJECT_URL_FORMAT = os.getenv("CONSOLE_PROJECT_URL_FORMAT", "OFF")
+CONSOLE_PROJECT_DATA_URL_FORMAT = os.getenv("CONSOLE_PROJECT_DATA_URL_FORMAT", "OFF")
```
tobicodes: Can we refer to these as `NIMBUS_PROJECT...` since the project info actually comes from Nimbus. Also, we already have some Nimbus-related code in the repo (e.g. the Nimbus client)

edenjnbaptiste-pinterest: Isn't this exposing internal Pinterest code? Especially the nimbus client. This seems like a huge security risk.

PR #666, opened 2019-04-12
remove launch config button for ASG groups

lixmgl:
Only display launch config button for non-ASG groups:
in the following screenshots:
deploy-sentinel-docker is ASG group
deploy-sentinel is non-ASG group
Screen Shot 2019-04-12 at 12 20 44 PM
Screen Shot 2019-04-12 at 12 21 01 PM

### DIFF
--- a/deploy-board/deploy_board/templates/groups/group_details.html
+++ b/deploy-board/deploy_board/templates/groups/group_details.html
@@ -37,12 +37,14 @@
         </a>
     </div>
     {% endif %}
+    {% if group_info == None %}
     <div class="row">
         <button id="launchInstance" class="deployToolTip btn btn-default btn-block"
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/templates/groups/group_details.html
```diff
@@ -37,12 +37,14 @@ <h4 class="panel-title pull-left">Groups</h4>
         </a>
     </div>
     {% endif %}
+    {% if group_info == None %}
```
euccas: is `group_info` a valid value only when the group is an ASG group?

lixmgl: right, non-ASG it is none.

PR #671, opened 2019-04-19
add a button linking to environment owner in the deployment menu.

euccas:
This change adds a button linking to environment owner in the deployment menu. The link is displayed when environment owner information is available.

The link button is added next to "Rollback"

Screen Shot 2019-04-19 at 2 55 33 PM


yongwen:
Agree with Rainie that the link next to rollback seems a bit strange. Maybe put the link next after the "active deployment" ? 
Also the name "environment owner" maybe better to be "project: <name>" with link. To make it less confusion with the telegram environment. As a telegram environment might have multiple stage and there is possibility that different stage belongs to different project. And the explicit name helps user to not need to click the link to see the project it belongs to.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -344,6 +344,9 @@ def get(self, request, name, stage=None):
             basic_cluster_info = clusters_helper.get_cluster(request, env.get('clusterName'))
             capacity_info['cluster'] = basic_cluster_info
 
+        console_project_exist = environs_helper.get_nimbus_project(name)
```
yongwen: the variable "existing_stage_identifier" above should already has the console_project info. please use that instead of calling the nimbus API again for better performance.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -374,6 +377,8 @@ def get(self, request, name, stage=None):
                 "csrf_token": get_token(request),
                 "display_stopping_hosts": DISPLAY_STOPPING_HOSTS,
                 "project_name_is_default": project_name_is_default,
+                "console_project_exist": console_project_exist,
```
yongwen: can we use "console_project_env_url" to determine this exist flag? so we pass in one less variable.

deploy-board/deploy_board/webapp/env_views.py
yongwen: also, the "name" variable here is the env name, it is not the project name, so calling the nimbus api with the env name is not correct. please look at existing_stage_identifier.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -344,6 +344,9 @@ def get(self, request, name, stage=None):
             basic_cluster_info = clusters_helper.get_cluster(request, env.get('clusterName'))
             capacity_info['cluster'] = basic_cluster_info
 
+        console_project_exist = environs_helper.get_nimbus_project(name)
+        console_project_env_url = environs_helper.get_console_project_env_url(name)
```
yongwen: same here the variable "name" here should be the nimbus project name associated with the stage.

euccas:
> Can I ask why add this link next to rollback button?

The intent was to add this link to a place where operators can easily notice and see. So I added it to the left of the existing operation buttons. I can move it more to the left, since it doesn't seem to related to rollback or other operations.

PR #672, opened 2019-04-26
kill ngapp2 deploy page

lixmgl:
Screen Shot 2019-05-06 at 2 08 35 PM

### DIFF
--- a/deploy-board/deploy_board/templates/environs/simple_envs.tmpl
+++ b/deploy-board/deploy_board/templates/environs/simple_envs.tmpl
@@ -2,9 +2,6 @@
 {% load static %}
 <div class="row">
 <ul class="list-group">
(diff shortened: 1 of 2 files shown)


lixmgl:
> can you remove all the urls related to ngapp2? also i think the ngapp2 is hardcoded into the home page as the recent environment, we may need to remove that as well.

Can I confirm should we just remove: /ngapp2/deploy/
Or we should remove all urls from this list?
ngapp2/deploy/
ngapp2/new_deploy/
ngapp2/promote_to/
ngapp2/deploy_to/
ngapp2/update_deploy_progress/
ngapp2/deploys/
ngapp2/compare_deploys/

If second case, I can just remove ngapp2_urls.py file, which is my current PR is doing.

PR #674, opened 2019-05-10
fix spiffe_id when clone cluster

yongwen:
when clone a cluster, the spiffe_id from the source cluster was cloned over which is not the expected behavior. the spiffe_id should be regenerated from the dest_env_name and dest_stage_name.
using the create_cluster_with_env function instead of create_cluster function will regenerate the correct spiffe_id for the dest cluster.


yongwen:
tested local deploy_board pointing to rodimus-integ

PR #675, opened 2019-05-14
add constraints to commit search results.

euccas:
- when searching a commit from UI search box, return commits that match
both SHA and build name.
- change unit tests of BuildDAO to add testing for the added search
constraints.
- change tools script to publish builds with build name matching env
stage configuration, so that deploy can be created for these builds.
- update .gitignore to include vscode temp directory.

**Example:**

Commits 1034481 with multiple different build names exist (sample-service-0, sample-service-1, sample-service-2):

commit_build_0
commit_build_1
commit_build_2

Search commit "1034481" returns only one commit, which has build name "sample-service-0".

search_result

Display "build name" in build search result:

pick_a_build

Search a build from stage:

build_search_from_stage

Search a build from global build search on the top corner:

build_search_from_top

### DIFF
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -981,7 +981,7 @@ def deploy_build(request, name, stage, build_id):
 
 def deploy_commit(request, name, stage, commit):
     env = environs_helper.get_env_by_stage(request, name, stage)
-    builds = builds_helper.get_builds_and_tags(request, commit=commit)
+    builds = builds_helper.get_builds_and_tags(request, name=env.get('buildName', ''), commit=commit)
     current_build = None
     if env.get('deployId'):
         deploy = deploys_helper.get(request, env['deployId'])
--- a/.gitignore
+++ b/.gitignore
@@ -6,11 +6,14 @@ deploy-service/teletraanservice/teletraan/
 .idea/
 *.iml
 
-# C extensions
-*.so
-
 # vim
 *.swp
+
+# VSCode
+.vscode/
+
+# C extensions
+*.so
 
 # Ignore dirs
 build_id.properties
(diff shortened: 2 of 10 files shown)


yongwen:
can you display the buildname info in the build page, so that it can be distinguishable for each commit?
especially, if user search a commit from the top right corner build box, we should show all buildname associated with the commit. and if search within a stage/env, it should filter by the buildname.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -981,7 +981,7 @@ def deploy_build(request, name, stage, build_id):
 
 def deploy_commit(request, name, stage, commit):
     env = environs_helper.get_env_by_stage(request, name, stage)
-    builds = builds_helper.get_builds_and_tags(request, commit=commit)
+    builds = builds_helper.get_builds_and_tags(request, name=env.get('buildName', ''), commit=commit)
```
lixmgl: Is this possible if env does not contain 'buildName'? Do we need handle the case when name = None?

euccas: In my understanding it's possible but rare for an env to not have a 'buildName'. When it does not have a buildName, Teletraan will not display builds on the left for it. Here the implementation is when buildName does not exist, will search only by commit SHA.

euccas:
> can you display the buildname info in the build page, so that it can be distinguishable for each commit?
> especially, if user search a commit from the top right corner build box, we should show all buildname associated with the commit. and if search within a stage/env, it should filter by the buildname.

Searching a commit from the top right corner box will show all commits regardless of the build name. That search box calls another function, where only "commit" is used as a search parameter.

I will add the buildname column in the build page. It can help a lot to tell the source of the commit SHA.

PR #676, opened 2019-05-14
copy the external_id in clone cluster

yongwen:
when the "clone cluster" button is clicked in the Capacity page, the external_id need to be copied over to the new stage.

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -622,12 +622,14 @@ def clone_cluster(request, src_name, src_stage):
         ##0. teletraan service get src env buildName
         src_env = environs_helper.get_env_by_stage(request, src_name, src_stage)
         build_name = src_env.get('buildName', None)
+        external_id = src_env.get('externalId', None)
 
         ##1. teletraan service create a new env
         dest_env = environs_helper.create_env(request, {
             'envName': dest_name,
             'stageName': dest_stage,
-            'buildName': build_name
+            'buildName': build_name,
+            'externalId': external_id
         })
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -622,12 +622,14 @@ def clone_cluster(request, src_name, src_stage):
         ##0. teletraan service get src env buildName
         src_env = environs_helper.get_env_by_stage(request, src_name, src_stage)
         build_name = src_env.get('buildName', None)
+        external_id = src_env.get('externalId', None)
```
euccas: When external_id is None, do we need display a warning? The warning can be displayed here when cloning cluster happens, or somewhere else in the source env stage's related pages.

edenjnbaptiste-pinterest:
Are there no tests in Teletraan?

deploy-board/deploy_board/webapp/cluster_view.py
lixmgl: If external Id is None, do we still need clone it or need create it instead?

yongwen: for non-pinterest, external_id will be none. so cloning should still happen.  for pinterest, the external_id should not be none from source cluster. if so, there may already an issue from the source cluster. i feel it is better to just clone just what is from the source.

yongwen:
> Are there no tests in Teletraan?

for UI there is no unit test.

deploy-board/deploy_board/webapp/cluster_view.py
euccas: yes considering when external_id is missing we will have the "Missing project ownership info" warning, here we can just clone from source.

PR #677, opened 2019-05-21
Do not allow customers to pass in spiffe_id during post call update c…

lixmgl:
…luster

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -30,7 +30,7 @@ import logging
 from helpers import baseimages_helper, hosttypes_helper, securityzones_helper, placements_helper, \
     autoscaling_groups_helper, groups_helper, cells_helper
 from helpers import clusters_helper, environs_helper, environ_hosts_helper
-from helpers.exceptions import NotAuthorizedException
+from helpers.exceptions import NotAuthorizedException, TeletraanException
 import common
 import traceback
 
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -220,6 +220,17 @@ def post(self, request, name, stage):
...
+            current_cluster = clusters_helper.get_cluster(request, cluster_name)
+            log.info("getting current Cluster Configuration is {}", current_cluster)
+            if 'spiffe_id' not in current_cluster and 'spiffe_id' in cluster_info:
+                log.info("Do not allow pass in spiffe_id if current cluster does not have one{}", cluster_info['spiffe_id'])
+                del cluster_info['spiffe_id']
+            if 'spiffe_id' in current_cluster and 'spiffe_id' in cluster_info:
+                if current_cluster['spiffe_id'] != cluster_info['spiffe_id']:
+                    log.info("Do not allow users to update customized spiffe_id {} when current spiffe_id {}",
+                    cluster_info['spiffe_id'], current_cluster['spiffe_id'])
+                    cluster_info['spiffe_id'] = current_cluster['spiffe_id']
```
yongwen: can we throw exception when the spiffe_id does not match? instead of silently ignore it.

lixmgl: Do you mean if user pass in a different spiffe_id , we need throw exception and prevent user from calling update_cluster?
My code currently will upgrade user's spiffe_id for them, and continue.

lixmgl: > can we throw exception when the spiffe_id does not match? instead of silently ignore it.
You mean when user pass in a different spiffe_id , we need throw an exception and prevent user from calling update_cluster?
My current code is upgrading user's spiffe_id, and continue call upgrade_cluster.
Is my current logic expected?

krockpot:
@lixmgl  is this only for the update call? What about cluster creation? We want spiffe to be machine generated, so we should also ignore / catch an exception if one is passed in at creation time.

deploy-board/deploy_board/webapp/cluster_view.py
lixmgl: Updated, please review.

lixmgl:
> @lixmgl is this only for the update call? What about cluster creation? We want spiffe to be machine generated, so we should also ignore / catch an exception if one is passed in at creation time.

Thanks for reminding, added same check when create cluster.

PR #680, opened 2019-06-11
Change Hotfix API to authorize publisher role.

euccas:
Change Hotfix update API to authorize publisher role. Due to a recent security change, this change in Teletraan is required so that Jenkins private jobs can be created.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Hotfixs.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Hotfixs.java
@@ -84,7 +84,7 @@ public class Hotfixs {
     @Path("/{id : [a-zA-Z0-9\\-_]+}")
     public void update(@Context SecurityContext sc, @PathParam("id") String id,
         HotfixBean hotfixBean) throws Exception {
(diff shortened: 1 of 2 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Hotfixs.java
```diff
@@ -84,7 +84,11 @@ public HotfixBean get(@PathParam("id") String id) throws Exception {
     @Path("/{id : [a-zA-Z0-9\\-_]+}")
     public void update(@Context SecurityContext sc, @PathParam("id") String id,
         HotfixBean hotfixBean) throws Exception {
-        authorizer.authorize(sc, new Resource(Resource.ALL, Resource.Type.SYSTEM), Role.OPERATOR);
+        try {
+            authorizer.authorize(sc, new Resource(Resource.ALL, Resource.Type.SYSTEM), Role.PUBLISHER);
+        } catch (Exception e) {
+            LOG.info("Failed to authorize hotfix {} with {}.", id, hotfixBean);
```
yongwen: We should let the exception throw out instead of catch and swallow.

euccas: Good point. Changed to throw out exception.

PR #682, opened 2019-06-24
update logic to disallow post cluster info contains spiffe_id

lixmgl:


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -224,17 +224,9 @@ def post(self, request, name, stage):
...
-            log.info("getting current Cluster Configuration is {}", current_cluster)
-            if 'spiffe_id' not in current_cluster and 'spiffe_id' in cluster_info:
-                log.error("Teletraan does not support user to pass in customized spiffe_id %s if exising cluster does not have" % cluster_info['spiffe_id'])
-                raise TeletraanException("Teletraan does not support user to pass in customized spiffe_id")
-            if 'spiffe_id' in current_cluster and 'spiffe_id' in cluster_info:
-                if current_cluster['spiffe_id'] != cluster_info['spiffe_id']:
-                    log.error("Teletraan does not support user to change spiffe_id %s , current spiffe_id %s"
-                        % (cluster_info['spiffe_id'], current_cluster['spiffe_id']))
-                    raise TeletraanException("Teletraan does not support user to change spiffe_id")
+            if 'spiffe_id' in cluster_info:
```
yongwen: did you test the case that a normal update cluster configuration?  i thought in the normal case, the unchanged spiffe_id is also passed into the cluster_info and is required.

lixmgl: good catch, thanks. Please review updated change.

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -30,6 +30,7 @@
 from helpers import autoscaling_metrics_helper
 from diff_match_patch import diff_match_patch
 from deploy_board import settings
+from helpers.exceptions import NotAuthorizedException, TeletraanException
```
euccas: Is "NotAuthorizedException" used somewhere?

Other changes look good to me.

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -179,6 +180,11 @@ def update_launch_config(request, group_name):
             launchRequest["assignPublicIp"] = True
         else:
             launchRequest["assignPublicIp"] = False
+
+        if params["userData"].find("spiffe_id"):
```
yongwen: what if there was a spiffe_id in the userData and user did not change it but change other part of the userdata, we should still support this case.

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -76,9 +76,10 @@ def post(self, request, name, stage):
             cluster_info = json.loads(request.body)
 
             log.info("Create Capacity in the provider")
-            if 'spiffe_id' in cluster_info:
-                log.error("Teletraan does not support user to change spiffe_id %s"  % cluster_info['spiffe_id'])
-                raise TeletraanException("Teletraan does not support user to create spiffe_id")
+            if 'configs' in cluster_info:
```
yongwen: looks like we change the behavior here, was it a bug before that spiffe_id supposed to be in cluster_info['config'] but not in cluster_info?

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -224,17 +226,16 @@ def post(self, request, name, stage):
...
-                if current_cluster['spiffe_id'] != cluster_info['spiffe_id']:
-                    log.error("Teletraan does not support user to change spiffe_id %s , current spiffe_id %s"
-                        % (cluster_info['spiffe_id'], current_cluster['spiffe_id']))
-                    raise TeletraanException("Teletraan does not support user to change spiffe_id")
+            if 'configs' in current_cluster and 'configs' in cluster_info:
+                if 'spiffe_id' in current_cluster['configs'] and 'spiffe_id' in cluster_info['configs']:
+                    if current_cluster['configs']['spiffe_id'] != cluster_info['configs']['spiffe_id']:
+                       log.error("Teletraan does not support user to change spiffe_id %s" % cluster_info['spiffe_id'])
+                       raise TeletraanException("Teletraan does not support user to create spiffe_id")
+
```
yongwen: what if current_cluster has spiffe_id but the new cluster_info remove it, this logic will allow it through.

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -224,17 +226,16 @@ def post(self, request, name, stage):
...
-            if 'spiffe_id' in current_cluster and 'spiffe_id' in cluster_info:
-                if current_cluster['spiffe_id'] != cluster_info['spiffe_id']:
-                    log.error("Teletraan does not support user to change spiffe_id %s , current spiffe_id %s"
-                        % (cluster_info['spiffe_id'], current_cluster['spiffe_id']))
-                    raise TeletraanException("Teletraan does not support user to change spiffe_id")
+            if 'configs' in current_cluster and 'configs' in cluster_info:
+                if 'spiffe_id' in current_cluster['configs'] and 'spiffe_id' in cluster_info['configs']:
+                    if current_cluster['configs']['spiffe_id'] != cluster_info['configs']['spiffe_id']:
+                       log.error("Teletraan does not support user to change spiffe_id %s" % cluster_info['spiffe_id'])
+                       raise TeletraanException("Teletraan does not support user to create spiffe_id")
```
yongwen: exception msg is different than the log message above.

deploy-board/deploy_board/webapp/group_view.py
lixmgl: Yes, it is used by other existing functions. Thanks for reminding.

deploy-board/deploy_board/webapp/cluster_view.py
lixmgl: Yes, it was a bug.

deploy-board/deploy_board/webapp/cluster_view.py
lixmgl: updated, thanks for reminding.

deploy-board/deploy_board/webapp/cluster_view.py
lixmgl: Good catch. Added this case.

deploy-board/deploy_board/webapp/group_view.py
lixmgl: params is coming from request. I think this protection is good enough. 
The main purpose here is preventing user to update spiffe_id.
If data coming from user which is params does not contain spiffe_id, we are good.

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -179,6 +180,30 @@ def update_launch_config(request, group_name):
...
+                if ":" in item:
+                    pair = item.split(":")
+                    if pair[0] == "spiffe_id" and userPassInSpiffeID is None:
+                        log.error("Teletraan does not support user to remove spiffe_id in userData field")
+                        raise TeletraanException("Teletraan does not support user to remove spiffe_id")
+
+                    if pair[0] == "spiffe_id" and pair[1] != userPassInSpiffeID:
+                        log.error("Teletraan does not support user to update spiffe_id in userData field")
+                        raise TeletraanException("Teletraan does not support user to update spiffe_id")
+
```
yongwen: looks like there is still a case that pair[0] != "spiffe_id" and userPassInSpiffeID

lixmgl: Nice catch :) Updated, thanks for pointing out.

deploy-board/deploy_board/webapp/group_view.py
yongwen: looks like NotAuthorizedException is not used in this file.

PR #684, opened 2019-08-08
implement whitelisting stages to allow private build deploy

yongwen:
we want to disallow private build deploy by default. only the whitelist stages are able to deploy private build, by setting the allow_private_build column in the environs table to 1.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -59,8 +59,9 @@ import java.io.Serializable;
  * cluster_name        VARCHAR(128)
  * max_parallel_rp     INT           NOT NULL DEFAULT 1,
  * schedule_id         VARCHAR(22)   DEFAULT NULL,
- * deploy_constraint_id VARCHAR(22)   DEFAULT NULL
- * external_id CHAR(36)
+ * deploy_constraint_id VARCHAR(22)   DEFAULT NULL,
+ * external_id CHAR(36),
+ * allow_private_build TINYINT(1)    DEFAULT 0
  * <p>
  * PRIMARY KEY   (env_id)
  * );
@@ -179,6 +180,9 @@ public class EnvironBean implements Updatable, Serializable {
 
     @JsonProperty("externalId")
     private String external_id;
+
+    @JsonProperty("allowPrivateBuild")
+    private Boolean allow_private_build;
 
     public String getWebhooks_config_id() {
         return webhooks_config_id;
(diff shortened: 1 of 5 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -204,13 +204,21 @@ public Response create(
...
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, "Build id can not be empty.");
         }
         BuildBean buildBean = buildDAO.getById(buildId);
+        
+        // check build name must match stage config
         if(!buildBean.getBuild_name().equals(envBean.getBuild_name())) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("Build name (%s) does not match stage config (%s).", buildBean.getBuild_name(), envBean.getBuild_name()));
         }
+        // check if the stage is whitelisted (allow_private_build) for private build
+        if(buildBean.getScm_branch().equals("private") && ! envBean.getAllow_private_build()) {
```
lixmgl: One general question here:
How we whitelist stage for customers? 
By setting their branch name to private?

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -59,8 +59,9 @@
  * cluster_name        VARCHAR(128)
  * max_parallel_rp     INT           NOT NULL DEFAULT 1,
  * schedule_id         VARCHAR(22)   DEFAULT NULL,
- * deploy_constraint_id VARCHAR(22)   DEFAULT NULL
- * external_id CHAR(36)
+ * deploy_constraint_id VARCHAR(22)   DEFAULT NULL,
+ * external_id CHAR(36),
+ * allow_private_build TINYINT(1)    DEFAULT 0
  * <p>
```
lixmgl: What happen if a new whitelist stage came? Overwrite previous whitelist stage?

yongwen:
> General question: How can we whitelist the stage?
right now we don't have the UI to do the whitelist. to whitelist, the teletraan admin need to do a manual DB update to the  allow_private_build column in the "environs" table.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
yongwen: when a stage is created, the allow_private_build column/flag is default to false, meaning that no private build is allowed.  Customer need to request Teletraan Admin to whitelist this stage by setting the column/flag to 1.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
yongwen: do a manual DB update to the allow_private_build column in the "environs" table.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
lixmgl: Thanks for explaining. What is next step?  Once Teletraan Admin set flag to 1 in DB for them, customers can use "public private build" button to upload their build, right?

PR #685, opened 2019-08-16
Update error message to point to process doc

jmadler:
Update error message to point to process doc for permission requests for private builds

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
@@ -216,7 +216,7 @@ public class EnvDeploys {
         }
         // check if the stage is whitelisted (allow_private_build) for private build
         if(buildBean.getScm_branch().equals("private") && ! envBean.getAllow_private_build()) {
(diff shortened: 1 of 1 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -216,7 +216,7 @@ public Response create(
         }
         // check if the stage is whitelisted (allow_private_build) for private build
         if(buildBean.getScm_branch().equals("private") && ! envBean.getAllow_private_build()) {
-            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please Contact #teletraan and #security-related to whitelist your stage for deploying private build"));
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please visit https://w.pinadmin.com/x/_SKNB to request permission"));
```
yongwen: thanks. Just realized that we may need to defer the URL from a environment variable. please make the URL to defer from env defined in pinterest sepecific teletraanconfig.

tylerwowen:
Close as PR is too old.

PR #689, opened 2019-09-18
Open "create env" modal if redirect_create_env_page_url not supplied

### LINKED ISSUE #686: Create new environment button on UI redirects to /None
Hello,

I'm setting up Teletraan in a containerized environment, and have deploy-board and deploy-service running in containers and, by all accounts, successfully communicating. When I go to click "Create environment" in the UI, I am redirected to "{ base_url }/None". Checking on the template that generates this button, I see that the button's URL is populated by this env var: https://github.com/pinterest/teletraan/blob/1faf2b8d9fc6f1f4c735c2d0f64e395ab04198a2/deploy-board/deploy_board/settings.py#L231

Which is then referenced in the templates here (as an example): https://github.com/pinterest/teletraan/blob/4315016ee5c0d15e8256191bb333889ba6f69027/deploy-board/deploy_board/templates/environs/envs_base.html#L78

Since I do not have this env var set, I am being redirected to /None. Here is a demo: https://ookla.d.pr/i5iDig

I tried setting this env var to a valid URL, but the Create Environment button then acts as a simple link to whatever I set the redirect URL to. At no point am I able to actually create an environment or presented with the correct dialog from the button click action.

Let me know if there is any more detail I can provide about my set up. Thanks!

moshe:
I came into similar issue like @gvwirth at #686 
Added new condition for checking `redirect_create_env_page_url`

closes #686

### DIFF
--- a/deploy-board/deploy_board/templates/environs/envs_base.html
+++ b/deploy-board/deploy_board/templates/environs/envs_base.html
@@ -75,7 +75,7 @@
         <h4 class="panel-title pull-left">Environments</h4>
     </div>
 	<div class="row">
-      {% if redirect_create_env_page_url %}
+      {% if redirect_create_env_page_url is not none %}
         <a href='{{ redirect_create_env_page_url }}' target='_blank' class="deployToolTip btn btn-default btn-block">
       {% else %}
         <a class="deployToolTip btn btn-default btn-block" data-original-title="" title="" data-toggle="modal" data-target="#newEnvModalId">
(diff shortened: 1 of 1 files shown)


yongwen (dismissed):

tuck182:
This link also appears in [templates/environs/envs_landing.html](https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/templates/environs/envs_landing.html#L18). I believe the same change should be applied there.

gvwirth:
Hi -- any updates on getting these fixes merged? We're hoping to update our Teletraan installs and it'd be great if these fixes were merged.

PR #691, opened 2019-09-26
Add option to specify token for anonymous users

moshe:


yongwen:
what is the rationale/usecase for a token for anonymous user?

moshe:
The main use-case:
A load balancer / reverse-proxy is handling authentication for the deploy-board and the deploy-service is protected by token authentication.
That way requests coming from deploy-board will have permissions to access the deploy-service without getting "Auth required" errors.

PR #693, opened 2019-11-10
add a new flag in build db to protect deploy only happens from truste…

lixmgl:
…d artifact build

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -61,7 +61,8 @@ import java.io.Serializable;
  * schedule_id         VARCHAR(22)   DEFAULT NULL,
  * deploy_constraint_id VARCHAR(22)   DEFAULT NULL,
  * external_id CHAR(36),
- * allow_private_build TINYINT(1)    DEFAULT 0
+ * allow_private_build TINYINT(1)    DEFAULT 0,
+ * ensure_trusted_build TINYINT(1)    DEFAULT 0,
  * <p>
  * PRIMARY KEY   (env_id)
  * );
@@ -183,6 +184,9 @@ public class EnvironBean implements Updatable, Serializable {
 
     @JsonProperty("allowPrivateBuild")
     private Boolean allow_private_build;
+
+    @JsonProperty("ensureTrustedBuild")
+    private Boolean ensure_trusted_build;
 
     public String getWebhooks_config_id() {
         return webhooks_config_id;
@@ -488,6 +492,14 @@ public class EnvironBean implements Updatable, Serializable {
         this.allow_private_build = allow_private_build;
     }
 
+    public Boolean getEnsure_trusted_build() {
+        return ensure_trusted_build;
+    }
+
+    public void setEnsure_trusted_build(Boolean ensure_trusted_build) {
+        this.ensure_trusted_build = ensure_trusted_build;
+    }
+
     @Override
     public SetClause genSetClause() {
         SetClause clause = new SetClause();
@@ -529,6 +541,7 @@ public class EnvironBean implements Updatable, Serializable {
         clause.addColumn("deploy_constraint_id", deploy_constraint_id);
         clause.addColumn("external_id", external_id);
         clause.addColumn("allow_private_build", allow_private_build);
+        clause.addColumn("ensure_trusted_build", ensure_trusted_build);
         return clause;
     }
 
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
@@ -219,6 +219,11 @@ public class EnvDeploys {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please Contact #teletraan and #security-related to whitelist your stage for deploying private build"));
         }
         
+        // check if the build is from a trusted artifact url
+        if(!buildBean.getArtifact_url().startsWith("https://soxrepo.pinadmin.com") && envBean.getEnsure_trusted_build()) {
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This build is not allowed to deploy. Please Contact #teletraan and #security-related to ensure the build artifact is published to a trusted url"));
+        }
+
         String deployId = deployHandler.deploy(envBean, buildId, description, operator);
         LOG.info("Successfully create deploy {} for env {}/{} by {}.", deployId, envName, stageName, operator);
 
--- a/tools/mysql/schema-update-6.sql
+++ b/tools/mysql/schema-update-6.sql
@@ -1,0 +1,10 @@
+-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+-- ALWAYS BACKUP YOUR DATA BEFORE EXECUTING THIS SCRIPT
+-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+# This script upgrade DB schema from version 5 to version 6
+
+ALTER TABLE environs ADD COLUMN ensure_trusted_build TINYINT(1) NOT NULL DEFAULT 0;
+
+-- make sure to update the schema version to 5
+UPDATE schema_versions SET version=6;
--- a/deploy-service/common/src/main/resources/sql/deploy.sql
+++ b/deploy-service/common/src/main/resources/sql/deploy.sql
@@ -41,6 +41,7 @@ CREATE TABLE IF NOT EXISTS environs (
     deploy_constraint_id      VARCHAR(22),
     external_id          CHAR(36),
     allow_private_build TINYINT(1)    NOT NULL DEFAULT 0,
+    ensure_trusted_build TINYINT(1)    NOT NULL DEFAULT 0,
     PRIMARY KEY   (env_id)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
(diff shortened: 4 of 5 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/bean/BuildBean.java
```diff
@@ -36,6 +36,7 @@
  * publish_date     BIGINT              NOT NULL,
  * publisher        VARCHAR(64),
  * scm_info         VARCHAR(512),
+ * ensure_trusted_build  TINYINT(1)    DEFAULT 0
```
yongwen: for build table, i think we don't need this field.  the enforcement is from environment table. so when an environment is set to ensure_trusted_build, the build url need to be from the trusted_build_url

tools/mysql/schema-update-6.sql
```diff
@@ -0,0 +1,10 @@
+-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+-- ALWAYS BACKUP YOUR DATA BEFORE EXECUTING THIS SCRIPT
+-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+# This script upgrade DB schema from version 5 to version 6
+
+ALTER TABLE builds ADD COLUMN ensure_trusted_build TINYINT(1) NOT NULL DEFAULT 0;
```
yongwen: your sql file "deploy.sql" change is in environs table

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -219,6 +219,11 @@ public Response create(
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please Contact #teletraan and #security-related to whitelist your stage for deploying private build"));
         }
         
+        // check if the build is from trusted artifact url
+        if(!buildBean.getArtifact_url().contains("deployrepo.pinadmin.com") && ! buildBean.getEnsure_trusted_build()) {
```
yongwen: the trusted_url should be configured in the config file. 
the enforcement should be checking the environBean.ensure_trusted_build, if ensure_trusted, the artifact_url need to be from trusted_url.  probably should use prefix instead of contains.

lixmgl: Can I ask where is this "trusted_url" set?

yongwen: could be similar to https://github.com/pinterest/teletraan/pull/692 and the config is in server.yaml. if the config is not set, could ignore the check. for non-opensource config, it is in server-prod.yaml or server-integ.yaml

deploy-board/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:2.7
+FROM 998131032990.dkr.ecr.us-east-1.amazonaws.com/python2.7.11:latest
```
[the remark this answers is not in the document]
lixmgl: Thanks for pointing out. Reverted.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/BuildBean.java
lixmgl: Thanks for pointing out. updated.

tools/mysql/schema-update-6.sql
lixmgl: Updated.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
lixmgl: Thanks for explaining, currently I added 2 checks 1) invalid url check 2) untrusted url check. Please review.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/BuildBean.java
```diff
@@ -184,8 +184,8 @@ public Long getPublish_date() {
         return publish_date;
     }
 
-    public void setPublish_date(Long publish_date) {
-        this.publish_date = publish_date;
+    public void setEnsure_trusted_build(Boolean ensure_trusted_build) {
```
yongwen: do we need this in buildBean?

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -488,6 +492,14 @@ public void setAllow_private_build(Boolean allow_private_build) {
         this.allow_private_build = allow_private_build;
     }
 
+    public void setPublish_date(Long publish_date) {
```
yongwen: should it be setEnsure.....

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -218,7 +218,17 @@ public Response create(
         if(buildBean.getScm_branch().equals("private") && ! envBean.getAllow_private_build()) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please Contact #teletraan and #security-related to whitelist your stage for deploying private build"));
         }
+
+        // check if the build is from a valid artifact url
+        if(!buildBean.getArtifact_url().startsWith("https://soxrepo.pinadmin.com") && !buildBean.getArtifact_url().startsWith("https://deployrepo.pinadmin.com/")) {
```
yongwen: for non open source case, this will prevent any deploy. may need to support that case.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -218,7 +218,17 @@ public Response create(
...
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a build from invalid url. Please use a valid url either from deployrepo or soxrepo!"));
+        }
         
+        // check if the build is from a trusted artifact url
+        if(buildBean.getArtifact_url().startsWith("https://soxrepo.pinadmin.com") && !envBean.getEnsure_trusted_build()) {
```
yongwen: should it be:   if (envBean.getEnsure_trusted_build() && ! buildBean.getArtifact_url().startsWith("https://soxrepo.pinadmin.com")) .....

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/BuildBean.java
lixmgl: Good catch, deleted.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
lixmgl: Removed that case to unblock normal deploy. Thanks for pointing out.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
lixmgl: Good catch, updated.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
lixmgl: Updated.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -219,6 +219,11 @@ public Response create(
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please Contact #teletraan and #security-related to whitelist your stage for deploying private build"));
         }
         
+        // check if the build is from a trusted artifact url
+        if(!buildBean.getArtifact_url().startsWith("https://soxrepo.pinadmin.com") && envBean.getEnsure_trusted_build()) {
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a build from untrusted url. Please Contact #teletraan and #security-related to whitelist your sox url for deploying"));
```
yongwen: please change "Contact ..... " to ensure the build artifact is published to a trusted url."

lixmgl: Updated :)

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -219,6 +219,11 @@ public Response create(
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This stage does not allow deploying a private build. Please Contact #teletraan and #security-related to whitelist your stage for deploying private build"));
         }
         
+        // check if the build is from a trusted artifact url
+        if(!buildBean.getArtifact_url().startsWith("https://soxrepo.pinadmin.com") && envBean.getEnsure_trusted_build()) {
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, String.format("This build is not allowed to deploy. Please Contact #teletraan and #security-related to ensure the build artifact is published to a trusted url"));
```
krockpot: @lixmgl  this error message probably shouldn't contain pinterest specific information, given this is open source

lixmgl: Thanks Jeremy, I will follow up with you requirement details:)

PR #695, opened 2019-11-24
Address com.fasterxml.jackson.core:jackson-databind version for secur…

lixmgl:
…ity vulnerability


lixmgl:
To address this issue
Screen Shot 2019-11-24 at 12 09 38 AM

PR #704, opened 2020-03-12
Added a helper text in Add/Edit Constraint Modal

rowoot:
Added a helper text when creating a constraint

image


deploy-board/deploy_board/templates/deploys/edit_constraint_modal.tmpl
```diff
@@ -41,6 +41,7 @@
                                    data-placement="right"
                                    data-trigger="manual"
                                    title="numbers only." value="{{ max_parallel }}">
+                            <small>Note: This number MUST be smaller the config/max_parallel_hosts</small>
```
euccas: smaller **than** ?

rowoot: thanks for finding this -- i updated the text and added a link to the env/stage config page as well for easy access to user

image

PR #705, opened 2020-03-12
Mark DeployConstraint as ERROR only if SQLException

rowoot:
After discussion w/ Suli -- incorporated her suggestion to 
1. Query DeployTag once lock is acquired.
2. Add a check in exception to ensure STATE=ERROR only when it is NOT a SQL Exception --> this will ensure retry if there is any connection pool issues
3. Export error with stacktrace

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
@@ -130,9 +130,7 @@ public class DeployTagWorker implements Runnable {
                         processEachEnvironConstraint(latestJob);
                     } catch (Exception e) {
                         LOG.error("failed to process job: {} Error {} stack {}", latestJob.toString(), ExceptionUtils.getRootCauseMessage(e), ExceptionUtils.getStackTrace(e));
-                        if (e instanceof SQLException) {
-                            // Don't do anything
-                        } else {
+                        if (!SQLException.class.isInstance(e)) {
                             latestJob.setState(TagSyncState.ERROR);
                             deployConstraintDAO.updateById(job.getConstraint_id(), latestJob);
                             LOG.error("updated job state to {}", TagSyncState.ERROR);
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
@@ -122,8 +122,8 @@ public class DeployTagWorker implements Runnable {
             for (DeployConstraintBean job : jobs) {
                 LOG.info("process job: {}", job);
                 String lockName = String.format("DeployTagWorker-%s", job.getConstraint_id());
-                DeployConstraintBean latestJob = deployConstraintDAO.getById(job.getConstraint_id());
                 Connection connection = utilDAO.getLock(lockName);
+                DeployConstraintBean latestJob = deployConstraintDAO.getById(job.getConstraint_id());
                 if (connection != null) {
                     try {
                         processEachEnvironConstraint(latestJob);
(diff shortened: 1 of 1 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
```diff
@@ -122,14 +124,20 @@ private void processBatch() throws Exception {
...
+                        processEachEnvironConstraint(latestJob);
                     } catch (Exception e) {
-                        LOG.error("failed to process job: {}", job.toString(), e);
-                        job.setState(TagSyncState.ERROR);
-                        deployConstraintDAO.updateById(job.getConstraint_id(), job);
-                        LOG.error("updated job state to {}", TagSyncState.ERROR);
+                        LOG.error("failed to process job: {} Error {} stack {}", latestJob.toString(),
+                                ExceptionUtils.getRootCauseMessage(e), ExceptionUtils.getFullStackTrace(e));
+                        if (e instanceof SQLException) {
+                            // Don't do anything
```
euccas: when it's an SQL exception, shall we add some messages to the log?

rowoot: Line #132 should capture the error regardless. 
That said, updated code to just `if` instead of an empty `if` followed by an else

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
```diff
@@ -122,14 +124,17 @@ private void processBatch() throws Exception {
                 LOG.info("process job: {}", job);
                 String lockName = String.format("DeployTagWorker-%s", job.getConstraint_id());
                 Connection connection = utilDAO.getLock(lockName);
+                DeployConstraintBean latestJob = deployConstraintDAO.getById(job.getConstraint_id());
```
yongwen: do we need this? the job and job_id should be 1-1 and not mutable.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
```diff
@@ -122,14 +124,17 @@ private void processBatch() throws Exception {
...
                     try {
-                        processEachEnvironConstraint(job);
+                        processEachEnvironConstraint(latestJob);
                     } catch (Exception e) {
-                        LOG.error("failed to process job: {}", job.toString(), e);
-                        job.setState(TagSyncState.ERROR);
-                        deployConstraintDAO.updateById(job.getConstraint_id(), job);
-                        LOG.error("updated job state to {}", TagSyncState.ERROR);
+                        LOG.error("failed to process job: {} Error {} stack {}", latestJob.toString(), ExceptionUtils.getRootCauseMessage(e), ExceptionUtils.getStackTrace(e));
+                        if (!SQLException.class.isInstance(e)) {
```
yongwen: you can also do multiple catch block to make the logic a bit clear, like catch (SQLException e) then catch (Exception e) as a catch all. but your approach also works.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
```diff
@@ -122,14 +124,17 @@ private void processBatch() throws Exception {
...
+                            latestJob.setState(TagSyncState.ERROR);
+                            deployConstraintDAO.updateById(job.getConstraint_id(), latestJob);
```
yongwen: we use job and latestJob here. maybe they should be the same?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
rowoot: @sulixu mentioned that there is a good chance that `job` may get mutated by a worker by the time it enters this state. The idea is to acquire lock first, then query the job to make sure mutations on it don't happen.``

rowoot: nvm -- I agree w/ your comment. I'll revert the latest job code

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
rowoot: +1 will do

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/DeployTagWorker.java
rowoot: https://github.com/pinterest/teletraan/pull/705#discussion_r394589964

PR #706, opened 2020-03-18
fix exceptions when failing to create an identifier.

euccas:
This change fixes the incorrect error messages shown on Teletraan UI when a user create a new stage while Teletraan can't create a nimbus identifier.

image

### DIFF
--- a/deploy-board/deploy_board/webapp/helpers/nimbusclient.py
+++ b/deploy-board/deploy_board/webapp/helpers/nimbusclient.py
@@ -41,10 +41,20 @@ class NimbusClient(object):
         return self.handle_response(response)
 
     def create_one_identifier(self, data, token=None):
+        """
+        Create a Nimbus Identifier according to the input request data.
+        If the request data does not have all the information needed for creating a Nimbus identifier, this method will return None.
+        """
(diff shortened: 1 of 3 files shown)


deploy-board/deploy_board/webapp/helpers/nimbusclient.py
```diff
@@ -41,20 +41,41 @@ def get_one_identifier(self, name, token=None):
...
+        """
+        Create a Nimbus Identifier according to the input request data.
+        If the request data does not have all the information needed for creating a Nimbus identifier, this method will return None.
+        """
+
+        requiredParams = ['projectName', 'env_name', 'stage_name']
+        for param in requiredParams:
+            if data.get(param) is None or len(data.get(param)) == 0:
+                log.error("Missing %s in the request data, cannot create a Nimbus identifier" % param)
+                return None
```
rowoot: Why `return None`? Wouldn't it be helpful to throw an error and catch it in the controller?

euccas: changed to raise a Teletraan Exception.

PR #707, opened 2020-03-19
fix exceptions when failing to create an identifier.

euccas:
This change fixes the incorrect error messages shown on Teletraan UI when a user create a new stage while Teletraan can't create a nimbus identifier.

76919291-b7b12380-6885-11ea-90cc-8a4754d63548

### DIFF
--- a/deploy-board/deploy_board/webapp/helpers/nimbusclient.py
+++ b/deploy-board/deploy_board/webapp/helpers/nimbusclient.py
@@ -1,7 +1,7 @@
 """Helper class to connect Nimbus service"""
 import logging
 from decorators import singleton
-from deploy_board.settings import NIMBUS_SERVICE_URL, NIMBUS_SERVICE_VERSION, TELETRAAN_PROJECT_URL_FORMAT
+from deploy_board.settings import IS_PINTEREST, NIMBUS_SERVICE_URL, NIMBUS_SERVICE_VERSION, TELETRAAN_PROJECT_URL_FORMAT
 from exceptions import NotAuthorizedException, TeletraanException, FailedAuthenticationException
 import requests
 requests.packages.urllib3.disable_warnings()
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/webapp/helpers/nimbusclient.py
```diff
@@ -41,6 +41,18 @@ def get_one_identifier(self, name, token=None):
...
     def create_one_identifier(self, data, token=None):
+        """
+        Create a Nimbus Identifier according to the input request data.
+        If the request data does not have all the information needed for creating a Nimbus identifier, this method will raise a Teletraan Exception.
+        """
+        requiredParams = ['projectName', 'env_name', 'stage_name']
+        for param in requiredParams:
+            if data.get(param) is None or len(data.get(param)) == 0:
+                log.error("Missing %s in the request data, cannot create a Nimbus identifier" % param)
+                if IS_PINTEREST:
```
yongwen: this logic will only raise exception when IS_PINTEREST.  maybe always raise exception only modify exception msg to wrap "contect ....." under IS_PINTEREST

yongwen: or better yet, wrap all nimbus integration under IS_PINTEREST in the future.

deploy-board/deploy_board/webapp/helpers/nimbusclient.py
```diff
@@ -52,9 +64,15 @@ def create_one_identifier(self, data, token=None):
         payload['platformName'] = 'teletraan'
         payload['projectName'] = data.get('projectName')
 
+        cellName = None
         for property in data['propertyList']['properties']:
             if property['propertyName'] == 'cellName':
                 cellName = property['propertyValue']
+        if cellName is None:
+            log.error("Missing cellName in the request data, cannot create a Nimbus identifier")
+            if IS_PINTEREST:
```
yongwen: same comment as above.

deploy-board/deploy_board/webapp/helpers/nimbusclient.py
euccas: thanks. i updated the code to always raise an exception.

deploy-board/deploy_board/webapp/helpers/nimbusclient.py
euccas: updated.

PR #708, opened 2020-03-19
add and fix logs around worker's db connection logic.

euccas:
small changes for making sure worker logic has proper logs around getting and releasing db connections. each db connection attempt should have three log messages:
- successfully get lock
- successfully release lock
- failed to get lock

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
@@ -398,10 +398,12 @@ public class CommonHandler {
         String lockName = String.format("STATE_TRANSITION-%s", deployId);
         Connection connection = utilDAO.getLock(lockName);
         if (connection != null) {
+            LOG.debug(String.format("Successfully get lock: %s", lockName));
             try {
                 internalTransition(deployId, envBean);
             } finally {
                 utilDAO.releaseLock(lockName, connection);
+                LOG.debug(String.format("Successfully released lock: %s", lockName));
             }
         } else {
             LOG.warn(String.format("Failed to get lock: %s", lockName));
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
@@ -398,15 +398,15 @@ public class CommonHandler {
         String lockName = String.format("STATE_TRANSITION-%s", deployId);
         Connection connection = utilDAO.getLock(lockName);
         if (connection != null) {
-            LOG.debug(String.format("Successfully get lock: %s", lockName));
+            LOG.info(String.format("DB lock operation is successful: get lock %s", lockName));
             try {
                 internalTransition(deployId, envBean);
             } finally {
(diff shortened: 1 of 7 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -398,10 +398,12 @@ public void transitionDeployState(String deployId, EnvironBean envBean) throws E
         String lockName = String.format("STATE_TRANSITION-%s", deployId);
         Connection connection = utilDAO.getLock(lockName);
         if (connection != null) {
+            LOG.debug(String.format("Successfully get lock: %s", lockName));
```
yongwen: the prod env set the log level as INFO.  i think for lock msg, it might be ok to use LOG.info for the current critical debugging.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -398,10 +398,12 @@ public void transitionDeployState(String deployId, EnvironBean envBean) throws E
...
             try {
                 internalTransition(deployId, envBean);
             } finally {
                 utilDAO.releaseLock(lockName, connection);
+                LOG.debug(String.format("Successfully released lock: %s", lockName));
```
yongwen: maybe we can use a more "grep/searchable"-able msg for check for lock leak, something like "Lock Successfully released or get" in case the search does not support regex

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
euccas: we can change the log level to be DEBUG when we need debugging lock related issues. get/release locks happen frequently. currently teletraan already generates a large amount of INFO logs, that's why i use DEBUG level for these logs.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
euccas: these messages can be searched by "Successfully released lock", "Successfully get lock".

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
yongwen: my view is we are in active debugging of the lock leak issue so we can make it info for now and once stable then turn it to debug. and across the system there are some other place (see the other file in the PR) has the lock msg DEBUG.  and it will generally take some time to identify the leak so turning on DEBUG for prod temporarily might not be sufficient and need to restart for log level change.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
yongwen: it is better to search by one filter so you can see the get/release pair next to either to identify the leak.

euccas: i see what you mean. will make the change.

PR #710, opened 2020-03-27
Fix the deploy board webapp urls regexps to allow builds with dots in their name

pernu:
Currently, it's impossible to view the details of a build that has a name which contains a dot (for instance `counter-18.04` see https://jira.pinadmin.com/browse/CDP-1664), this simple fix should solve that.


yongwen:
here is an interesting discussion on allowing dot in url: https://beamtic.com/dots-in-urls  but i agree that it is ok to support that.

PR #718, opened 2020-05-27
warning message on pausing a deploy

ankilosaurus:


ankilosaurus:
Screen Shot 2020-05-27 at 12 32 04 PM

ankilosaurus:
> make the warning msg in RED and add more verbose msg saying something like "pause is not recommended. it should only be used in emergency case. say the implication...."

@yongwen I changed message to more verbose however, left color as-is to be consistent with other warnings/alerts in UI for ex: have warning yellow and error red.

ankilosaurus:
Changed to alert danger. looks better.
Screen Shot 2020-05-27 at 11 42 59 PM

yongwen:
verbing suggestion:  environment will not be deployed properly on new host(s) added manually or through ASG autoscaling .......

ankilosaurus:
> verbing suggestion: environment will not be deployed properly on new host(s) added manually or through ASG autoscaling .......

since this is merged, i have taken care of this in https://github.com/pinterest/teletraan/pull/721/files

PR #722, opened 2020-05-27
limit pause action only to env admin but still allow an operator to r…

ankilosaurus:
…esume


ankilosaurus:
Decide not to merge this change. Closing this issue.

PR #725, opened 2020-06-19
security vulnerability fixes

ankilosaurus:


yongwen:
as there are a few common lib upgrade and some is big version bump, please do more integration testing in integ env.

PR #737, opened 2020-08-05
Add basic validation to environment parameters

krockpot:
The API currently does no validation of the parameters sent to the environment details page. This data is rendered back to users so we should do some validation to ensure scripts cannot be injected. 

I didn't see where our HTML templates get expanded, but am happy to move the HMTL escape to wherever we build/render the HTML for the user instead of storing the strings that way. Let me know!

PR #741, opened 2020-08-12
Add basic validation to environment parameters 

krockpot:
The API currently does no validation of the parameters sent to the environment details page. This data is rendered back to users so we should do some validation to ensure scripts cannot be injected.

I didn't see where our HTML templates get expanded, but am happy to move the HMTL escape to wherever we build/render the HTML for the user instead of storing the strings that way. Let me know!

Github actions was added after https://github.com/pinterest/teletraan/pull/737 so the build seems to not want to trigger. Creating a new PR.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -91,15 +91,15 @@ public class EnvironBean implements Updatable, Serializable {
     private String description;
 
     @JsonProperty("buildName")
-    @Pattern(regexp="^[A-Za-z0-9_\\-]*$")
+    @Pattern(regexp="^[A-Za-z0-9_\\.\\/\\-]*$")
     private String build_name;
 
     @JsonProperty("branch")
-    @Pattern(regexp="^[A-Za-z0-9_\\-]*$")
+    @Pattern(regexp="^[A-Za-z0-9_\\.\\,\\/\\-]*$")
     private String branch;
 
     @JsonProperty("chatroom")
-    @Pattern(regexp="^[A-Za-z0-9_\\-]*$")
+    @Pattern(regexp="^[A-Za-z0-9_\\#\\,\\-]*$")
     private String chatroom;
 
     @JsonProperty("deployId")
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -235,6 +235,7 @@ public class EnvironBean implements Updatable, Serializable {
     }
 
     public void setDescription(String description) {
+        // Escape user input which could contain injected Javascript
         this.description = StringEscapeUtils.escapeHtml(this.description);
     }
 
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -15,12 +15,16 @@
  */
 package com.pinterest.deployservice.bean;
 
+import com.pinterest.deployservice.common.DeployInternalException;
+
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.hibernate.validator.constraints.NotEmpty;
 import org.hibernate.validator.constraints.Range;
 
 import java.io.Serializable;
+import javax.validation.constraints.Pattern;
 
 /**
  * Keep the bean and table in sync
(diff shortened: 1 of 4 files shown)


krockpot:
ready for review @yongwen  @ankilosaurus , not sure why the build didn't run. I have test results on the original PR (#737)

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -68,14 +72,17 @@
  * );
  */
 public class EnvironBean implements Updatable, Serializable {
+
```
ankilosaurus, line 75: white space

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -15,12 +15,16 @@
  */
 package com.pinterest.deployservice.bean;
 
+import com.pinterest.deployservice.common.DeployInternalException;
+
```
ankilosaurus, line 19: white space

ankilosaurus:
hi @krockpot thanks making this change. One question, with this change, it will return failure for existing environments? It would be good to test out how if behaves for existing env violating this pattern.

krockpot:
@ankilosaurus based on my experiments it seems teletraan doesn't even support environments or stages that aren't alphanumeric (I get error responses/failure codes, even without the new code). That validation is already done on the web UI / client side, this is just enforcing it on the API itself.

ankilosaurus:
> @ankilosaurus based on my experiments it seems teletraan doesn't even support environments or stages that aren't alphanumeric (I get error responses/failure codes, even without the new code). That validation is already done on the web UI / client side, this is just enforcing it on the API itself.

@krockpot sure, env/stage name might be in place in UI so, enforcing check in api is fine. However, there are other fields which are validated with this fix. May be we can just check if there aren't any existing violations before rolling out.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/exception/GenericExceptionMapper.java
```diff
@@ -48,6 +49,21 @@ public Response toResponse(Throwable t) {
                 sb.append("\n").append(sw.toString());
                 return Response.serverError().entity(sb.toString()).build();
             }
+        } else if (t instanceof ConstraintViolationException) {
```
ankilosaurus, line 52: I see a bit of duplicate code in toResponse method which i think could be extracted out and put in helper method. However, I see it's existing code, so leave it up to you.

ankilosaurus:
@krockpot Please verify the error displayed to user makes sense (client error with 400 error code).

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/exception/GenericExceptionMapper.java
```diff
@@ -48,6 +49,21 @@ public Response toResponse(Throwable t) {
...
+            StringBuilder sb = new StringBuilder();
+            if (t.getMessage() != null) {
+                sb.append("\nMessage: ").append(t.getMessage());
+            }
+
+            if (clientError.equals(Constants.CLIENT_ERROR_SHORT)) {
+                return Response.serverError().status(Response.Status.BAD_REQUEST).entity(sb.toString()).build();
```
ankilosaurus, line 59: In this case it's not server error, correct? Wondering what error message looks like in UI?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/exception/GenericExceptionMapper.java
```diff
@@ -48,6 +49,21 @@ public Response toResponse(Throwable t) {
...
+            } else {
+                StringWriter sw = new StringWriter();
+                PrintWriter pw = new PrintWriter(sw);
+                t.printStackTrace(pw);
+                sb.append("\n").append(sw.toString());
+                return Response.serverError().status(Response.Status.BAD_REQUEST).entity(sb.toString()).build();
```
ankilosaurus, line 65: check if this says server error with 400 code? In that case you should remove "serverError()"

krockpot: https://user-images.githubusercontent.com/5506038/89687929-f5a2f880-d8b5-11ea-8735-0bacad30f57a.png

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/exception/GenericExceptionMapper.java
krockpot: https://user-images.githubusercontent.com/5506038/89687929-f5a2f880-d8b5-11ea-8735-0bacad30f57a.png

krockpot:
> @krockpot Please verify the error displayed to user makes sense (client error with 400 error code).

how does this look @ankilosaurus https://user-images.githubusercontent.com/5506038/89687929-f5a2f880-d8b5-11ea-8735-0bacad30f57a.png

krockpot:
After discussing I have updated the error message. Please see the attached image.
Screen Shot 2020-08-19 at 11 47 53 AM

cc @ankilosaurus 

I'll sync with you offline around the other fields where we added validation.

ankilosaurus:
@krockpot thank you for making changes. Env/stage name is teletraan defined fields. Values for other fields are coming from different system(for ex: build/branch etc from scm, chatroom from slack etc.). So, it's good to double check what's allowed for those values.
Otherwise change looks good.

ankilosaurus:
To -> https://github.com/pinterest/teletraan/pull/745

PR #745, opened 2020-08-27
Knockpot sox validation

ankilosaurus:
Fix by @krockpot on environment fields validation.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -15,12 +15,15 @@
  */
 package com.pinterest.deployservice.bean;
 
+import com.pinterest.deployservice.common.DeployInternalException;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.hibernate.validator.constraints.NotEmpty;
 import org.hibernate.validator.constraints.Range;
 
 import java.io.Serializable;
+import javax.validation.constraints.Pattern;
 
 /**
  * Keep the bean and table in sync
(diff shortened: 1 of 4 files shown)


ankilosaurus:
From: https://github.com/pinterest/teletraan/pull/741

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/exception/GenericExceptionMapper.java
```diff
@@ -48,6 +49,13 @@ public Response toResponse(Throwable t) {
                 sb.append("\n").append(sw.toString());
                 return Response.serverError().entity(sb.toString()).build();
             }
+        } else if (t instanceof ConstraintViolationException) {
+            StringBuilder sb = new StringBuilder();
+            if (t.getMessage() != null) {
+                sb.append("\nMessage: ").append(t.getMessage());
+            }
+            sb.append("\nParameters in request violate configured constraints.");
```
yongwen, line 57: can we add what is the "configured constraint" is, like the regex pattern. otherwise, user will not have a clue or instruction to fix it.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -72,10 +75,12 @@
     private String env_id;
 
     @NotEmpty
+    @Pattern(regexp="^[A-Za-z0-9_\\-]*$")
```
yongwen, line 78: can we do a DB review on the existing data to see if there is any violation? otherwise, some existing name might not work.

ankilosaurus: yup, Jeremy and I discussed this in his original PR. I will double check.

ankilosaurus:
Screen Shot 2020-09-25 at 6 38 15 PM

krockpot (approved):

PR #746, opened 2020-08-31
Deploy agent v1.2.18

ankilosaurus:
deploy agent changes to include az and asg in ping request

Ready to merge.

### DIFF
--- a/deploy-agent/deployd/client/client.py
+++ b/deploy-agent/deployd/client/client.py
@@ -50,6 +50,7 @@ class Client(BaseClient):
             group_key = self._config.get_facter_group_key()
             az_key = self._config.get_facter_az_key()
             asg_tag_key = self._config.get_facter_asg_tag_key()
+            ec2_tags_key = self._config.get_facter_ec2_tags_key()
             keys_to_fetch = set()
             # facter call is expensive so collect all keys to fetch first
             if not self._hostname and name_key:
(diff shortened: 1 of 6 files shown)


ankilosaurus:
@jaymecox-pinterest

deploy-agent/deployd/client/client.py
```diff
@@ -29,31 +29,64 @@
...
             log.info("Use facter to get host info")
-            if not self._hostname and self._config.get_facter_name_key():
-                self._hostname = utils.get_info_from_facter(self._config.get_facter_name_key())
+            name_key = self._config.get_facter_name_key()
+            ip_key = self._config.get_facter_ip_key()
+            id_key = self._config.get_facter_id_key()
+            group_key = self._config.get_facter_group_key()
+            az_key = self._config.get_facter_az_key()
+            keys_to_fetch = set()
+            # facter call is expensive so collect all keys to fetch first
```
yongwen, line 53: How expensive it is? is it just reading from cache in file? If it is related to the facter cache file lock, can we only call the facter once during the first deployd get_info_from_facter and store these info in a deployd property file? assuming these info will never change after the host launched.

ankilosaurus: It's not related to cache file. For hostname, hostip, hostid, group, we currently exec "facter -p" 4 times(one for each). Jayme mentioned we can call "facter -p" with list of keys. So, converting those 4 calls into single call. 

Do we want to further put that in file and never call facter again?

yongwen: r+ now. we can iterate to improve later.

PR #749, opened 2020-09-09
use preparestatemet with parameter for sql statement

yongwen:
this is to prevent SQL injection by using preparestatement with parameter.


yurushao (approved):
Thanks for the fix. For other reviewers: `QueryRunner.query(...) eventually` uses PreparedStatement, see line 360: https://commons.apache.org/proper/commons-dbutils/apidocs/src-html/org/apache/commons/dbutils/QueryRunner.html

ankilosaurus (approved):

krockpot:
i see a bunch of other SQL queries that use string formatting in that file--please update them all to use prepared statements.

PR #761, opened 2020-09-16
only allow deploy trusted build URL if ensure_trusted flag is true

yongwen:
add the schema change to schema sql. this is need for a fresh DB.
and fix the whitespace in migration script.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/whitelists/Whitelist.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/whitelists/Whitelist.java
@@ -2,4 +2,5 @@ package com.pinterest.deployservice.whitelists;
 
 public interface Whitelist {
     Boolean approved(String attempt);
+    Boolean trusted(String attempt);
 }
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
@@ -27,6 +27,7 @@ import com.pinterest.deployservice.dao.AgentDAO;
 import com.pinterest.deployservice.handler.ConfigHistoryHandler;
 import com.pinterest.deployservice.handler.DeployHandler;
 import com.pinterest.deployservice.handler.EnvironHandler;
+import com.pinterest.deployservice.whitelists.Whitelist;
 import com.pinterest.teletraan.TeletraanServiceContext;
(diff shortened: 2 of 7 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/whitelists/Whitelist.java
```diff
@@ -2,4 +2,5 @@
 
 public interface Whitelist {
```
krockpot, line 3: while we're here do you think we could update whitelist->allowlist?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -212,16 +215,24 @@ public Response create(
...
+            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, 
+                String.format("Build name (%s) does not match stage config (%s).", 
+                    buildBean.getBuild_name(), envBean.getBuild_name()));
         }
+        // if the stage is not whitelisted (allow_private_build)
+        if(! envBean.getAllow_private_build()) { 
+            // only allow deploy if it is not private build
+            if (buildBean.getScm_branch().equals("private")) {
+                throw new TeletaanInternalException(Response.Status.BAD_REQUEST, 
+                    "This stage does not allow deploying a private build. Please Contact #teletraan to whitelist your stage for deploying private build");
```
krockpot, line 227: whitelist->allow

deploy-service/common/src/main/java/com/pinterest/deployservice/whitelists/Whitelist.java
yongwen: good point. will do

yongwen: decide to not do this re-factor at this time. this PR will focus on supporting the trusted URL. We can do the refactor in a later time.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
yongwen: see comments above.

krockpot (approved):
should we make “trustedBuild” instead something more concrete, like productionBuildLocations or something? in looking at the code its hard to logically understand the difference between the 2 lists passed in

ankilosaurus (approved):

PR #768, opened 2020-09-24
Deployd read shard data facter

ankilosaurus:
1. For Pinterest, read shard keys always from facter. 
2. For non-Pinterest, pass use-host-info to deployd and set shard keys (AZ, ASG, Stagetype) to use sharding feature.


yongwen:
my view is to keep this code in deploy-agent client so we can still support non pinterest/non-aws hosts.  and force pinterest deployd always start with --use-facter option.  this way, at least the local development (laptop) will still work, and other hosts (vagrant, aws) that don't have facter install.

yongwen (approved):

PR #773, opened 2020-09-30
Manual validation

ankilosaurus:


deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -200,6 +195,30 @@
...
+        }
+        String stageRegEx = "^[A-Za-z0-9_\\-]*$";
+        if (!this.stage_name.matches(stageRegEx)) {
+            throw new IllegalArgumentException(String.format("Stage name must match regex %s", stageRegEx));
+        }
+        String buildRegEx = "^[A-Za-z0-9_\\.\\/\\-]*$";
+        if (!this.build_name.matches(buildRegEx)) {
+            throw new IllegalArgumentException(String.format("Build name must match regex %s", buildRegEx));
+        }
+        String branchRegEx = "^[A-Za-z0-9_\\:\\.\\,\\/\\-]*$";
```
krockpot, line 212: nit: could make these static / consts.

could also probably unify / reuse the ones that are the same

krockpot (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
ankilosaurus: strings are immutable in java

krockpot (approved):

PR #774, opened 2020-10-05
Add egress support

bpalermo:
Add Nimbus egress support

### DIFF
--- a/deploy-board/deploy_board/settings.py
+++ b/deploy-board/deploy_board/settings.py
@@ -301,6 +301,8 @@ if IS_PINTEREST:
 
     # Nimbus service url
     NIMBUS_SERVICE_URL = os.getenv("NIMBUS_SERVICE_URL", None)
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/settings.py
```diff
@@ -301,6 +301,8 @@
 
     # Nimbus service url
     NIMBUS_SERVICE_URL = os.getenv("NIMBUS_SERVICE_URL", None)
+    NIMBUS_EGRESS_URL = os.getenv("NIMBUS_EGRESS_URL", None)
```
yongwen, line 304: can we just use "NIMBUS_SERVICE_URL" directly?

bpalermo: @yongwen we also need the EGRESS_URL so we can send the request to localhost:19193 but set the host header with NIMBUS_SERVICE_URL.

yongwen (approved):

PR #779, opened 2020-10-15
expose env system priority in config UI

yongwen:
the env Config UI will have a new field called "Sidecar System Priority" to expose it for sidecar specifically and readonly. this is an internal priority that service owner should not change by themself.

### DIFF
--- a/deploy-board/deploy_board/templates/configs/env_config.tmpl
+++ b/deploy-board/deploy_board/templates/configs/env_config.tmpl
@@ -248,6 +248,19 @@
                     </div>
                 </div>
                 <div class="form-group">
+                    <label for="syspriority" class="deployToolTip control-label col-xs-2"
+                        data-toggle="tooltip"
+                        title="System Priority for Sidecars">
+                        Sidecar System Priority
+                    </label>
+
+                    <div class="col-xs-4">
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/templates/configs/env_config.tmpl
```diff
@@ -247,6 +247,19 @@
...
                     </div>
                 </div>
+                <div class="form-group">
+                    <label for="syspriority" class="deployToolTip control-label col-xs-2"
+                        data-toggle="tooltip"
+                        title="System Priority for Sidecars">
+                        Sidecar System Priority
+                    </label>
+
+                    <div class="col-xs-4">
```
ankilosaurus, line 257: Can we just put the value in label?

ankilosaurus:
How about reject any API update on this field?

ankilosaurus (commented):
How about reject any API update on this field?

ankilosaurus (approved):

deploy-board/deploy_board/templates/configs/env_config.tmpl
yongwen: we can reserve the option to make the field writable in the future

PR #783, opened 2020-11-04
Use proxysql if provided in config

ankilosaurus:
Read config(tested earlier using hard coded proxy url/port in code) and use proxysql if configured else default to use zk config file.

Also, including non-blocking sidecar deploy changes from #775 for combined testing in Integ environment.


yongwen:
please break up unblocking change from the sqlproxy config change

PR #788, opened 2020-11-30
Agent count table cache implementation

ankilosaurus:
Context: When Teletraan agent service receives ping request from deployd running on host, agent service does DB lookups to figure out what needs to be deployed on that host. Once, agent service figures out an environment which needs to be deployed, agent service then look s for couple of other things for ex: max parallel threshold, deploy schedule, deploy constraint configurations etc. for that environment.
For max parallel threshold, agent service needs to check how many hosts are currently in process of deploying a given environment. To do this, agent service runs select count query to get number of hosts actively deploying.  This query is expensive specifically for an environment which gets deployed on large number of hosts (fleet wide deploys) which typically is the case for infrastructure services or sidecars.
We tested caching of select count queries using proxysql, however, it would break max parallel threshold guarantees. 

Fix:
Alternatively, we could maintain count in database which is this PR. Typically, we want to +1 count when environment starts (first 
 deploy stage) deploying on a given host and -1 count (last deploy stage) when environment finishes deploying on a given host.
Maintaining count separately, could be error prone. So, we invalidate count w/ ttl making count table more like cache instead of source of truth and in that case, we will +1 count when deploy starts on a hosts but instead of doing -1 when deploy ends, rely on cache expiration and refresh by repopulating count from db. Additionally, since we are invalidating count w/ ttl, we can update agent count and agent state without transaction.
Also, 1. cleaning up canDeploy api signature 2. revert proxysql setting

Testing...
1. Max parallel deploy of a service
2. Optimal cache ttl settings 1s vs 10s


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/ZKMysqlDataSourceFactory.java
```diff
@@ -70,8 +64,8 @@ public void setPool(String pool) {
     public BasicDataSource build() throws Exception {
         // get from local ZK maintained DB config file
         DBConfigReader reader = new DBConfigReader();
-        String host = this.proxyHost != null ? this.proxyHost : reader.getHost(replicaSet);
-        int port = this.proxyPort != null ? Integer.parseInt(this.proxyPort) : reader.getPort(replicaSet);
+        String host = reader.getHost(replicaSet);
```
yongwen, line 67: i think it is still good to keep the functionality to support proxy

yongwen (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/ZKMysqlDataSourceFactory.java
ankilosaurus: discussed on slack for the reason behind cleaning up.

PR #789, opened 2020-12-01
Default cluster type

ankilosaurus:
Changes:
- Introduces additional cluster/stage type "DEFAULT" (which implicitly defaults to PRODUCTION)
- Adds restriction on changing cluster/stage type once set to one of non-DEFAULT types
- Adds tooltip for stage type


yongwen (approved):
can we keep the proxy support?

PR #803, opened 2021-02-06
Pick up latest deploy on host when hostgroup is moved between stages of same envionment

ankilosaurus:
Test pending.


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -162,9 +162,20 @@ public String toString() {
             environ = existingAgentEnv.getOrDefault(report.getEnvId(), new EnvironBean());
```
ankilosaurus, line 162: todo: revert changes in this file as it's not part of this PR.

PR #806, opened 2021-03-02
Add Python 3 support to deployd

jogo:
Add python 3 support in a backwards compatible manor so Python 2 still works by using [futurize](https://python-future.org/).

Make sure all unit tests pass via running 'tox'

### DIFF
--- a/deploy-agent/deployd/client/client.py
+++ b/deploy-agent/deployd/client/client.py
@@ -175,8 +175,8 @@ class Client(BaseClient):
             if self._read_host_info():
                 reports = [status.report for status in env_reports.values()]
                 for report in reports:
-                    if isinstance(report.errorMessage, str):
-                        report.errorMessage = unicode(report.errorMessage, "utf-8")
+                    if isinstance(report.errorMessage, bytes):
+                        report.errorMessage = report.errorMessage.decode('utf-8')
 
                     # We ignore non-ascii charater for now, we should further solve this problem on
                     # the server side:
--- a/deploy-agent/deployd/common/config.py
+++ b/deploy-agent/deployd/common/config.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # Copyright 2016 Pinterest, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,13 +15,15 @@
 
 import logging
 import os
-from ConfigParser import SafeConfigParser
 
 from deployd.common.exceptions import DeployConfigException
 from deployd.common.types import DeployType
 from deployd.common.utils import exit_abruptly
 from deployd.types.deploy_stage import DeployStage
 from deployd.types.opcode import OperationCode
+
+from configparser import ConfigParser
+
 
 log = logging.getLogger(__name__)
 
@@ -36,7 +39,7 @@ class Config(object):
             self._config_reader = config_reader
             return
 
-        self._config_reader = SafeConfigParser()
+        self._config_reader = ConfigParser()
         if not filenames:
             return
 
--- a/deploy-agent/deployd/common/env_status.py
+++ b/deploy-agent/deployd/common/env_status.py
@@ -42,7 +42,7 @@ class EnvStatus(object):
             with self._lock, open(self._status_fn, 'r+') as config_fn:
                 data = json.load(config_fn)
                 log.debug('load status file: {}'.format(data))
-                envs = {key: DeployStatus(json_value=d) for key, d in data.iteritems()}
+                envs = {key: DeployStatus(json_value=d) for key, d in data.items()}
         except IOError:
             log.info("Could not find file {}. It happens when run deploy-agent the "
                      "first time, or there is no deploy yet.".format(self._status_fn))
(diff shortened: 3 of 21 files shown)


deploy-agent/deployd/client/client.py
```diff
@@ -175,8 +175,8 @@ def send_reports(self, env_reports=None):
             if self._read_host_info():
                 reports = [status.report for status in env_reports.values()]
                 for report in reports:
-                    if isinstance(report.errorMessage, str):
-                        report.errorMessage = unicode(report.errorMessage, "utf-8")
+                    if isinstance(report.errorMessage, bytes):
```
yongwen, line 178: will this work for both py2 and py3?  surprising to see the type "str" and "bytes" are the same

deploy-agent/deployd/common/config.py
```diff
@@ -14,14 +15,16 @@
...
 import os
-from ConfigParser import SafeConfigParser
 
 from deployd.common.exceptions import DeployConfigException
 from deployd.common.types import DeployType
 from deployd.common.utils import exit_abruptly
 from deployd.types.deploy_stage import DeployStage
 from deployd.types.opcode import OperationCode
 
+from configparser import ConfigParser
```
yongwen, line 25: just double check it is also supported in py2 and it is a safe.

deploy-agent/deployd/common/env_status.py
```diff
@@ -42,13 +42,13 @@ def load_envs(self):
...
-                envs = {key: DeployStatus(json_value=d) for key, d in data.iteritems()}
+                envs = {key: DeployStatus(json_value=d) for key, d in data.items()}
         except IOError:
             log.info("Could not find file {}. It happens when run deploy-agent the "
                      "first time, or there is no deploy yet.".format(self._status_fn))
             return {}
-        except Exception as e:
-            log.error(e.message)
+        except Exception:
+            log.exception("Something went wrong in load_envs")
```
yongwen, line 51: would be nice to log the exception messsage as the original intent

deploy-agent/deployd/common/utils.py
```diff
@@ -45,7 +46,7 @@ def exit_abruptly(status=0):
 
 def touch(fname, times=None):
     try:
-        with file(fname, 'a'):
+        with open(fname, 'a'):
```
yongwen, line 49: can we still use "file" in py3? assuming "open" will actually open the file handle and "file" does not.

deploy-agent/requirements.txt
```diff
@@ -3,10 +3,9 @@
 # tests
 tox==1.6.1
 coverage
-pytest==2.3.5
+pytest
```
yongwen, line 6: it is better to pin down a version

deploy-agent/setup.py
```diff
@@ -26,10 +26,12 @@
 
 install_requires = [
     "requests==2.20.0",
-    "gevent>=1.0.2,<=1.2.2",
+    "gevent>=1.0.2,<=1.2.2; python_version < '3'",
+    "gevent>=1.0.2,<=1.5.0; python_version >= '3'",
```
yongwen, line 30: in the past,  upgrading gevent caused issues. just need to do more testing on the new version of gevent in py3.

yongwen (commented):
Thanks Joe.

deploy-agent/deployd/client/client.py
jogo: yes, under PY2 `bytes==str`

deploy-agent/deployd/common/config.py
jogo: yup confirmed by running unit tests work on PY2 and PY3

deploy-agent/deployd/common/env_status.py
jogo: log.exception already does that

deploy-agent/deployd/common/utils.py
jogo: file==open 

https://portingguide.readthedocs.io/en/latest/builtins.html#removed-file

https://docs.python.org/2.7/library/functions.html#file

deploy-agent/requirements.txt
jogo: for testing only I would rather use the latest version, but happy to pin to a newer version of pytest if you want

deploy-agent/setup.py
jogo: note the update is only for PY3 and not PY2.  Do you have any links to previous issues?

deploy-agent/requirements.txt
yongwen: can you pin down to a latest version? i think it is better to have deterministic test

jogo: I don't think it is actually, we don't pin coverage either

yongwen: i think we should always pin down to a version for pip. we are not using coverage, so it is ok to remove this dependency all together.

yongwen (approved):

PR #807, opened 2021-03-10
Host cleanup

ankilosaurus:
Problem:
- hosts deleted from outside of Teletraan (ex: deleted using aws console or cli etc.) are not cleaned up from Teletraan hosts/agents table. (If host is terminated from teletraan using Terminate/Force Terminate option works fine)
Current query - "SELECT DISTINCT hosts.* FROM hosts INNER JOIN hosts_and_envs ON hosts.host_name=hosts_and_envs.host_name WHERE hosts.last_update<?" - does not cleanup all the stale hosts.

Due to this hosts and agents table keeps growing with those dead hosts. 

Fix:
Remove stale hosts/agents based on last update timestamp from hosts and agents tables.

Test pending.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
@@ -162,10 +162,18 @@ public class GoalAnalyst {
             environ = existingAgentEnv.getOrDefault(report.getEnvId(), new EnvironBean());
         }
 
+        int getUninstallPriority(EnvironBean environ) {
+            Integer systemPriority = environ.getSystem_priority();
+            if (systemPriority != null) {
+                return systemPriority;
+            }
+            return environ.getPriority().getValue();
+        }
+
         @Override
         public int compareTo(UninstallCandidate uninstallCandidate) {
-            int priority1 = environ.getPriority().getValue();
-            int priority2 = uninstallCandidate.environ.getPriority().getValue();
+            int priority1 = getUninstallPriority(environ);
+            int priority2 = getUninstallPriority(uninstallCandidate.environ);
 
             return priority2 - priority1;
         }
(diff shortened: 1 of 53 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -162,9 +162,20 @@ public String toString() {
...
+            if (systemPriority != null) {
+                return systemPriority;
+            }
+            return environ.getPriority().getValue();
+        }
+
         @Override
         public int compareTo(UninstallCandidate uninstallCandidate) {
-            return 0;
+            int priority1 = getUninstallPriority(environ);
```
ankilosaurus, line 175: not part of the change. will revert

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/SimpleAgentJanitor.java
```diff
@@ -43,30 +42,25 @@
...
     // remove the stale host from db
     void removeStaleHost(String id) throws Exception {
-        try {
-            hostDAO.deleteAllById(id);
-            hostAgentDAO.delete(id);
-            LOG.info("AgentJanitor delete all records for host {}.", id);
-        } catch (Exception e) {
-            LOG.error("AgentJanitor Failed to delete all records for host {}", id, e);
-        }
+        LOG.info(String.format("Delete records of stale host {}", id));
```
disparter, line 62: Without this try catch the exception will be handled in 

SimpleAgentJanitor Run.
LOG.error("SimpleAgentJanitor Failed.", t);

Is this what we want?

ankilosaurus: actually exception is caught in removeHost. but still better to catch uncaught. updated diff.

disparter (approved):

yongwen (approved):

PR #808, opened 2021-03-13
Upperbound for max parallel threshold and make it configurable

ankilosaurus:
Test pending

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
@@ -94,6 +94,7 @@ public class PingHandler {
     private LoadingCache<String, DeployBean> deployCache;
     private List<PingRequestValidator> validators;
     private Long agentCountCacheTtl;
+    private Long maxParallelThreshold;
 
     public PingHandler(ServiceContext serviceContext) {
         agentDAO = serviceContext.getAgentDAO();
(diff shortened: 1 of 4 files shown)


disparter:
Can you improve with a test plan description? Thank you

yongwen (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -227,7 +229,7 @@ boolean canDeploy(EnvironBean envBean, String host, AgentBean agentBean) throws
         String envId = envBean.getEnv_id();
         AgentCountBean agentCountBean = agentCountDAO.get(envId);
         long totalNonFirstDeployAgents = (isAgentCountValid(envId, agentCountBean) == true) ? agentCountBean.getExisting_count() : agentDAO.countNonFirstDeployingAgent(envId);
-        long parallelThreshold = getFinalMaxParallelCount(envBean, totalNonFirstDeployAgents);
+        long parallelThreshold = Math.min(this.maxParallelThreshold, getFinalMaxParallelCount(envBean, totalNonFirstDeployAgents));
```
disparter, line 232: (Opnion) This Line should be covered by Unit Testing.

disparter (approved):
I believe it's good to be land it as it's but at least a new Unit testing (this can be done later together with the other points of parallelization if they will be required

Issue #700, opened 2020-01-09
Missing migrations from the schema upgrade script.


mignaulo:
Hello,
We have a fairly old Teletraan setup that we are planning to update to the latest version. This requires us to upgrade the MySQL schema. We have been doing testing the MySQL upgrade script and comparing them to https://github.com/pinterest/teletraan/blob/master/deploy-service/common/src/main/resources/sql/deploy.sql and noticed some discrepancies.

After some sleuthing, I tracked the changes down to this PR:  https://github.com/pinterest/teletraan/pull/491 
That PR updated the complete schema without creating a new schema_version. These are the changes missing: https://github.com/pinterest/teletraan/pull/491/files#diff-380d85d61edaf8019bd4a4cd16a48c36
I believe this migration would in theory live somewhere between schema_version 1 and 2.  

We can definitely patch our setup without trouble, but I figured I would raise this issue regardless. I could create a PR to update the MySQL upgrade script to handle this and help others that might run into this. I could add the missing migrations to the end of `tools/mysql/schema-update-1.sql` or the beginning of `tools/mysql/schema-update-2.sql`. Thoughts?

PR #809, opened 2021-03-18
Fix the first run using docker

disparter:
Fixed bug on schema-updates

Removed the extra comma in the schema-update-8.sql (syntax issue)
Changed the NO-BREAK-SPACE to a Space in schema-update-(11-7-9-8)

Expectations
In the first run of Teletraan, after create database it has to be on version 11

Test Plan
1) docker-compose up --build
2) docker exec teletraan_pinterest_mysql /var/teletraan/tools/deploydb.sh -f /var/tools/check_version.sql

### DIFF
--- a/tools/mysql/schema-update-11.sql
+++ b/tools/mysql/schema-update-11.sql
@@ -4,7 +4,7 @@
 
 # This script upgrade DB schema from version 10 to version 11
 
-ALTER TABLE environs MODIFY COLUMN stage_type VARCHAR(32) NOT NULL DEFAULT 'DEFAULT';
+ALTER TABLE environs MODIFY COLUMN stage_type VARCHAR(32) NOT NULL DEFAULT 'DEFAULT';
 
 -- make sure to update the schema version to 11
 UPDATE schema_versions SET version=11;
--- a/tools/mysql/schema-update-7.sql
+++ b/tools/mysql/schema-update-7.sql
@@ -4,7 +4,7 @@
 
 # This script upgrade DB schema from version 6 to version 7
 
-ALTER TABLE environs MODIFY chatroom VARCHAR(128) DEFAULT NULL;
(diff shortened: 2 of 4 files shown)


disparter:
I think this PR fix this issue:
https://github.com/pinterest/teletraan/issues/700
But I will double check tomorrow

deploy-agent/deployd/common/config.py
```diff
@@ -252,7 +252,7 @@ def get_verify_https_certificate(self):
         return self.get_var('verify_https_certificate', 'False')
 
     def get_deploy_agent_version(self):
-        return self.get_var('deploy_agent_version', None)
+        return self.get_var('deploy_agent_version', '1.2.24')
```
yongwen, line 255: if we hardcode the version here, do we need to update this everytime we upgrade deploy_agent? 
maybe there is a better way not to hardcode the version here.

deploy-agent/deployd/conf/agent.conf
```diff
@@ -52,3 +52,4 @@ teletraan_service_token =
 # Verify the API HTTPS certificate chain
 verify_https_certificate = False
 
+deploy_agent_version = 1.2.24
```
yongwen, line 55: same comment as above

deploy-agent/deployd/common/config.py
ankilosaurus: what's the reason for change in this file in particular? 
Note, this affects Prod not just macos and could give wrong version if not populated.
I think better to default to None/Unknown if not populated instead of populating version which might be wrong.

disparter:
Removed the hardcoded version and reusing version declaration in init
that solves on Mac and we will only changing version in the same place

robbintt (approved):
lgtm

PR #813, opened 2021-04-02
Max parallel upperbound

disparter:
[DO NOT MERGE IT - SCRATCH PHASE]

Summary:
[Enforce maximum parallelization factor for large stages (based on N number of instances)]
  - For single side car
  - For multiple sidecars
  - For all configured stages
   
   
Test Plan:
Create unit tests to clear the math behind parallelization counts
Deploy N numbers above constraints and see if holds under all parallelization rules

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
@@ -94,6 +94,7 @@ public class PingHandler {
     private LoadingCache<String, DeployBean> deployCache;
     private List<PingRequestValidator> validators;
     private Long agentCountCacheTtl;
+    private Long maxParallelThreshold;
 
     public PingHandler(ServiceContext serviceContext) {
         agentDAO = serviceContext.getAgentDAO();
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
@@ -229,7 +229,7 @@ public class PingHandler {
         String envId = envBean.getEnv_id();
         AgentCountBean agentCountBean = agentCountDAO.get(envId);
         long totalNonFirstDeployAgents = (isAgentCountValid(envId, agentCountBean) == true) ? agentCountBean.getExisting_count() : agentDAO.countNonFirstDeployingAgent(envId);
-        long parallelThreshold = Math.min(this.maxParallelThreshold, getFinalMaxParallelCount(envBean, totalNonFirstDeployAgents));
+        long parallelThreshold = extracted(envBean, totalNonFirstDeployAgents);
 
         try {
             //Note: This count already excludes first deploy agents, includes agents in STOP state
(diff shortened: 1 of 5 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -227,7 +229,7 @@ boolean canDeploy(EnvironBean envBean, String host, AgentBean agentBean) throws
         String envId = envBean.getEnv_id();
         AgentCountBean agentCountBean = agentCountDAO.get(envId);
         long totalNonFirstDeployAgents = (isAgentCountValid(envId, agentCountBean) == true) ? agentCountBean.getExisting_count() : agentDAO.countNonFirstDeployingAgent(envId);
-        long parallelThreshold = getFinalMaxParallelCount(envBean, totalNonFirstDeployAgents);
+        long parallelThreshold = extracted(envBean, totalNonFirstDeployAgents);
```
ankilosaurus, line 232: what's extracted ?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -822,4 +825,20 @@ public final static int getFinalMaxParallelCount(EnvironBean environBean, long t
         }
         return ret;
     }
+
+    public final static int getFinalMaxParallelCount(EnvironBean envBean, long totalNonFirstDeployAgents, int maxParallelThreshold) throws Exception{
+        return Math.min(maxParallelThreshold, getFinalMaxParallelCount(envBean, totalNonFirstDeployAgents));
+    }
+
+    public final static int getFinalMaxParallelCount(DeployConstraintBean deployConstraintBean, EnvironBean envBean, int totalHosts) throws Exception{
```
ankilosaurus, line 833: where is this used?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
disparter: This PR is already closed after our discussion.
But I extracted the method to be able to unit testing the calculation without the need of complex testing for this.
The name was unhappy.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
disparter: This is the extracted method name from the above question. The refactor was misplaced.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
ankilosaurus: yeah, I realized that it's closed. Please ignore comments :)

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
ankilosaurus: plz ignore

PR #815, opened 2021-04-07
Update requirements, tox.ini to enable py3 support

amckenna-pinterest:


disparter (changes requested):
Please, share a Reason for the PR as a Summary.
Also, can you provide a test plan?
I understand two different reasons, the python3 update + splitting dependencies from core context to test only and you also put some syntax improvements.
Maybe can you split in two different PRS? (Not a requirement)

deploy-agent/tox.ini
```diff
@@ -1,10 +1,10 @@
 [tox]
-envlist=py27,py3
+envlist=py37
```
yongwen, line 2: i think we still need to support py2 for a while. so it is better to keep both py2 and py3

deploy-agent/requirements.txt
```diff
@@ -1,15 +1,7 @@
...
-mock==1.0.1
-flake8==2.5.4
-pep8
 PyYAML==5.3.1
 zipp==1.2.0
 configparser==4.0.2
 python-daemon==2.0.6
-setuptools==44.1.1
-
+setuptools==54.2.0
```
yongwen, line 7: did we test this version is available in py2?

yongwen (approved):

disparter (approved):

PR #819, opened 2021-04-10
show empty table when there is an exception during getting commit info from Phab

liyaqin1:


disparter (approved):
It will be nice to summarize the change in the PR.

ankilosaurus (changes requested):
hi @liyaqin1 , can you please update description with 1. Problem 2. Fix.

Also, it seems you are just throwing an exception with fixed message, but do we know what's the actual problem? 
It would be good to fetch error code and response from SCM.

liyaqin1:
@ankilosaurus @disparter thanks for the comments.
Right now, the code will return empty table if there is an exception.
Is there an example of returning useful info to UI from backend for me to take a look? Thanks.

PR #825, opened 2021-04-22
Changing HostType Identifier for backend on Cluster Configurations views

disparter:
Using ID instead of Abstract Name to avoid overwriting HostType in User Experience

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -457,7 +457,7 @@
                 currentCell: currentCluster.cellName,
                 hostTypeOptions: info.hostTypes.map(function(item, idx) {
                     return {
-                        value: item.abstract_name,
+                        value: item.id,
                         text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",
                         isSelected: item.abstract_name === currentCluster.hostType
                     }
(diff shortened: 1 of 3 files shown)


deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -457,7 +457,7 @@ <h4 class="panel-title pull-left pointer-cursor">
                 currentCell: currentCluster.cellName,
                 hostTypeOptions: info.hostTypes.map(function(item, idx) {
                     return {
-                        value: item.abstract_name,
+                        value: item.id,
```
ankilosaurus, line 460: @disparter how will UI look like with this change? I understand it will show IDs in dropdown instead of abstract_name/provider_name?

Ideal state: drop down should show -> Abstract name or Provider name or both

disparter: Value is the field that is sent into the form
    <input value="">
And the text is the value that will be shown to user
So in the case is:
 text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -457,7 +457,7 @@ <h4 class="panel-title pull-left pointer-cursor">
...
                         text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",
                         isSelected: item.abstract_name === currentCluster.hostType
```
ankilosaurus, line 462: since we are at it can we change this text to a format:
<Provider_name> (Description: <Abstract_name>, Spec: <8 cores etc...>)

this applied to all 3 files.

ankilosaurus: @yongwen what do you think ?

ankilosaurus: ah, makes sense.

ankilosaurus (approved):

ankilosaurus:
Please add tests

PR #829, opened 2021-04-28
Opensource cleanup

disparter:
Summary:

Removing Pinterest specifics.
1) Cluster management for everyone
2) Deploy Board many urlpatterns
3) No contact #teletraan channel
4) Feedback Modal Enabled
5) Alarm Button Enabled
6) Deploy Addons for all
7) Cold freeze enabling
8) Enabling Hotfixes

Test Plan:
Deployd and Deploy-Board have to work in Pinterest environment and outside as well.
Check for import errors

### DIFF
--- a/deploy-agent/deployd/__init__.py
+++ b/deploy-agent/deployd/__init__.py
@@ -12,9 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
-# Pinterest specific settings
-IS_PINTEREST = True if os.getenv("IS_PINTEREST", "false") == "true" else False
-
 __version__ = '1.2.25'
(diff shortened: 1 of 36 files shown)


deploy-agent/deployd/__init__.py
```diff
@@ -12,9 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
-# Pinterest specific settings
-IS_PINTEREST = True if os.getenv("IS_PINTEREST", "false") == "true" else False
```
ankilosaurus, line 18: what's reason for removing this flag and related checks?

disparter: In oder to achieve the same goal, we should parametrize this entries so they are not attached to Pinterest environment at all.
For example, better than say: "This only works for pinterest"
is to say: "Ok you need to configure X in order to get X to working, for example, cluster configs or Phobo"

PR #830, opened 2021-04-28
Removed from Deploy-Agent setup, pinterest specific packages

disparter:
Summary:
In order to prevent a dependency confusion in deploy-agent environment exploring non existing dependencies on a non Pinterest environment.
Removed pinlogger and pinstatd packages from setup. 
Test Plan:
1) deploy-agent install happens safely even in Pinterest environment


ankilosaurus:
plz bump deploy-agent patch version so we can test in puppet canary.

ankilosaurus (approved):

PR #831, opened 2021-05-04
Store ping timestamp per host separately.

ankilosaurus:
Problem:
----------
Host table keeps inventory of all the hosts registered to Teletraan.
1. It keeps timestamp of all the registered hosts -> used by AgentJanitor to cleanup unreachable or dead hosts.
2. It also keep mapping between host and group. (For example: helloworlddummyservice-server-dev-0a01def3 <> CMP, helloworlddummyservice-server-dev-0a01def3 <> mcrouter-prod-sidecar, ... etc.) -> used in UI/API to populate hosts which are part of the specific environment.

Due to this timestamp needs to be updated per host <-> group mapping for a given host instead of just per host.

Fix:
----
1. Extend existing hosts_and_agents table to keep per host data (specifically timestamp). 
With this change, 
- last_update field of hosts table will not be updated on every ping. Instead last_update in hosts_and_agents table will be updated on every ping once per host id.
- host <-> group mapping will be inserted once for new host and then after for any addition/deletion of host <-> group mapping.
This potentially reduces inserts by 7 folds per host per ping.
(Note: fields from hosts table are not removed for backward compatibility)

2. Improved on AgentJanitor logic by removing group level processing as that loop iterate through all the groups in the system to find unreachable hosts -> which I think is a processing waste. 
AgentJanitor will continue to go through all the stale hosts in the system with added asg info in hosts_and_agents table so as to pull host launch latency threshold config from ASG for a given stale host.

--
Other fields - state, can_retire - can also be moved to host_and_agents table as those are also per host. However, as that requires substantial changes at API level and also client side,  those changes are not included in the scope.

Tests:
---
1. Verify host_name, host_id, ip, asg, create_date, last_update, agent_version of a host in hosts_and_agents table.
2. Verify last_update field is not updated in hosts table on every ping if there is no change in host <-> group mapping.
3. Verify that new deploy groups are updated in hosts table.
4. Verify that stale deploy groups are removed from hosts table.
2. Terminate host part of ASG and verify it's cleaned up.
2. Remove deployd from host part of ASG and verify it's marked unreachable.
3. Terminate standalone hosts and verify it's cleaned up.
4. Remove deployd from standalone host and verify it's marked unreachable.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostAgentDAOImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostAgentDAOImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Pinterest, Inc.
+ * Copyright 2016 Pinterest, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,10 +15,11 @@
  */
 package com.pinterest.deployservice.db;
 
-
 import com.pinterest.deployservice.bean.HostAgentBean;
-import com.pinterest.deployservice.dao.HostAgentDAO;
+import com.pinterest.deployservice.bean.HostState;
 import com.pinterest.deployservice.bean.SetClause;
+import com.pinterest.deployservice.dao.HostAgentDAO;
+
 import org.apache.commons.dbcp.BasicDataSource;
 import org.apache.commons.dbutils.QueryRunner;
 import org.apache.commons.dbutils.ResultSetHandler;
@@ -30,10 +31,14 @@ import java.util.List;
 import java.util.Set;
 
 public class DBHostAgentDAOImpl implements HostAgentDAO {
-    private static final String INSERT_HOST_AGENT = "INSERT INTO hosts_and_agents SET %s ON DUPLICATE KEY UPDATE %s";
+    private static final String INSERT_HOST_TEMPLATE = "INSERT INTO hosts_and_agents SET %s ON DUPLICATE KEY UPDATE %s";
+    private static final String UPDATE_HOST_BY_ID = "UPDATE hosts_and_agents SET %s WHERE host_id=?";
     private static final String DELETE_HOST_BY_ID = "DELETE FROM hosts_and_agents WHERE host_id=?";
-    private static final String GET_HOST_BY_ID = "SELECT * FROM hosts_and_agents WHERE host_id=?";
-    private static final String GET_HOSTS_BY_AGENT = "SELECT * FROM hosts_and_agents WHERE agent_version=? ORDER BY host_id LIMIT ?,?";
+    private static final String GET_HOST_BY_NAME = "SELECT * FROM hosts_and_agents WHERE host_name=?";
+    private static final String GET_HOST_BY_HOSTID = "SELECT * FROM hosts_and_agents WHERE host_id=?";
+    private static final String GET_STALE_HOST = "SELECT DISTINCT hosts_and_agents.* FROM hosts_and_agents WHERE hosts_and_agents.last_update<?";
+    private static final String GET_STALE_ENV_HOST = "SELECT DISTINCT hosts_and_agents.* FROM hosts_and_agents INNER JOIN hosts_and_envs ON hosts_and_agents.host_name=hosts_and_envs.host_name WHERE hosts_and_agents.last_update<?";
+    private static final String GET_HOSTS_BY_AGENT = "SELECT * FROM hosts_statuses WHERE agent_version=? ORDER BY host_id LIMIT ?,?";
 
     private BasicDataSource dataSource;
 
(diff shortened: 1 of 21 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -144,17 +144,50 @@ public DeployBean load(String deployId) throws Exception {
...
+                LOG.warn("Remove host {} to group {} mapping", hostName, recordedGroup);
                 this.hostDAO.removeHostFromGroup(hostId, recordedGroup);
             }
         }
+    }
 
-
+    void updateHostStatus(String hostId, String hostName, String hostIp, String agentVersion, String asg, Set<String> groups) throws Exception {
+        HostAgentBean hostAgentBean = hostAgentDAO.getHostById(hostId);
+        if (hostAgentBean == null) {
```
disparter, line 172: I prefer to split this function into 2 more to improve code readness. 
This isn't a requirement we can discuss this if want.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
```diff
@@ -49,18 +50,23 @@ public AgentJanitor(ServiceContext serviceContext, int minStaleHostThreshold,
...
             Collection<String> resultIds = rodimusManager.getTerminatedHosts(Collections.singletonList(hostId));
             if (resultIds.isEmpty()) {
                 terminatedHosts.remove(hostId);
             }
         }
+        return terminatedHosts;
+    }
+
+    private void processStaleHosts(Collection<String> staleHostIds, boolean isMarkedUnreachable) throws Exception {
+        if (staleHostIds.isEmpty()) {
```
disparter, line 65: This collection can be nullable? if yes I guess you could check using Collections

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
```diff
@@ -87,113 +93,60 @@ private void processStaleHosts(Collection<String> staleHostIds, boolean isMarked
     }
 
     private Long getInstanceLaunchGracePeriod(String clusterName) throws Exception {
-        Long launchGracePeriod = rodimusManager.getClusterInstanceLaunchGracePeriod(clusterName);
+        Long launchGracePeriod = (clusterName != null) ? rodimusManager.getClusterInstanceLaunchGracePeriod(clusterName) : null;
```
disparter, line 96: You don't need parenthesis.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
```diff
@@ -87,113 +93,60 @@ private void processStaleHosts(Collection<String> staleHostIds, boolean isMarked
...
+        long current_time = System.currentTimeMillis();
         // If host fails to ping for longer than min stale threshold,
         // either mark them as UNREACHABLE, or remove if confirmed with source of truth
         long minThreshold = current_time - minStaleHostThreshold;
-        List<HostBean> minStaleHosts = hostDAO.getStaleHosts(minThreshold);
+        List<HostAgentBean> minStaleHosts = hostAgentDAO.getStaleHosts(minThreshold);
         Set<String> minStaleHostIds = new HashSet<>();
-        for (HostBean host : minStaleHosts) {
-            minStaleHostIds.add(host.getHost_id());
+        for (HostAgentBean hostAgentBean : minStaleHosts) {
```
disparter, line 108: What about using lambdas in these 2 fours?
Maybe can improve readability.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
```diff
@@ -87,113 +93,60 @@ private void processStaleHosts(Collection<String> staleHostIds, boolean isMarked
...
         Set<String> staleHostIds = new HashSet<>();
-        for (HostBean host : maxStaleHosts) {
+        
+        for (HostAgentBean hostAgentBean : maxStaleHosts) {
+            Long launchGracePeriod = getInstanceLaunchGracePeriod(hostAgentBean.getAuto_scaling_group());
+            HostBean host = hostDAO.getHostsByHostId(hostAgentBean.getHost_id()).get(0);
             if (host.getState() == HostState.PROVISIONED) {
-                if (current_time - host.getLast_update() >= maxLaunchLatencyThreshold) {
-                    staleHostIds.add(host.getHost_id());
+                if (current_time - hostAgentBean.getLast_update() >= launchGracePeriod) {
```
disparter, line 128: Maybe you can merge the two ifs.
Also can extract the ifelse logic as well.
Not a requirement, just an option

ankilosaurus: thought of not touching this part on the first go but I think it makes sense.

ankilosaurus: diff updated.

disparter (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostAgentDAOImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Pinterest, Inc.
+ * Copyright 2016 Pinterest, Inc.
```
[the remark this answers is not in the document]
ankilosaurus: actually, it was a typo. this file was introduced last year.

yongwen (approved):

PR #833, opened 2021-05-10
Fix bugfix on hosttype in cluster_configuration with default hosttype

disparter:
After changing Cluster API to use ID instead of Abstract Name, this is an adjusment to the UI to send ID instead of Abstract name on cluster configurations

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -495,7 +495,7 @@
                         isSelected: item.abstract_name === currentCluster.securityZone
                     }
                 }),
-                selectedHostTypeValue: currentCluster.hostType,
+                selectedHostTypeValue: capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === currentCluster.hostType).id,
                 selectedSecurityZoneValue: currentCluster.securityZone,
                 selectedPlacements: currentPlacements,
                 showBaseImageHelp: false,
@@ -718,7 +718,8 @@
                     var scope = this;
                     var provider = capacitySetting.currentProvider;
                     value = 'cmp_base-ebs';
-                    if (this.selectedHostTypeValue.startsWith('EbsCompute')) {
+                    const selectedHostTypeObject = capacityCreationInfo.hostTypes.find(hostType => hostType.id === this.selectedHostTypeValue);
+                    if (selectedHostTypeObject.abstract_name.startsWith('EbsCompute')) {
                         capacitySetting.imageNameValue = value;
                         capacitySetting.imagenames = info.baseImageNames.map(
                             function (o) {
--- a/deploy-board/deploy_board/templates/configs/new_capacity.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity.html
@@ -135,7 +135,7 @@ var capacitySetting = new Vue({
                 isSelected: item.abstract_name === capacityCreationInfo.defaultSeurityZone
             }
         }),
-        selectedHostTypeValue: capacityCreationInfo.defaultHostType,
+        selectedHostTypeValue: capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType).id,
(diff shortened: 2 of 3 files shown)


deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -373,7 +373,8 @@ <h4 class="panel-title pull-left">Capacity</h4>
             var scope = this;
             var provider = capacitySetting.currentProvider;
             value = 'cmp_base-ebs-18.04';
-            if (this.selectedHostTypeValue.startsWith('EbsCompute')) {
+            const selectedHostTypeObject = capacityCreationInfo.hostTypes.get({"id":this.selectedHostTypeValue});
```
ankilosaurus, line 376: why use get here and find at other places?

ankilosaurus: never mind.

deploy-board/deploy_board/templates/configs/new_capacity.html
```diff
@@ -135,7 +135,7 @@ <h4 class="panel-title pull-left pointer-cursor">Basic Settings</h4>
                 isSelected: item.abstract_name === capacityCreationInfo.defaultSeurityZone
             }
         }),
-        selectedHostTypeValue: capacityCreationInfo.defaultHostType,
+        selectedHostTypeValue: capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType).id,
```
ankilosaurus, line 138: can we not use abstract_name here?

ankilosaurus: If I create a host type with same abstract_name as default one, this would fail, correct?

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -718,7 +718,8 @@ <h4 class="panel-title pull-left pointer-cursor">
                     var scope = this;
                     var provider = capacitySetting.currentProvider;
                     value = 'cmp_base-ebs';
-                    if (this.selectedHostTypeValue.startsWith('EbsCompute')) {
+                    const selectedHostTypeObject = capacityCreationInfo.hostTypes.find(hostType => hostType.id === this.selectedHostTypeValue);
+                    if (selectedHostTypeObject.abstract_name.startsWith('EbsCompute')) {
```
ankilosaurus, line 722: This existing check itself does not look good to me. If user does not create host type with EbsCompute prefix, this would not work.
Is there any other property we can use to check if host is ebs/non-ebs ?

deploy-board/deploy_board/templates/configs/new_capacity.html
disparter: How can we get the ID for the defaultHostType?
this option is hardcoded in environment, I don't feel comfortable in putting a default ID on that file.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
disparter: Thank you for noticing that. I'm fixing this.

deploy-board/deploy_board/templates/configs/new_capacity.html
ankilosaurus: I am not suggesting to use ID. we can use default provider_name when provider is "AWS".

deploy-board/deploy_board/settings.py
```diff
@@ -282,8 +282,8 @@
     #Pinterest Default AMI image name
     DEFAULT_CMP_IMAGE = 'cmp_base-ebs-18.04'
 
-    #Pinterest Default Host Type
-    DEFAULT_CMP_HOST_TYPE = 'EbsComputeLo(Recommended)'
+    #Pinterest Default Host Provider Name
+    DEFAULT_PROVIDER_NAME = 'c5.2xlarge'
```
ankilosaurus, line 286: revert this.

ankilosaurus (approved):

PR #835, opened 2021-05-12
Add is_sox to environ, update readme for schema update versions.

robbintt:
Add `is_sox` column to environs.  This describes whether that env/stage is under sox.

### DIFF
--- a/tools/mysql/README.dev
+++ b/tools/mysql/README.dev
@@ -3,5 +3,3 @@ Follow the steps below to update deploy DB schema ( assume current DB schema ver
 1. Create file schema-update-11.sql, and put the DB schema change statements there
 2. Make sure the last statement reads:
    UPDATE schema_versions SET version=11;
-3. Make sure you update the last statement in ../../deploy-service/common/src/main/resources/sql/deploy.sql as well:
-   INSERT INTO schema_versions (version) VALUES (11);
--- a/tools/mysql/schema-update-14.sql
+++ b/tools/mysql/schema-update-14.sql
@@ -6,5 +6,4 @@
 
 ALTER TABLE environs ADD COLUMN is_sox TINYINT(1) NOT NULL DEFAULT 0;
 
--- make sure to update the schema version to 5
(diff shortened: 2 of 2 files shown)


tools/mysql/README.dev
```diff
@@ -3,5 +3,3 @@ Follow the steps below to update deploy DB schema ( assume current DB schema ver
 1. Create file schema-update-11.sql, and put the DB schema change statements there
 2. Make sure the last statement reads:
    UPDATE schema_versions SET version=11;
-3. Make sure you update the last statement in ../../deploy-service/common/src/main/resources/sql/deploy.sql as well:
```
robbintt, line 6: I looked at this file, it seems that we are not using it. We should remove the line from the file as well or should also update it there.

robbintt: looks like schema version is no longer updated per README.dev, i can remove the line or we can start updating it:

References:
https://github.com/pinterest/teletraan/pull/831/files?file-filters%5B%5D=.sql
https://github.com/pinterest/teletraan/pull/835/files#r631319018
https://github.com/pinterest/teletraan/blob/master/deploy-service/common/src/main/resources/sql/deploy.sql#L312

tools/mysql/schema-update-14.sql
```diff
@@ -0,0 +1,10 @@
+-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+-- ALWAYS BACKUP YOUR DATA BEFORE EXECUTING THIS SCRIPT
+-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+# This script upgrade DB schema from version 4 to version 5
+
+ALTER TABLE environs ADD COLUMN is_sox TINYINT(1) NOT NULL DEFAULT 0;
```
ankilosaurus, line 7: column name can be simply "sox" ?

ankilosaurus (approved):

PR #838, opened 2021-05-14
[DRAFT] partial impl of sox environ check

robbintt:


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -235,6 +236,13 @@ public Response create(
             }
         }
 
+        // only allow deploy of sox if not private and from sox_compliant source
```
ankilosaurus, line 239: can be moved to helper method.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -227,6 +227,7 @@ public Response create(
                     buildBean.getArtifact_url()));
         }
         // if the stage is not allowed (allow_private_build)
+        // TODO: remove this code once is_sox db column fully populated (before merging this)
```
ankilosaurus, line 230: Should have followup jira for tracking this?

deploy-service/common/src/main/java/com/pinterest/deployservice/allowlists/BuildAllowlistImpl.java
```diff
@@ -35,4 +37,14 @@ public Boolean trusted(String buildName) {
         }
         return false;
     }
+
+    // sox_compliant checks if build matches trusted URL allow list
+    public Boolean sox_compliant(String buildName) {
+        for (String pattern : soxBuildURLs) {
+            if (buildName.matches(pattern)) {
```
ankilosaurus, line 44: can buildNamebe null? same as other places though.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
robbintt: Good idea, done.

deploy-service/common/src/main/java/com/pinterest/deployservice/allowlists/BuildAllowlistImpl.java
robbintt: It's not nullable in the db schema, thanks for checking

robbintt:
merged 2 PRs that shared getters & setters

PR #839, opened 2021-05-15
Healthcheck endpoint

ankilosaurus:
Status: Testing

Problem: Default health check is done on admin port. 
```
curl http://localhost:80/healthcheck
{"deadlocks":{"healthy":true},"generic":{"healthy":true,"message":"Teletraan looks healthy!"}}
```
but

```
curl http://localhost:8080/healthcheck
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
<title>Error 404 Not Found</title>
</head>
<body><h2>HTTP ERROR 404</h2>
<p>Problem accessing /healthcheck. Reason:
<pre>    Not Found</pre></p><hr><i><small>Powered by Jetty://</small></i><hr/>

</body>
</html>
```

Fix: Provide Healthcheck endpoint on service port to be used by ELB/envoy.

Verify HTTP enpoint /healthcheck works on

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/TokenAuthFilter.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/TokenAuthFilter.java
@@ -32,6 +32,7 @@ import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.ext.Provider;
+import javax.ws.rs.core.UriInfo;
 
 @Provider
 @Priority(Priorities.AUTHENTICATION)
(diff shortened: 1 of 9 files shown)


disparter (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/TokenAuthFilter.java
```diff
@@ -66,7 +66,8 @@ public void filter(ContainerRequestContext context) throws IOException {
     private SecurityContext authenticate(ContainerRequestContext context) throws Exception {
         String authCredentials = context.getHeaderString(AUTHENTICATION_HEADER);
         if (StringUtils.isEmpty(authCredentials)) {
-            throw new DeployInternalException("Can not find HTTP header: Authorization!");
+            return userDataHelper.getUserSecurityContext(null);
+            //throw new DeployInternalException("Can not find HTTP header: Authorization!");
```
ankilosaurus, line 70: just for testing. I don't think we should commit this file. It basically bypasses authentication. 
Authorization could fail though.

disparter (approved):

yongwen:
Can we use the existing  :80 health check end point instead of the service port? They  are in the same process. And a separate thread for health check might be an advantage than sharing the service listener

ankilosaurus:
> Can we use the existing :80 health check end point instead of the service port? They are in the same process. And a separate thread for health check might be an advantage than sharing the service listener

hi @yongwen I am going to merge the PR as this is to provide healthcheck endpoint on service port. we can decide on usage of this vs keep using admin port in a separate thread with envoy.

PR #841, opened 2021-05-18
Fix issue in migration script - refers to a table that doesn't exist

mignaulo:
Hello Pinterest, 
`tools/mysql/schema-update-8.sql` seems to refer to a table that does not exist, `groups`. 

logs from the migration script:
```
Will upgrade schema from version 6 to version 13
upgrading to version 7 with schema-update-7.sql...
upgrading to version 8 with schema-update-8.sql...
ERROR 1146 (42S02) at line 8: Table 'deploy.groups' doesn't exist
```
My assumption is that the table that was meant to be modified is actually `groups_and_roles`. 
If `groups` should exist, it's not being tracked in any of the existing SQL files.

### DIFF
--- a/tools/mysql/schema-update-8.sql
+++ b/tools/mysql/schema-update-8.sql
@@ -5,7 +5,7 @@
 # This script upgrade DB schema from version 7 to version 8
 
 ALTER TABLE environs ADD COLUMN stage_type VARCHAR(32) NOT NULL DEFAULT 'PRODUCTION';
-ALTER TABLE groups MODIFY group_name VARCHAR(128) NOT NULL;
+ALTER TABLE groups_and_roles MODIFY group_name VARCHAR(128) NOT NULL;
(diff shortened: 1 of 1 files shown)


tools/mysql/schema-update-8.sql
```diff
@@ -5,7 +5,7 @@
 # This script upgrade DB schema from version 7 to version 8
 
 ALTER TABLE environs ADD COLUMN stage_type VARCHAR(32) NOT NULL DEFAULT 'PRODUCTION';
-ALTER TABLE groups MODIFY group_name VARCHAR(128) NOT NULL;
+ALTER TABLE groups_and_roles MODIFY group_name VARCHAR(128) NOT NULL;
```
ankilosaurus, line 8: groups_and_roles contains ldap group to role mapping and so does not require this ALTER query.

you are correct about non-exisitng groups table though.

mignaulo: thanks @ankilosaurus. Should I modify this PR to just take out the `groups` alter?

PR #843, opened 2021-05-19
[DRAFT] Restrict modifications of `is_sox` to SYSTEM Role

robbintt:
The new `is_sox` column will be used to enforce SOX builds cannot have private builds. However, we need to prevent environment users from changing this status with the normal PUT endpoint.

We should restrict the ability to change `is_sox` to authorized users. For now, we can restrict it to SYSTEM users.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -544,11 +544,11 @@ public class EnvironBean implements Updatable, Serializable {
         this.stage_type = stage_type;
     }
 
-    public Boolean getSOX() {
+    public Boolean getIs_sox() {
         return is_sox;
     }
 
-    public void setSOX(Boolean is_sox) {
+    public void setIs_sox(Boolean is_sox) {
         this.is_sox = is_sox;
     }
 
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -67,6 +67,7 @@ import javax.validation.constraints.Pattern;
  * allow_private_build TINYINT(1)    DEFAULT 0,
  * ensure_trusted_build TINYINT(1)    DEFAULT 0,
  * stage_type VARCHAR(32) NOT NULL DEFAULT PRODUCTION,
+ * is_sox TINYINT(1)    DEFAULT 0,
  * <p>
  * PRIMARY KEY   (env_id)
  * );
@@ -194,6 +195,9 @@ public class EnvironBean implements Updatable, Serializable {
 
     @JsonProperty("stageType")
     private EnvType stage_type;
+
+    @JsonProperty("isSOX")
+    private Boolean is_sox;
 
     public void validate() throws Exception {
         // A bunch of these fields will always be alphanumeric (with _ and -)
@@ -540,6 +544,14 @@ public class EnvironBean implements Updatable, Serializable {
         this.stage_type = stage_type;
     }
 
+    public Boolean getSOX() {
+        return is_sox;
+    }
+
+    public void setSOX(Boolean is_sox) {
+        this.is_sox = is_sox;
+    }
+
     @Override
     public SetClause genSetClause() {
         SetClause clause = new SetClause();
@@ -583,6 +595,7 @@ public class EnvironBean implements Updatable, Serializable {
         clause.addColumn("allow_private_build", allow_private_build);
         clause.addColumn("ensure_trusted_build", ensure_trusted_build);
         clause.addColumn("stage_type", stage_type);
+        clause.addColumn("is_sox", is_sox);
         return clause;
     }
 
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/Role.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/Role.java
@@ -30,13 +30,16 @@ package com.pinterest.deployservice.bean;
  *      plus the ability specify new OPERATORS and ADMINs for said environment.
  *      When a new environment is created the creating user is the designated the
  *      first ADMIN.
+ * SYSTEMADMIN:
+ *      System level administrator. This role is for system-wide use.
  */
 public enum Role {
     READER(0),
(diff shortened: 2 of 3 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/bean/Role.java
```diff
@@ -30,13 +30,16 @@
...
+ *      System level administrator. This role is for system-wide use.
  */
 public enum Role {
     READER(0),
     PINGER(1),
     PUBLISHER(1),
     OPERATOR(10),
-    ADMIN(20);
+    ADMIN(20),
+    SYSTEMADMIN(100);
```
robbintt, line 42: may not be used if we can check SYSTEM

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
```diff
@@ -103,6 +103,17 @@ public void update(@Context SecurityContext sc,
         if (origBean.getStage_type() != EnvType.DEFAULT && origBean.getStage_type() != environBean.getStage_type()) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, "Modification of stage type is not allowed!");
         }
+        // TODO: isUserInRole takes a string and does not seem implemented... it should take the value from Role enum
+        if(environBean.getSOX() && !sc.isUserInRole("SYSTEMADMIN")) {  // TODO: see note above
```
robbintt, line 107: This isn't working, need to get a better idea of how to check in this situation.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -540,6 +544,14 @@ public void setStage_type(EnvType stage_type) {
         this.stage_type = stage_type;
     }
 
+    public Boolean getSOX() {
+        return is_sox;
+    }
+
+    public void setSOX(Boolean is_sox) {
```
ankilosaurus, line 551: I think getters/setters needs to match property name.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
ankilosaurus: how about call authorizer with required role "SYSTEMADMIN" in case request contains SOX field change.

ankilosaurus: Calling authorizer with required role "SYSTEMADMIN" looks clean if possible.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
robbintt: Which part of the getter/setter? I'm not seeing what is not matching.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
```diff
@@ -103,6 +103,10 @@ public void update(@Context SecurityContext sc,
         if (origBean.getStage_type() != EnvType.DEFAULT && origBean.getStage_type() != environBean.getStage_type()) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, "Modification of stage type is not allowed!");
         }
+        if(environBean.getSOX()) {
+            // TODO: future - should a more granular resource be authorized on?
+            authorizer.authorize(sc, new Resource(environBean.getEnv_name(), Resource.Type.ENV), Role.SYSTEMADMIN);
```
robbintt, line 108: Not sure about the authorization, seems like we are authorizing on too broad an environment. Might be worth a closer look in the future.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
ankilosaurus: If property name is is_sox then getIs_sox()/setIs_sox().

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
ankilosaurus: Are we going to reject PUT API call for all SOX env in general? This will break clients. 
We should reject only if  non SYSADMIN user is trying to modify SOX property.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/Role.java
ankilosaurus: Will this role be available in UI to select?

robbintt: Can you share some more info about your question? I can add additional code to support adding a role, if needed.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
robbintt: Good point, I will add a condition to check if is_sox is different than on origBean.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/Role.java
robbintt: I will check for: API, UI.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
robbintt: 2 to check:
POST /v1/envs/{envName}/user_roles
POST /v1/envs/{envName}/group_role

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/Role.java
robbintt: 2 to check:
POST /v1/envs/{envName}/user_roles
POST /v1/envs/{envName}/group_role

robbintt:
merged 2 drafts that shared getters and setters

PR #846, opened 2021-05-20
Hosts load UI

disparter:
Summary:
The env landing page shows all the hosts and some stages are very heavy loaded and this can lead to 500 status code when Teletraan is too busy.

This PR will act in both frontend and backand, changing env landing to not load up thousand of hosts.

Also  the hosts page will now be used to get the details on the hosts instead of the env landing, but also we want to paginate this and reorder in a way that this page lists the most wanted hosts first so users can inspect as easily as before but without overloading the system.

[This PR is not fully completed, please do not merge it]

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvAgents.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvAgents.java
@@ -73,6 +73,19 @@ public class EnvAgents {
     }
 
     @GET
+    @ApiOperation(
+            value = "Get deploy agents paginated",
+            notes = "Returns a list of all the deploy agent objects for a given environment name and stage by page",
+            response = AgentBean.class, responseContainer = "List")
+    public List<AgentBean> getAllAgentsPaginated(
+            @ApiParam(value = "Environment name", required = true)@PathParam("envName") String envName,
+            @ApiParam(value = "Stage name", required = true)@PathParam("stageName") String stageName,
+            @ApiParam(value = "Page") Integer page, @ApiParam(value = "Size") Integer size) throws Exception {
+        EnvironBean envBean = Utils.getEnvStage(environDAO, envName, stageName);
+        return agentDAO.getAllByEnvPaginated(envBean.getEnv_id(), page, size);
+    }
+
+    @GET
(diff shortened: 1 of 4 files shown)


ankilosaurus (commented):
@disparter thanks for making this change. 
Few suggestions:
1. List down the tests you are going to perform.
2. Attach snapshot of the UI with this change. (landing page & host listing page)
3.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -1161,7 +1161,11 @@ def disable_all_env_change(request):
 def get_hosts(request, name, stage):
     envs = environs_helper.get_all_env_stages(request, name)
     stages, env = common.get_all_stages(envs, stage)
-    agents = agents_helper.get_agents(request, env['envName'], env['stageName'])
+
+    index = int(request.GET.get('page', '1'))
+    size = int(request.GET.get('size', DEFAULT_PAGE_SIZE))
```
ankilosaurus, line 1166: where is DEFAULT_PAGE_SIZE defined?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvAgents.java
```diff
@@ -72,6 +72,20 @@ public EnvAgents(TeletraanServiceContext context) {
...
+    @GET
+    @Path("/paginated")
+    @ApiOperation(
+            value = "Get deploy agents paginated",
+            notes = "Returns a list of all the deploy agent objects for a given environment name and stage by page",
+            response = AgentBean.class, responseContainer = "List")
+    public List<AgentBean> getAllAgentsPaginated(
+            @ApiParam(value = "Environment name", required = true)@PathParam("envName") String envName,
+            @ApiParam(value = "Stage name", required = true)@PathParam("stageName") String stageName,
+            @QueryParam(value = "Page") Integer page, @QueryParam(value = "Size") Integer size) throws Exception {
```
ankilosaurus, line 84: Page size is optional. So what will be the behavior when not passed?

ankilosaurus: Either set default size in API or make param required.

PR #849, opened 2021-06-03
enforce build validation for both deploy and promote

yongwen:
enforce build validation for both deploy and promote.
1. move the build validation logic into a function module
2. call build_validation for both new deploy and promote


robbintt (approved):
Looks like a better place, thank you!

PR #851, opened 2021-06-07
Optimize agent upsert query operations

ankilosaurus:
Problem
---
Entries in agents table gets updated on every ping even when there is no change. This insert/update happens in ping flow from each host Teletraan manages and is a high volume query specifically during fleet wide deploy. 

Fix
---
Don't update agents table if there is no update. This should reduce database upsert operations into agents table by 100% when there is no deploy(no state change). Note, Last_update field will not be updated on every ping.

Tests:
----
1. Deploy new build
2. Restart current build
3. Pause/resume a deploy in stage
4. Pause/resume deploy on a given host
5. Rollback deploy

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
@@ -154,11 +154,13 @@ public class GoalAnalyst {
     }
 
     public class UninstallCandidate implements Comparable<UninstallCandidate> {
+        AgentBean updateBean;
         PingReportBean report;
         EnvironBean environ;
 
-        UninstallCandidate(PingReportBean report) {
+        UninstallCandidate(AgentBean updateBean, PingReportBean report) {
             this.report = report;
+            this.updateBean = updateBean;
             environ = existingAgentEnv.getOrDefault(report.getEnvId(), new EnvironBean());
         }
 
@@ -273,8 +275,29 @@ public class GoalAnalyst {
         }
     }
 
+    boolean shouldUpdateAgentRecord(AgentBean origBean, AgentBean updateBean) {
+        
+        if (origBean == null || updateBean == null) {
+            return true;
+        }
+        if (origBean.getHost_id() != null && origBean.getHost_id().equals(host_id) && 
+            origBean.getDeploy_id() != null && origBean.getDeploy_id().equals(updateBean.getDeploy_id()) &&
+            origBean.getEnv_id() != null && origBean.getEnv_id().equals(updateBean.getEnv_id()) && 
+            origBean.getFail_count() != null && origBean.getFail_count().equals(updateBean.getFail_count()) &&
+            origBean.getStatus() != null && origBean.getStatus().equals(updateBean.getStatus()) && 
+            origBean.getLast_err_no() != null && origBean.getLast_err_no().equals(updateBean.getLast_err_no()) &&
+            origBean.getState() != null && origBean.getState().equals(updateBean.getState()) && 
+            origBean.getDeploy_stage() != null && origBean.getDeploy_stage().equals(updateBean.getDeploy_stage())) {
+            LOG.debug("Skip updating agent record for env_id {}, deploy_id {} on host {}",
+                    origBean.getEnv_id(), origBean.getDeploy_id(), origBean.getHost_id());
+            return false;
+        }
+        LOG.info("Agent record for env_id {}, deploy_id {} on host {} needs update",
+                origBean.getEnv_id(), origBean.getDeploy_id(), origBean.getHost_id());
+        return true;
+    }
+
     // Generate new agent bean based on the report & current agent record,
-    // This is intended to be used for update agent record, not for goal,
     // We populate all the fields, since this could be used for insertOrUpdate as well
     AgentBean genUpdateBeanByReport(PingReportBean report, AgentBean agent) {
         // We generate complete bean in case we need to insertOrUpdate it into agents table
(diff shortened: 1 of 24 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -277,6 +277,25 @@ AgentState proposeNewAgentState(PingReportBean report, AgentBean agent) {
     // This is intended to be used for update agent record, not for goal,
     // We populate all the fields, since this could be used for insertOrUpdate as well
     AgentBean genUpdateBeanByReport(PingReportBean report, AgentBean agent) {
+        AgentState newAgentState = proposeNewAgentState(report, agent);
+
+        // Don't update if not changed.
+        if (agent != null && agent.getHost_id() == host_id && agent.getDeploy_id() == report.getDeployId() &&
```
disparter, line 283: Is this checks are fine? I mean, they are Strings, why not use equals instead?

disparter (changes requested):
The building is failing also, after dealing with comment I think its good to go

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
ankilosaurus: yup, this won't work. Working on changing it.

ankilosaurus:
> The building is failing also, after dealing with comment I think its good to go

yeah I was expecting test failures and was surprised when it didn't fail earlier. Now it's failing on tests which is good.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -314,7 +333,7 @@ AgentBean genUpdateBeanByReport(PingReportBean report, AgentBean agent) {
     // Generate new agent bean based on the report & current agent record,
     // This is intended to be used for deploy goal to install next stage
     AgentBean genNextStageUpdateBean(EnvironBean env, PingReportBean report, AgentBean agent) {
-        AgentBean agentBean = genUpdateBeanByReport(report, agent);
+        AgentBean agentBean = genAgentBeanByReport(report, agent);
```
yongwen, line 336: can we still call genUpdateBeanByReport() to minimize/isolate the changes. as the logic is to done mainly  inside the genUpdateBeanByReport()

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -461,7 +480,7 @@ void process(String envId, EnvironBean env, PingReportBean report, AgentBean age
         AgentBean updateBean = null;
         if (report != null) {
             updateBean = genUpdateBeanByReport(report, agent);
-            if (!StringUtils.isEmpty(report.getEnvId())) {
+            if (!StringUtils.isEmpty(report.getEnvId()) && updateBean != null) {
```
yongwen, line 483: we are changing the logic here a lot, wondering if we can minimize the logic change and only handle not to update the DB part.  The unit test is failing so worth to try to isolate the change.

ankilosaurus: yes, agree. I will upload new diff.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
ankilosaurus: updating a diff.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -273,6 +273,28 @@ AgentState proposeNewAgentState(PingReportBean report, AgentBean agent) {
...
+                    origBean.getEnv_id(), origBean.getDeploy_id(), origBean.getHost_id());
+            return false;
+        }
+        LOG.info("Agent record for env_id {}, deploy_id {} on host {} needs update",
+                origBean.getEnv_id(), origBean.getDeploy_id(), origBean.getHost_id());
+        return true;
+    }
+
     // Generate new agent bean based on the report & current agent record,
     // This is intended to be used for update agent record, not for goal,
```
robbintt, line 299: update comment

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -646,6 +671,7 @@ void process(String envId, EnvironBean env, PingReportBean report, AgentBean age
             LOG.debug("GoalAnalyst case 3.1 - add an uninstall candidate to instruct host {} to remove the retired env {}", host, report.getEnvId());
             uninstallCandidates.add(new UninstallCandidate(report));
             updateBean.setState(AgentState.DELETE);
+            needUpdateAgents.put(report.getEnvId(), updateBean);
```
robbintt, line 674: would be better if this happened implicitly when updateBean.setState()

robbintt: discussed: move to caller

robbintt (changes requested):
see inline comments

yongwen (approved):

robbintt (approved):

PR #853, opened 2021-06-10
display the ongoing sidecar deploys and total running deploy count

yongwen:
1. add a new api: /envs/sidecars  to return all sidecar envs that is identified as having system_priority
2. display the ongoing sidecar deploy in the home page in a table
3. display the total running deploy count in the "deploy statistic" section on the homepage
Screen Shot 2021-06-10 at 12 49 25 PM

### DIFF
--- a/deploy-board/deploy_board/templates/landing.html
+++ b/deploy-board/deploy_board/templates/landing.html
@@ -18,19 +18,29 @@
     </div>
 </div>
 
-    <div class="panel panel-default">
-    {% include "panel_heading.tmpl" with panel_title="Deployment Statistics" panel_body_id="deployStatsId" direction="down" %}
-        <div id="deployStatsId" class="collapse in panel-body">
-            <div id="dailyDeployCountId"></div>
-        </div>
+<div class="panel panel-default">
+    {% include "panel_heading.tmpl" with panel_title="Ongoing Sidecar Deployments" panel_body_id="onGoingSidecarDeployId" direction="down" %}
+    <div id="onGoingSidecarDeployId" class="collapse in panel-body table-responsive">
     </div>
+</div>
+
+<div class="panel panel-default">
+{% include "panel_heading.tmpl" with panel_title="Deployment Statistics" panel_body_id="deployStatsId" direction="down" %}
+    <div id="deployStatsId" class="collapse in panel-body">
+        <div id="dailyDeployCountId"></div>
+    </div>
+</div>
 <script>
     $(function () {
         $('#onGoingDeployId').load('/deploys/ongoing/');
-
         setInterval(function() {
             $('#onGoingDeployId').load('/deploys/ongoing/');
-        }, 30000);
+        }, 60000);
+
+        $('#onGoingSidecarDeployId').load('/deploys/ongoing_sidecar/');
+        setInterval(function() {
+            $('#onGoingSidecarDeployId').load('/deploys/ongoing_sidecar/');
+        }, 60000);
 
         $('#dailyDeployCountId').load('/deploys/dailycount');
     });
--- a/deploy-board/deploy_board/webapp/deploy_views.py
+++ b/deploy-board/deploy_board/webapp/deploy_views.py
@@ -28,6 +28,36 @@ from helpers import builds_helper, deploys_helper, environs_helper, tags_helper
 DEFAULT_PAGE_SIZE = 30
 DEFAULT_ONGOING_DEPLOY_SIZE = 10
 
+def _get_running_deploys_count(request):
+    deploy_states = ["RUNNING"]  # current running deploys
+    page_size = 1                # only need to return 1 detail, the return structure has the "total" value
+    deployResult = deploys_helper.get_all(request, deployState=deploy_states, pageSize=page_size)
+    return deployResult['total']
+
+def _get_sidecars(request):
+    # returns a list of env id for sidecars which are identified by having a system priority
+    envs = environs_helper.get_all_sidecar_envs(request)
+    env_ids = []
+    for env in envs:
+        env_ids.append(env['id'])
+    return env_ids
+
+def _get_ongoing_sidecar_deploys(request):
+    deploy_summaries = []
+    env_ids = _get_sidecars(request)
+    if env_ids:
+        deploy_states = ["RUNNING", "FAILING"]
+        deployResult = deploys_helper.get_all(request, envId=env_ids, deployState=deploy_states)
+        for deploy in deployResult['deploys']:
+            env = environs_helper.get(request, deploy['envId'])
+            build = builds_helper.get_build(request, deploy['buildId'])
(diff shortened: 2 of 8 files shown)


deploy-board/deploy_board/templates/landing.html
```diff
@@ -18,19 +18,29 @@
     </div>
 </div>
 
-    <div class="panel panel-default">
-    {% include "panel_heading.tmpl" with panel_title="Deployment Statistics" panel_body_id="deployStatsId" direction="down" %}
-        <div id="deployStatsId" class="collapse in panel-body">
-            <div id="dailyDeployCountId"></div>
-        </div>
+<div class="panel panel-default">
```
ankilosaurus, line 21: can we also show count of ONGOING sidecars and all deploys?

yongwen: there is limit number of sidecar so the the ongoing sidecar count should be very small so the table should be sufficient.  since it is the home page, also want to limit the DB/api call to ensure the fast load of the home page.

ankilosaurus: Sure, It probably does not matter much for sidecar today but for ongoing deploy number in general for all services might be helpful. Not sure if we need additional db query for that, since we already have a list of ongoing deploys retrieved.

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBEnvironDAOImpl.java
```diff
@@ -96,6 +96,8 @@
         "SELECT env_id FROM environs";
     private static final String GET_ALL_ENVS =
         "SELECT * FROM environs";
+    private static final String GET_ALL_SIDECAR_ENVS =
+        "SELECT * FROM environs where system_priority > 0";
```
robbintt, line 100: Does new DAO still require unit tests?

deploy-board/deploy_board/webapp/deploy_views.py
```diff
@@ -28,6 +28,36 @@
...
+    return env_ids
+
+def _get_ongoing_sidecar_deploys(request):
+    deploy_summaries = []
+    env_ids = _get_sidecars(request)
+    if env_ids:
+        deploy_states = ["RUNNING", "FAILING"]
+        deployResult = deploys_helper.get_all(request, envId=env_ids, deployState=deploy_states)
+        for deploy in deployResult['deploys']:
+            env = environs_helper.get(request, deploy['envId'])
```
robbintt, line 52: nit: assign directly to dict key in summary

robbintt: same for build

deploy-board/deploy_board/webapp/deploy_views.py
```diff
@@ -28,6 +28,36 @@
...
+    deploy_states = ["RUNNING"]  # current running deploys
+    page_size = 1                # only need to return 1 detail, the return structure has the "total" value
+    deployResult = deploys_helper.get_all(request, deployState=deploy_states, pageSize=page_size)
+    return deployResult['total']
+
+def _get_sidecars(request):
+    # returns a list of env id for sidecars which are identified by having a system priority
+    envs = environs_helper.get_all_sidecar_envs(request)
+    env_ids = []
+    for env in envs:
```
robbintt, line 41: for easier readability: list comprehension candidate

env_ids = [env['id'] for env in envs]

deploy-board/deploy_board/webapp/deploy_views.py
```diff
@@ -57,6 +87,16 @@ def get_landing_page(request):
         "redirect_create_env_page_url": TELETRAAN_REDIRECT_CREATE_ENV_PAGE_URL
     })
 
+def get_ongoing_sidecar_deploys(request):
+    deploy_summeries = _get_ongoing_sidecar_deploys(request)
```
robbintt, line 91: typo: "summeries"

deploy-board/deploy_board/webapp/deploy_views.py
```diff
@@ -28,6 +28,36 @@
 DEFAULT_PAGE_SIZE = 30
 DEFAULT_ONGOING_DEPLOY_SIZE = 10
 
+def _get_running_deploys_count(request):
+    deploy_states = ["RUNNING"]  # current running deploys
+    page_size = 1                # only need to return 1 detail, the return structure has the "total" value
+    deployResult = deploys_helper.get_all(request, deployState=deploy_states, pageSize=page_size)
+    return deployResult['total']
+
+def _get_sidecars(request):
```
ankilosaurus, line 37: just a check here, "sidecar" is the term we want to use here?

robbintt (approved):
Added improvement ideas, also not sure if a unit test is needed for DAO change.

deploy-board/deploy_board/templates/landing.html
ankilosaurus: nm, I see ongoing deploys is a paginated query for good reason. So, yeah, I think we can avoid separate query for pulling count for now.

deploy-board/deploy_board/webapp/deploy_views.py
```diff
@@ -28,6 +28,36 @@
...
+    for env in envs:
+        env_ids.append(env['id'])
+    return env_ids
+
+def _get_ongoing_sidecar_deploys(request):
+    deploy_summaries = []
+    env_ids = _get_sidecars(request)
+    if env_ids:
+        deploy_states = ["RUNNING", "FAILING"]
+        deployResult = deploys_helper.get_all(request, envId=env_ids, deployState=deploy_states)
```
ankilosaurus, line 50: Although it does not matter for sidecar today, Can we not load list of all services or sidecars on on homepage? My suggestion is load on demand (explicitly clicked by user)

ankilosaurus: worried about get_all kind of calls slowing UI especially when it's not paginated.

deploy-board/deploy_board/webapp/deploy_views.py
yongwen: yes. I think it is better to have this concept as we've been using it everywhere

deploy-board/deploy_board/webapp/deploy_views.py
yongwen: it is a filtered call for limited sidecar envs. I also reduced the refreshed interval from 30sec to 60sec. if it turn out the queries are too expensive (with growing deploys tables), we can: 1. move old deploy history to archived table. 2. move the deploy stats into a separate page instead of home page.  as we are re-inventing the Teletraan UI, will delay any extensive UI changes

ankilosaurus: sure, np.

PR #855, opened 2021-06-16
Spiffe_id validation bug

ankilosaurus:
Status: testing

Problem:
---
Spiffe_id validation bug in group config UI is preventing users from making any changes.

Fix:
---
Fixed bug in Spiffe_id validation code. 
Note: Ideally this validation should be done in backend (API) as user can still bypass this check by using rest api. I will create separate rodimus PR for that.

Test:
---
1. modification of group config is allowed. 
2. CUD of spiffe_id in group config is dis-allowed.

### DIFF
--- a/deploy-board/deploy_board/webapp/group_view.py
+++ b/deploy-board/deploy_board/webapp/group_view.py
@@ -164,6 +164,12 @@ def get_pas_config(request, group_name):
         raise
     return HttpResponse(json.dumps(html), content_type="application/json")
 
+def get_spiffe_id(user_data):
+    for entry in user_data:
+        tokens = entry.split(":", 1)
+        if tokens[0].strip() == "spiffe_id":
+            return tokens[1].strip()
+    return None
 
 def update_launch_config(request, group_name):
     try:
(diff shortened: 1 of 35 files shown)


deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -181,33 +181,28 @@ def update_launch_config(request, group_name):
...
-        userPassInSpiffeID = None
+        desired_spiffe_id = None
         passInUserDataList = params["userData"].splitlines()
         for item in passInUserDataList:
-                if ":" in item:
-                    pair = item.split(":")
-                    if pair[0] == "spiffe_id":
-                        userPassInSpiffeID = pair[1]
-                        break
+            desired = item.split(":", 1)
```
yongwen, line 187: maybe we should do a trim() to remove the whitespace around the keyword

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -181,33 +181,28 @@ def update_launch_config(request, group_name):
...
-                        raise TeletraanException("Teletraan does not support user to remove spiffe_id")
-
-                    if pair[0] == "spiffe_id" and pair[1] != userPassInSpiffeID:
-                        log.error("Teletraan does not support user to update spiffe_id in userData field")
-                        raise TeletraanException("Teletraan does not support user to update spiffe_id")
-
-                    if pair[0] != "spiffe_id" and userPassInSpiffeID is not None:
-                        log.error("Teletraan does not support user to create spiffe_id in userData field")
-                        raise TeletraanException("Teletraan does not support user to create spiffe_id")
+                actual = item.split(":", 1)
```
yongwen, line 198: should we make this into a function like "getSpiffe_id" since it is being dup from previous code

yongwen (approved):

PR #857, opened 2021-06-16
Add same fix for default host type to advanced cluster creation view.

robbintt:
The new default host type value was missed for the advanced form in https://github.com/pinterest/teletraan/pull/833.

This PR adds the same value find method as the basic form in the advanced form.

### DIFF
--- a/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
@@ -183,7 +183,7 @@ var capacitySetting = new Vue({
                     isSelected: item.abstract_name === info.defaultSeurityZone
                 }
             }),
-        selectedHostTypeValue: info.defaultHostType,
+        selectedHostTypeValue: capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType).id,
         selectedSecurityZoneValue: info.defaultSeurityZone,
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -135,9 +135,8 @@ <h4 class="panel-title pull-left">Capacity</h4>
         hostTypeOptions: info.hostTypes.map(
             function (item, idx) {
                 return {
-                    value: item.id,
-                    text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",
-                    isSelected: item.abstract_name === info.defaultHostType
+                    value: item.id, text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",
+                    isSelected: item.abstract_name === capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType).id
```
ankilosaurus, line 139: comparing abstract_name with id ?

robbintt: Just checking if that was used as a side effect somewhere, not using that

ankilosaurus (approved):

PR #864, opened 2021-08-10
Scrub filtered query param values from response logs.

robbintt:
Goal: do not log queryparam values if the queryparam key has a case insensitive substring in the filter list `filteredQueryKeySubstrings`.

Summary:
- normalized internals of public interfaces
- added query parameter value filter function and filter list

Manual testing:
- call with scrubbed queries
- call without scrubbed queries

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
@@ -30,6 +30,7 @@ import java.util.Map;
 public class HTTPClient {
     private static final int TIMEOUT = 15*1000;  // http timeout in 15 seconds
     private static final Logger LOG = LoggerFactory.getLogger(HTTPClient.class);
+    private String[] filteredQueryKeySubstrings = {"token"};
 
     private String generateUrlAndQuery(String url, Map<String, String> params) throws Exception {
         if (params == null || params.isEmpty()) {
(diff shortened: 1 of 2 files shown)


ankilosaurus (commented):
url is logged in catch {} block as well.

deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -97,7 +97,8 @@ private String internalCall(String url, String method, String payload, Map<Strin
                     throw new DeployInternalException("HTTP request failed, status = {}, content = {}",
                         responseCode, ret);
                 }
-                LOG.info("HTTP Request returned with response code {} for URL {}", responseCode, url);
+                String urlNoQueryString = url.substring(0, url.lastIndexOf("?"));
```
ankilosaurus, line 100: Can we use URL object to get base url ?

deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -47,27 +48,65 @@ private String generateUrlAndQuery(String url, Map<String, String> params) throw
...
+        sb.append(url);
+        String prefix = "?";
+        String paramValue = "";
+        boolean isSensitiveValue = false;
+        for (Map.Entry<String, String> entry : params.entrySet()) {
+            isSensitiveValue = false;
+            paramValue = "";
+            sb.append(prefix);
+            prefix = "&";
+            for (String filteredQueryKeySubstring : filteredQueryKeySubstrings) {
```
ankilosaurus, line 66: 1. Put comment that this could be a perf. issue if list get too large or query gets too large
2. Only call query string scrubber for DEBUG level log ( INFO level prints only Base url, DEBUG prints Base + mask query params)

ankilosaurus (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -31,7 +31,7 @@
     private static final int TIMEOUT = 15*1000;  // http timeout in 15 seconds
     private static final Logger LOG = LoggerFactory.getLogger(HTTPClient.class);
 
-    private String generateUrlAndQuery(String url, Map<String, String> params) throws Exception {
+    public String generateUrlAndQuery(String url, Map<String, String> params, boolean scrubUrl) throws Exception {
```
robbintt, line 34: Temporarily converted to public method to discuss testing practices (see unit tests file).

deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -41,33 +41,51 @@ private String generateUrlAndQuery(String url, Map<String, String> params) throw
...
+            //        consider using it only in only a DEBUG logging context in the future
+            String reportedValue = scrubUrl ? scrubUrlQueryValue(entry.getKey(), entry.getValue()) : entry.getValue();
             sb.append(String.format("%s=%s", entry.getKey(),
-                URLEncoder.encode(entry.getValue(), "UTF-8")));
+                URLEncoder.encode(reportedValue, "UTF-8")));
         }
         return sb.toString();
     }
 
+    public String scrubUrlQueryValue(String queryParamKey, String queryParamValue) {
```
robbintt, line 53: Temporarily converted to public method to discuss testing practices (see unit tests file).

deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -41,33 +41,49 @@ private String generateUrlAndQuery(String url, Map<String, String> params) throw
         for (Map.Entry<String, String> entry : params.entrySet()) {
             sb.append(prefix);
             prefix = "&";
+            // note:  scrubUrlQueryValue could be expensive with many filtered values
+            //        consider using it only in only a DEBUG logging context in the future
+            String reportedValue = scrubUrl ? scrubUrlQueryValue(entry.getKey(), entry.getValue()) : entry.getValue();
```
ankilosaurus, line 46: since changing existing method, please verify
1. call with scrubbed queries
2. call without scrubbed queries

robbintt: I think the unit tests below cover this already, did you mean something else?

ankilosaurus: sure, UT is good verification. however, I was referring to Integ testing.

deploy-service/common/src/test/java/com/pinterest/deployservice/common/HTTPClientTest.java
```diff
@@ -0,0 +1,65 @@
...
+public class HTTPClientTest {;
+    @Test
+    public void testScrubUrlQueryValue() throws Exception {
+        HTTPClient httpClient = new HTTPClient();
+        // Future: there should be some better way to test HTTPClient private methods rather than use reflection on private methods
+        Method scrubUrlQueryValue = HTTPClient.class.getDeclaredMethod("scrubUrlQueryValue", String.class, String.class);
+        scrubUrlQueryValue.setAccessible(true);
+
+        assertEquals("xxxxxxxxx", scrubUrlQueryValue.invoke(httpClient, "access_token", "dangerous_stuff"));
+        assertEquals("xxxxxxxxx", scrubUrlQueryValue.invoke(httpClient, "token", "dangerous_stuff"));
```
ankilosaurus, line 36: Can we define constant ex: SECRET_MASK="xxxxxxxxx" in HTTPCLIENT and use here?

PR #865, opened 2021-08-10
Notify slack/email recipients of a env/stage on unreachable instances

ankilosaurus:
Status: Testing

Problem:
---
User is not notified when instance becomes unreachable. When this happens, instance could have either been dead or not able to talk to Teletraan server for various possible reasons. If instance is active, services would continue to run but, new deploy on instance would fail. 

Fix:
---
Alert slack and email recipients when this happens.

Tests:
---
1. Verify slack alert by makin instance unreachable either by removing deployd or by terminating instance using AWS console or API.
2. Verify slack alert by makin instance unreachable either by removing deployd or by terminating instance using AWS console or API.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
@@ -19,10 +19,15 @@ package com.pinterest.teletraan.worker;
 import com.pinterest.deployservice.ServiceContext;
 import com.pinterest.deployservice.bean.HostBean;
 import com.pinterest.deployservice.bean.HostAgentBean;
+import com.pinterest.deployservice.bean.EnvironBean;
 import com.pinterest.deployservice.bean.HostState;
+import com.pinterest.deployservice.dao.EnvironDAO;
 import com.pinterest.deployservice.group.HostGroupManager;
 import com.pinterest.deployservice.rodimus.RodimusManager;
 import com.pinterest.deployservice.handler.HostHandler;
+import com.pinterest.deployservice.handler.CommonHandler;
+import com.pinterest.deployservice.common.NotificationJob;
+import java.util.concurrent.ExecutorService;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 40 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
```diff
@@ -108,21 +121,40 @@ private void processHighWatermarkHosts() throws Exception {
...
+        notifyOnUnreachableHosts(unreachableCountPerASG);
+    }
+
+    private void notifyOnUnreachableHosts(Map<String, Integer> unreachableCountPerASG) throws Exception {
+        for (Map.Entry<String, Integer> entry : unreachableCountPerASG.entrySet()) {
+            EnvironBean environBean = entry.getValue() != null ? environDAO.getByCluster(entry.getKey()) : null;
+            if (environBean != null) {
+                String webLink = deployBoardUrlPrefix + String.format("/env/%s/%s", environBean.getEnv_name(), environBean.getStage_name());
+                String message = String.format("Instances not reporting (unreachable) to Teletraan.");
+                String subject = String.format("Cluster %s has %d unreachable instances.", webLink, entry.getValue());
```
yongwen, line 152: cluster is weblink?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
```diff
@@ -108,21 +121,40 @@ private void processHighWatermarkHosts() throws Exception {
...
             } else {
                 markUnreachableHost(staleId);
+                unreachableCountPerASG.put(entry.getValue(), unreachableCountPerASG.getOrDefault(entry.getValue(), 0) + 1);
             }
         }
+        notifyOnUnreachableHosts(unreachableCountPerASG);
+    }
+
+    private void notifyOnUnreachableHosts(Map<String, Integer> unreachableCountPerASG) throws Exception {
+        for (Map.Entry<String, Integer> entry : unreachableCountPerASG.entrySet()) {
```
yongwen, line 147: maybe add a protection on the number of notification can send so not to spam/overload email/slack server?  for example, if ASG map size > 20 (?), then only notifiy teletraan team.

yongwen (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AgentJanitor.java
ankilosaurus: After thinking a bit more, I think we probably don't want to slack notify on unreachable instances but rather export this as metric and have user setup alerts. Along with that, we can notify Teletraan on large number of unreachable instances.
I am thinking of closing this PR and instead work on metrics.

PR #866, opened 2021-08-17
Add unique parameter to sc_config_id. DANGER: Applying this will delete duplicated records.

robbintt:
Add unique parameter to environs.sc_config_id

DANGER: Applying this will delete duplicated records. Do not apply without evaluating existing sc_config_ids and migrating desired duplicates to unique IDs.

Ref: https://dev.mysql.com/doc/refman/5.7/en/alter-table.html#idm140368130931568


ankilosaurus:
Plz list tests you are planning to/have run in Integ.

ankilosaurus (approved):

ankilosaurus (changes requested):
Can you test how the error reflected in UI when constraint is violated?
We should populate error but not throw exception in UI.

robbintt:
> Can you test how the error reflected in UI when constraint is violated?
> We should populate error but not throw exception in UI.

There shouldn't be any way to violate this constraint in the UI. The only example of this bug is in the API, I will test that as well.

PR #869, opened 2021-08-24
Remove "update cluster form" call to vue function for update cluster name.  Remove inline references to base image.

robbintt:
1. Remove advanced call to vue function for update cluster name. This feature causes the user's visible fields to be altered and can cause issues.
2. Remove inline references to base image and get from settings.py. In the future this should come from some external source.
3. Fix outdated `cmp_base-ebs` reference in cluster configuration form (#2 on this list covers this fix)

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -59,7 +59,7 @@
                     </baseimage-select>
                     <base-image-help v-show="showBaseImageHelp" v-bind:data="baseImageHelpData"></base-image-help>
                     <label-select label="Host Type" title="Compute Capability of the host" v-model="selectedHostTypeValue" v-bind:selectoptions="hostTypeOptions"
-                        v-on:input="updateImageNameForEbsHost" showhelp="true" v-on:helpclick="hostTypeHelpClick"></label-select>
+                        showhelp="true" v-on:helpclick="hostTypeHelpClick"></label-select>
                     <hostype-help v-show="showHostTypeHelp" v-bind:data="hostTypeHelpData"></hostype-help>
                     <label-select label="Security Zone" title="Security zone to control inbound/outboud traffic" v-model="selectedSecurityZoneValue" v-bind:selectoptions="securityZones"
                     showhelp="true" v-on:helpclick="securityZoneHelpClick"></label-select>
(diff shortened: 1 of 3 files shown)


robbintt:
I think I removed the autoupdate function from the wrong one, doublecheck.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -372,7 +372,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
             var cell = capacitySetting.currentCell;
             var scope = this;
             var provider = capacitySetting.currentProvider;
-            value = 'cmp_base-ebs-18.04';
+            value = '{{ default_cmp_image }}';
```
ankilosaurus, line 375: can we just use black as default type and throw error form is submitted with black?

robbintt: It's used in the basic form creation. We should get rid of basic form in the future, but have to keep it default until those design considerations can be done.

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -717,7 +717,7 @@ <h4 class="panel-title pull-left pointer-cursor">
                     var cell = capacitySetting.currentCell;
                     var scope = this;
                     var provider = capacitySetting.currentProvider;
-                    value = 'cmp_base-ebs';
+                    value = '{{ default_cmp_image }}';
```
ankilosaurus, line 720: when instance_type is changed, image name(abstract_name) and image(provider_name or ami) both changes. Can  you make sure to test if image name and image are in sync. For ex: it should not be populating u18 image name and u14 based ami for selection.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
ankilosaurus: I think using " " as default should be fairly straightforward to do now but leave it up to you.

robbintt: It requires moving the empty field to the basic form. I think it's better in more of a design discussion rather than adding a new field to the basic form in this PR.

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
robbintt: I reviewed this functionality, it updates to the latest AMI for the correct instance name.  I don't think it's good to be updating this on the user, but it's will keep users up to date, so we can discuss as more of a feature revamp for this form in a follow up.

ankilosaurus (approved):

PR #872, opened 2021-09-14
Allow deploy to continue past STOPPING after running out of non-zero failures.

robbintt:
Implementation:

The new agent status is not in FATAL_AGENT_STATUSES, so the STOPPING code will no longer pause.

TODO:
- Review the [explicit stopping code](https://github.com/pinterest/teletraan/blame/master/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java#L241): the stopping behavior has a history and the original intent was to let stopping fall through, but it was changed.

Goal:

Preserve the ability of clusters to scale back up before all scale down hosts have terminated when a sidecar deploy has broken STOPPING on machines in the cluster. 

Background: 

During a deployment state change, if there is a lifecycle event then it will start the associated executable. If the executable fails there are a number of retries and a timeout. In the case of the STOPPING lifecycle event, for terminating a host, if the lifecycle timeout is too long or the executable failures take too long, the hosts will hang failed but still take up capacity in the cluster for a long time.

Many clusters rely on scale down to save cost and scale up to meet demand, it is the basic feature of autoscaling. However, if hosts in the STOPPING state take up capacity, then autoscaling will not be effective and hosts can fall below the needs of the service.


robbintt:
Probably going to close this PR and revert part of this change re-allowing STOPPING hosts to fall through: https://github.com/pinterest/teletraan/commit/a2dba4515a71aa7b862afb5ba4bdcfae2b7d3e2b

robbintt:
closed in favor of https://github.com/pinterest/teletraan/pull/873

PR #874, opened 2021-09-15
Slack app integration for notifications

ankilosaurus:
Slack webhook is deprecated in favor of slack app.

Fix:
---
1. use slack sdk to integrate with slack to send notifications.
2. Teletraan service config yaml will hold - key to read, reader and mail domain.

Test:
---
1. Deploy notifications are sent to user configured **public** slack channel.
2. Deploy notifications are sent to user configured **private** slack channel **when user has added app to channel**
2. Deploy notifications are sent to user. (Notifications to slack user does not work with code in master. with this PR, user will get notification from App)


deploy-service/common/src/main/java/com/pinterest/deployservice/chat/SlackChatManager.java
```diff
@@ -18,50 +18,96 @@
...
-                slack.sendToChannel(convertedRoom)
-                    .displayName(from)
-                    .push(new SlackMessage(message));
-                return;
+                ChatPostMessageResponse
+                    response = slack.methods(this.token).chatPostMessage(req -> req
+                    .channel(channel)
+                    .text(message)
+                );
+                System.out.println(response);
```
yongwen, line 51: Remove?

yongwen (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/chat/SlackChatManager.java
ankilosaurus: ah, good catch.

PR #875, opened 2021-09-16
Select machine by id, not abstract name.

robbintt:
Selected comparison is not correct which can cause invalid data to populate in the UI.

Also fixed 2 comparisons in "new capacity" and "new capacity advanced" although these comparisons should not cause any problem because they are not selecting existing data.  Potentially they will cause these fields to be unselected and require user interaction, but we haven't had a report of any submit issue on new cluster create.

Testing plan:
- reproduce issue DONE
- manual test this fix

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -459,7 +459,7 @@
                     return {
                         value: item.id,
                         text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",
-                        isSelected: item.abstract_name === currentCluster.hostType
+                        isSelected: item.id === currentCluster.hostType
                     }
                 }),
                 imageNameValue: currentCluster.baseImageName,
--- a/deploy-board/deploy_board/templates/configs/new_capacity.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity.html
@@ -119,7 +119,7 @@ var capacitySetting = new Vue({
             return {
                 value: item.id,
                 text: item.abstract_name+" ("+item.core+" cores, "+item.mem+" GB, "+item.storage+")",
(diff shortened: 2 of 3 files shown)


deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -459,7 +459,7 @@ <h4 class="panel-title pull-left pointer-cursor">
                     return {
                         value: item.id,
                         text: item.abstract_name + " (" + item.core + " cores, " + item.mem + " GB, " + item.storage + ", " + item.provider + ": " + item.provider_name + ")",
-                        isSelected: item.abstract_name === currentCluster.hostType
+                        isSelected: item.id === currentCluster.hostType
```
ankilosaurus, line 462: MIght want to check if we need to fix this at other places as well.

robbintt: I don't think so, I will test this since we know about this issue, hopefully there aren't too many other aliases for these in the codebase:
```💡 grep -rli "abstract_name.*host.*type" .
./get_host_type.tmpl
./cluster_configuration.html```

Comment (robbintt) [reply deploy-board/deploy_board/templates/clusters/cluster_configuration.html:462]: Hm, I was able to find some in cluster configuration in the template directory, will check those out:
```
💡 pwd
/home/trobbins/code/teletraan/deploy-board/deploy_board/templates/clusters
💡 grep -rli "abstract.*host.*type" ..
../clusters/get_host_type.tmpl
../clusters/cluster_configuration.html
../configs/new_capacity.html
../configs/new_capacity_adv.html
```

robbintt:
This fix isn't working, will have to retrace it.  I was able to reproduce the issue though.

robbintt:
Looks like this actually comes back to the find() function grabbing the first matching abstract name in this unresolved comment. The fix is to change this get method to rely on ID not abstract_name: https://github.com/pinterest/teletraan/pull/833#discussion_r630363364

PR #880, opened 2021-10-01
Allow graceful shutdown when deploy goal changes.

ankilosaurus:
Problem:
---
When deploy goal changes, deployd sets agent status code to ABORTED_BY_SERVER and in this case deployd sends  SIGKILL to in-progress application script and does not give a chance for application script to cleanup and gracefully terminate.

Fix:
---
When deploy goal changes (agent is ABORTED_BY_SERVER in deployd) send SIGTERM. It's possible that application script ignore or incorrectly handle SIGTERM. deployd has terminate case later in the case where it does send SIGKILL once timedout.

Note: not adding another sleep factor here and relying on existing termination logic based on timeout so that application which needs shorter timeout is forced to handle SIGTERM going forward.

Test:
---
Deploy service and while it's in progress redeploy and verify that SIGTERM is sent and not the SIGKILL


deploy-agent/deployd/common/executor.py
```diff
@@ -182,10 +182,21 @@ def _get_sleep_interval(self, start, interval):
...
             os.killpg(process.pid, signal.SIGKILL)
         except Exception as e:
             log.debug('Failed to kill process: {}'.format(e))
 
+    @staticmethod
+    def _graceful_shutdown(process):
+        try:
+            log.info("Gracefully shutdown currently running process")
+            os.killpg(process.pid, signal.SIGTERM)
+            process.wait(timeout=30)
```
fishcakez, line 195: Should we make this configurable? Perhaps a new config like the other ones defined in __init__?

ankilosaurus: I was thinking about it but is it really needed? Cleanup should happen in reasonable time?

fishcakez: I think it should be configurable so that it does not need a deployd rollout if we need to change it.

fishcakez (approved):

PR #881, opened 2021-10-01
Too many hosts in a single environment

kynanlalone:
* restructure env landing page, host agent state, and host failed views 
* fix failed host UX with actions per deploy, confirmation modals for every button on failed hosts, change language to reflect the operation on failed hosts, add agent state tool tips for environment landing


kynanlalone-pinterest (commented):
merging into feature

PR #892, opened 2021-10-21
add .settings dir to .gitignore

ankilosaurus:


deploy-service/common/src/main/java/com/pinterest/deployservice/chat/SlackChatManager.java
```diff
@@ -47,6 +47,7 @@ private void postMessage(String from, String channel, String message) throws Exc
                     response = slack.methods(this.token).chatPostMessage(req -> req
                     .channel(channel)
                     .text(message)
+                    .username(from)
```
robbintt, line 50: intended?

PR #893, opened 2021-10-22
Initialize host info from facter once per deployd run and reuse

ankilosaurus:
Problem:
---
Currently facter is called every time deployd tries to reach to Teletraan server. facter call is expensive specially during new host bring up as it slows down autoscaling. 

Fix:
---
Populate host info from facter once during init and reuse if possible.
Reduces number of facter calls by 8x per service installed on a given host.

Test:
---
 Deploy test service and verify that deploy agent is able to load all the facts.


robbintt (approved):

robbintt (approved):

deploy-agent/deployd/client/client.py
```diff
@@ -44,8 +44,23 @@ def __init__(self, config=None, hostname=None, ip=None, hostgroup=None,
         self._autoscaling_group = None
         self._availability_zone = None
         self._stage_type = None
+        self._host_info_populated = False
+        try:
+            if self._read_host_info() is False:
```
robbintt, line 49: instead you can raise before `return False` below and get rid of the duplicated code, since it will hit the exception flow below. You can add a noop exception class above:
```
class HostReadFailedException(Exception):
    pass
```

PR #895, opened 2021-10-29
Dockerize deploy-agent unit tests.  Improve flaky test env.

robbintt:
Status:

This is ready to merge.

* Tox entrypoint works in Python2 and Python3.
* ./unit_tests_docker.sh runs unit tests in docker context
* Removed useless dead code Dockerfile
* Tox uses recreate=True which gets rid of flaky tox environments outside of docker.

Goals:
- Run tests in docker

Notes:
- Ubuntu image with deps is not necessarily permanent, we can improve this. Maybe we can use different dockerfiles for python2 and python3 and break up the tox runs.  Might be a good use case for docker compose.
- The tox setup needs improved (not part of this PR)
- The python env and pip env setups need improved (not part of this PR)


deploy-agent/unit_tests_docker.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -ex
+
+BUILD_NAME=deployd_debug
+docker build -t "$BUILD_NAME" . -f Dockerfile.unittest
+docker run -it "$BUILD_NAME" bash
```
robbintt, line 7: fix

deploy-agent/unit_tests_docker.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -ex
+
+BUILD_NAME=deployd_debug
```
robbintt, line 5: rename

ankilosaurus (approved):

PR #899, opened 2021-11-10
Let the fields that are readonly in the Cluster Config > Advanced Settings page be configurable

ryan953:
Previously we had hard-coded checks to ensure that `spiffe_id` was readonly, and could not be changed by the user.

For context "User Data Configs" section would render like this, so spiffe couldn't be edited or removed:
Screen Shot 2021-11-15 at 2 45 19 PM

This makes the list of readonly fields a configuration setting, instead of being a hard-coded value. 
By making it a configuration we can create more readonly fields that will behave the same way that spiffe_id does.

Enables internal config change: D812125 which is in support of: D809778

---

**Test updates to an existing Cluster**
1. Run teletraan and navigate to an existing cluster which has a spiffe_id. Any internal cluster will have one.
2. View the `/env/$ENV/$STAGE/config/cluster/config/` path, and click on "Advanced Settings"
  - notice the `spiffe_id` textfield is readonly
3. Try to submit a different `spiffe_id`
   - You can try to edit the html, submit your own POST request outside the browser with curl...
   - Updates to the existing `spiffe_id` user-data value should be rejected.

**Test creating a new Cluster form**
1. Visit the new-cluster page, maybe for a new Stage
2. Create a user-data field with the key `spiffe_id` and any value
3. Try to submit this user-data
  - User-data with the key `spiffe_id` should be rejected no matter how you submit it.


deploy-board/deploy_board/settings.py
```diff
@@ -230,6 +230,7 @@
 
 TELETRAAN_DISABLE_CREATE_ENV_PAGE = True if os.getenv("TELETRAAN_DISABLE_CREATE_ENV_PAGE", "false") == "true" else False
 TELETRAAN_REDIRECT_CREATE_ENV_PAGE_URL = os.getenv("TELETRAAN_REDIRECT_CREATE_ENV_PAGE_URL", None)
+TELETRAAN_CLUSTER_READONLY_FIELDS = os.getenv("TELETRAAN_CLUSTER_READONLY_FIELDS", "spiffe_id").split(",")
```
robbintt, line 233: need to support default `None` in case the customer does not use spiffe id

robbintt:
This looks good.

deploy-board/deploy_board/settings.py
ryan953: Making this default to `None` would be a change in behavior for customers.
Previously `spiffe_id` was hard-coded in a bunch of places to be a readonly field, this reduces how many times it's typed into the code, but maintains the same behavior as before.

If we change behavior then upgraders who use a `spiffe_id` field would need to update their configs to keep it read-only, or else risk having their own users be allowed to edit that field unexpectedly.

I think the change in behavior could be it's own diff, and come with the needed documentation/upgrade notes and version bump. This diff enables that change.

robbintt: Thanks, I agree about minimizing changing the behavior in this diff.

ankilosaurus (approved):

PR #900, opened 2021-11-11
update environs table schema in deploy database

liyaqin1:


deploy-service/common/src/main/resources/sql/deploy.sql
```diff
@@ -24,26 +23,32 @@ CREATE TABLE IF NOT EXISTS environs (
...
     deploy_constraint_id      VARCHAR(22),
     external_id          CHAR(36),
     allow_private_build TINYINT(1)    NOT NULL DEFAULT 0,
     ensure_trusted_build TINYINT(1)    NOT NULL DEFAULT 0,
-    PRIMARY KEY   (env_id)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+    stage_type          VARCHAR(32)   NOT NULL DEFAULT 'DEFAULT',
+    is_sox              TINYINT(1)    NOT NULL DEFAULT 0,
+    PRIMARY KEY   (env_id),
+    UNIQUE KEY sc_config_id (sc_config_id),
```
ankilosaurus, line 48: hi @liyaqin1 are these changes missing in migrations?

robbintt: Btw - migrations are here... ideally these would live in the same place in the future:

https://github.com/pinterest/teletraan/tree/2110c7c987faf5336a917781480da34c135494c4/tools/mysql

robbintt: We don't want to update this file, instead we should apply migrations before running the tests

PR #905, opened 2021-11-16
update schema version and create missing table in update sql 8 to mak…

liyaqin1:
…e sure upgrade.sh work well

### DIFF
--- a/tools/mysql/schema-update-8.sql
+++ b/tools/mysql/schema-update-8.sql
@@ -4,6 +4,32 @@
 
 # This script upgrade DB schema from version 7 to version 8
 
+CREATE TABLE IF NOT EXISTS groups (
+    group_name varchar(128) NOT NULL,
+    launch_config_id varchar(81) DEFAULT NULL,
+    last_update bigint(20) NOT NULL DEFAULT '0',
+    chatroom varchar(64) DEFAULT NULL,
+    watch_recipients varchar(1024) DEFAULT NULL,
+    email_recipients varchar(1024) DEFAULT NULL,
+    launch_latency_th int(11) NOT NULL DEFAULT '600',
+    instance_type varchar(64) DEFAULT NULL,
+    image_id varchar(128) DEFAULT NULL,
+    security_group varchar(128) DEFAULT NULL,
+    user_data text,
+    subnets varchar(128) DEFAULT NULL,
+    iam_role varchar(128) DEFAULT NULL,
+    assign_public_ip tinyint(1) NOT NULL DEFAULT '0',
+    asg_status varchar(64) NOT NULL DEFAULT 'UNKNOWN',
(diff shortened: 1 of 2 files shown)


liyaqin1:
The table groups is missing in the current file https://github.com/pinterest/teletraan/blob/master/deploy-service/common/src/main/resources/sql/deploy.sql which caused the failure of the execution of upgrade.sh.
Besides this, the change in update 5 and 6 are already in the sql. So, I also changed the current schema version from 4 to 6.
Tested on my devapp.

liyaqin1:
The table groups is missing in the current file https://github.com/pinterest/teletraan/blob/master/deploy-service/common/src/main/resources/sql/deploy.sql which caused the failure of the execution of upgrade.sh.
Besides this, the change in update 5 and 6 are already in the sql. So, I also changed the current schema version from 4 to 6.
Tested on my devapp.

tools/mysql/schema-update-8.sql
```diff
@@ -4,6 +4,32 @@
 
 # This script upgrade DB schema from version 7 to version 8
 
+CREATE TABLE IF NOT EXISTS groups (
```
ankilosaurus, line 7: hi @liyaqin1 Is this table used anywhere in the code?

liyaqin1: great point.
This table is not in teletraan anymore. Instead, it was moved to rodimus.
I updated the pr and test it again.
The upgrade.sh works well:
```
yaqinli@dev-yaqinli:~/code/teletraan/tools/mysql$ docker exec teletraan_pinterest_mysql /var/teletraan/tools/deploydb.sh
yaqinli@dev-yaqinli:~/code/teletraan/tools/mysql$ docker exec teletraan_pinterest_mysql /var/teletraan/tools/upgrade.sh
Will upgrade schema from version 6 to version 15
upgrading to version 7 with schema-update-7.sql...
upgrading to version 8 with schema-update-8.sql...
upgrading to version 9 with schema-update-9.sql...
upgrading to version 10 with schema-update-10.sql...
upgrading to version 11 with schema-update-11.sql...
upgrading to version 12 with schema-update-12.sql...
upgrading to version 13 with schema-update-13.sql...
upgrading to version 14 with schema-update-14.sql...
upgrading to version 15 with schema-update-15.sql...
Successfully upgraded schema from version 6 to version 15
```

tools/mysql/upgrade.sh
```diff
@@ -32,6 +32,7 @@ usage() {
 resolve_target_version() {
   # list all files like schema-update-N.sql, and use the largest N as the target version
   TARGET_VERSION=0
+  cd /var/teletraan/tools/
```
robbintt, line 35: Is this always run from a docker container? If not, this will not work

robbintt (approved):

PR #906, opened 2021-11-16
[DO NOT MERGE SEE COMMENTS] Add black formatter

jogo:
Integrate into tox so it's run as part of github actions a well. Adding
the black formatter should make the code more standard, easier to read
and easier to contribute new code since you can automatically format
your code by using black locally.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -37,3 +37,10 @@ jobs:
       - name: Tests
         run: |
           tox -v
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
+        with:
+          src: "./deploy-agent"
--- a/deploy-agent/tox.ini
+++ b/deploy-agent/tox.ini
@@ -1,6 +1,7 @@
 [tox]
-envlist=py27,py37,py38,py39
+envlist=py27,py37,py38,py39,black
 recreate=True
+skip_missing_interpreters=True
 
 [flake8]
 exclude=.tox
@@ -9,6 +10,11 @@ max-line-length=100
 [testenv]
 deps=-rrequirements_test.txt
 commands=py.test {posargs}
+
+[testenv:black]
+commands=black --check {posargs} .
+
+
 
 [gh-actions]
 python =
--- a/deploy-agent/deployd/__init__.py
+++ b/deploy-agent/deployd/__init__.py
@@ -3,9 +3,9 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#  
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
-#    
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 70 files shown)


robbintt:
I agree with adding this this.  As for turning it on -- let's merge this now and set the CI job failure to **not** block merge for now so we can make a decision on how to start blocking PR merge on black as a team.

jogo:
> I agree with adding this this. As for turning it on -- let's merge this now and set the CI job failure to **not** block merge for now so we can make a decision on how to start blocking PR merge on black as a team.

Applying black to all existing code is straight forward so i was going to update this PR to that a well

.github/workflows/python.yml
```diff
@@ -37,3 +37,10 @@ jobs:
       - name: Tests
         run: |
           tox -v
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
```
jogo, line 44: from https://black.readthedocs.io/en/stable/integrations/github_actions.html

robbintt:
Ok great -- I agree with ripping the band-aid here.

Questions:
1. Python 2 and 3 compatibility (I've only used black with Python 3)

jogo:
> Ok great -- I agree with ripping the band-aid here.
> 
> Questions:
> 
> 1. Python 2 and 3 compatibility (I've only used black with Python 3)
Yes, https://black.readthedocs.io/en/stable/faq.html#does-black-support-python-2 and target-version https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html?highlight=target-version#command-line-options

Note: this now has black running and all tests are passing (even under PY2)

> 
> Future:
> 
> 1. Dev velocity - single button (single shell script) method for running black with identical settings on local.  (should be handled by tox)

already done `tox -eblack`

deploy-agent/tox.ini
```diff
@@ -1,6 +1,7 @@
 [tox]
-envlist=py27,py37,py38,py39
+envlist=py27,py37,py38,py39,black
 recreate=True
+skip_missing_interpreters=True
```
jogo, line 4: so when we run tox locally if a python version isn't installed it doesn't throw an error

deploy-agent/tox.ini
```diff
@@ -10,6 +11,11 @@ max-line-length=100
 deps=-rrequirements_test.txt
 commands=py.test {posargs}
 
+[testenv:black]
```
jogo, line 14: use `tox -eblack` to run black locally

robbintt: Thanks, looks great.  CI is somewhat self documenting, is there an appropriate place in the teletraan repo docs? If not it can wait.

jogo: Not sure where the best spot is I see a README pointing to the wiki. So unless you have some guidance perhaps we wait?

robbintt: SGTM

robbintt:
Holding this until Jan 1. I will take it over to free @jogo up.

jogo:
Thank you @robbintt, much appreciated.

robbintt:
This was not prioritized while the deploy method was changed, now that the deploy method is changed, I think it can be quickly completed.

PR #909, opened 2021-11-30
Use getValue instead of lookup in Django template to avoid exception

liyaqin1:
issue:
we didn't catch exception which caused a horrible error showing up on teletraan ui:
image (7)

related jira ticket: https://jira.pinadmin.com/browse/CDP-4432

fix:
there is a line in host_details.html which tries to get a value from non-existing key:
{% with env=agent_wrapper.env agent=agent_wrapper.agent hostdetailsvalue=host_details.items|lookup:"Phobos Link" %}
this caused an exception and redirected the user to the error page.
in this pr, I used getValue instead of lookup to avoid the exception in Django template.

test:
before:
Screen Shot 2021-11-30 at 3 36 42 PM

after:
Screen Shot 2021-12-01 at 2 04 01 PM

### DIFF
--- a/deploy-board/deploy_board/webapp/host_views.py
+++ b/deploy-board/deploy_board/webapp/host_views.py
@@ -16,7 +16,6 @@ from django.shortcuts import render
 from django.http import HttpResponse
 from django.views.generic import View
 import logging
-import traceback
 from helpers import environs_helper, agents_helper, autoscaling_groups_helper
 from helpers import environ_hosts_helper, hosts_helper
 from deploy_board.settings import IS_PINTEREST, CMDB_API_HOST, CMDB_INSTANCE_URL, CMDB_UI_HOST, PHOBOS_URL
@@ -148,7 +147,6 @@ class GroupHostDetailView(View):
 
 class HostDetailView(View):
     def get(self, request, name, stage, hostname):
-      try:
         envs = environs_helper.get_all_env_stages(request, name)
         stages, env = common.get_all_stages(envs, stage)
         duplicate_stage = ''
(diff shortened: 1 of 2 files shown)


ankilosaurus:
hi @liyaqin1 can you please add description - 1. issue 2. fix 3. test

deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -167,24 +169,25 @@ def get(self, request, name, stage, hostname):
...
-            'show_force_terminate': IS_PINTEREST,
-            'asg_group': asg,
-            'is_unreachable': is_unreachable,
-            'pinterest': IS_PINTEREST,
-            'host_information_url': CMDB_UI_HOST,
-            'instance_protected': is_protected,
-            'host_details': host_details,
-            'duplicate_stage': duplicate_stage,
+      except:
+        pass
```
ankilosaurus, line 173: can we log exception ?

liyaqin1: added the log ^^

liyaqin1:
> 
1. issue:
we didn't catch exception which caused a horrible error showing up on teletraan ui:
image (7)
related jira ticket: https://jira.pinadmin.com/browse/CDP-4432

2. fix:
there is a line in host_details.html which tries to get a value from non-existing key:
{% with env=agent_wrapper.env agent=agent_wrapper.agent hostdetailsvalue=host_details.items|lookup:"Phobos Link" %}
this caused an exception and redirected the user to the error page.
in this pr, I just catched the exception and avoid using the agent_wrapper augument in the render function.

3. test:
before:
Screen Shot 2021-11-30 at 3 36 42 PM
after:
Screen Shot 2021-11-30 at 5 41 29 PM

deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -186,6 +188,25 @@ def get(self, request, name, stage, hostname):
             'host_details': host_details,
             'duplicate_stage': duplicate_stage,
         })
+      except:
+        log.error(traceback.format_exc())
+        return render(request, 'hosts/host_details.html', {
```
robbintt, line 193: Can we reduce to just 1 render after the exception, rather than having 2 calls to render whether it excepted?

robbintt: We spoke briefly about this, sounds like the exception is in the render. I think we should fix the problem by passing the right info into render() if possible, rather than re-render on exception.

robbintt: For example, in this template, you can check if the key exists. If not, then don't render the line. you can use the Django Template language if statement
```fix:
there is a line in host_details.html which tries to get a value from non-existing key:
{% with env=agent_wrapper.env agent=agent_wrapper.agent hostdetailsvalue=host_details.items|lookup:"Phobos Link" %}```

robbintt: Another option is check in the dictionary being passed to `render()`. If the dictionary does not have the key, you could add the key with an empty value, but I don't think this way is as good.

liyaqin1: agree

robbintt (approved):

PR #912, opened 2021-12-03
add @ to ping the operator

liyaqin1:
issue: displaying name in the slack notification is not enough. it is better to ping the operator as well.
fix: add @ and the special tag <> to ping the operator.
test:
deployed the pr to teletraan integ and got pinged as expected.
Screen Shot 2021-12-03 at 1 52 24 PM


yongwen (approved):

yongwen:
maybe also add "operated by" for the failure case.

liyaqin1:
> maybe also add "operated by" for the failure case.

Hey Yongwen, I found that the notification tasks are done by teletraan service and teletraan worker together.
After I deployed the pr to teletraan worker, I never see this issue again.
Screen Shot 2021-12-08 at 11 47 51 AM

PR #913, opened 2021-12-04
Fix more abstract host problems.

robbintt:
Fix some more problems from abstract_host being treated as unique.

Hardcoded default reference is out of scope as we cannot use the host_type ID there. It is unique per service.   The PR won't address the larger problem with parsing abstract name strings in the deploy-board control flow.

Manual Testing:
- Cluster settings should should no longer fail to render correctly
- Cluster settings form should select the correct host type
- New cluster creation should work normally (no unexpected side effects)

### DIFF
--- a/deploy-board/deploy_board/settings.py
+++ b/deploy-board/deploy_board/settings.py
@@ -284,7 +284,8 @@ if IS_PINTEREST:
     DEFAULT_CMP_IMAGE = 'cmp_base-ebs-18.04'
 
     #Pinterest Default Host Type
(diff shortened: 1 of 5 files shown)


robbintt:
The downstream problems are worse than expected. The fixed reference in rodimus is now causing the form to default to the top entry again, presumably because the comparison fails between abstract_name in the form and the cluster's host_type.

robbintt:
Testing is complete, this is ready to merge concurrently with internal cluster management API changes.

liyaqin1 (approved):

ankilosaurus (approved):

PR #917, opened 2021-12-14
Add timestamps to the logs for downloader and stager

jogo:
Make it easier to understand how long things take from the logs, by
updating the default log format to include a timestamp. Note after this
change each env_name log will have more timestamps but many lines still
won't have timestamps if they are coming from the scripts (RESTARTING
etc.)

https://docs.python.org/3/library/logging.html#logrecord-attributes

### DIFF
--- a/deploy-agent/deployd/common/__init__.py
+++ b/deploy-agent/deployd/common/__init__.py
@@ -12,3 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


robbintt (approved):
LGTM

ankilosaurus:
Version bump before landing otherwise you will need another PR for it.

jogo:
I am happy for this to wait until the next release and don't need a special release just for this.

PR #921, opened 2022-01-15
move teletraan base image to u18

liyaqin1:


liyaqin1:
the dockerfile is in teletraanconfig

PR #923, opened 2022-01-28
Metrics caching

kynanlalone:
Add caching functionality to metric collection and sending

- Uses simple health checking before sending any metrics
- If health check fails, a cache file is written to 
- When the health check succeeds the cache file is flushed and metrics are sent normally

### DIFF
--- a/deploy-agent/deployd/agent.py
+++ b/deploy-agent/deployd/agent.py
@@ -180,7 +180,6 @@ class DeployAgent(object):
                 time.sleep(self._config.get_daemon_sleep_time())
                 self.load_status_file()
 
-
     def serve_once(self):
         log.info("Running deploy agent in non daemon mode")
         try:
@@ -206,7 +205,6 @@ class DeployAgent(object):
             if envId == value.report.envId:
                 return name
         return None
-
 
     def process_deploy(self, response):
         self.stat_time_elapsed_internal.resume()
@@ -427,6 +425,7 @@ class DeployAgent(object):
 
         return False
 
+
 # make sure only one instance is running
 instance = SingleInstance()
 
@@ -518,6 +517,7 @@ def main():
     # timing stats - agent exit time
     create_sc_timing('deployd.stats.internal.time_end_sec',
                      int(time.time()))
+
 
 if __name__ == '__main__':
     main()
--- a/deploy-agent/deployd/agent.py
+++ b/deploy-agent/deployd/agent.py
@@ -180,6 +180,7 @@ class DeployAgent(object):
                 time.sleep(self._config.get_daemon_sleep_time())
                 self.load_status_file()
 
+
     def serve_once(self):
         log.info("Running deploy agent in non daemon mode")
         try:
@@ -205,6 +206,7 @@ class DeployAgent(object):
             if envId == value.report.envId:
                 return name
         return None
+
 
     def process_deploy(self, response):
         self.stat_time_elapsed_internal.resume()
@@ -425,7 +427,6 @@ class DeployAgent(object):
 
         return False
 
-
 # make sure only one instance is running
 instance = SingleInstance()
 
@@ -517,7 +518,6 @@ def main():
     # timing stats - agent exit time
     create_sc_timing('deployd.stats.internal.time_end_sec',
                      int(time.time()))
-
 
 if __name__ == '__main__':
     main()
--- a/deploy-agent/deployd/common/stats.py
+++ b/deploy-agent/deployd/common/stats.py
@@ -109,7 +109,7 @@ class MetricCache:
     """ local cache for metrics
         creates empty cache file
     """
-    def __init__(self, path):
+    def __init__(self, path=METRIC_CACHE_PATH):
         self.path = path
         # maximum cache size in bytes
         self.max_size = (10 * 1024 * 1024)
--- a/deploy-agent/deployd/common/stats.py
+++ b/deploy-agent/deployd/common/stats.py
@@ -13,9 +13,27 @@
 # limitations under the License.
 
 import logging
-from deployd import IS_PINTEREST
-from deployd import __version__
+from deployd import __version__, IS_PINTEREST, METRIC_PORT_HEALTH, METRIC_CACHE_PATH
 import timeit
+import socket
+import json
+import os
+
+if IS_PINTEREST:
+    from pinstatsd.statsd import sc, sc_v2
+else:
+    class sc:
+        @staticmethod
+        def increment(name, sample_rate, tags):
+            pass
+
+        @staticmethod
+        def gauge(name, value, sample_rate=None, tags=None):
+            pass
+
+
+    class sc_v2(sc):
+        pass
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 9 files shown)


deploy-agent/deployd/agent.py
```diff
@@ -427,6 +425,7 @@ def plan_changed(old_response, new_response):
 
         return False
 
+
```
ankilosaurus, line 428: revert white space.

deploy-agent/deployd/agent.py
```diff
@@ -180,7 +180,6 @@ def serve_forever(self):
                 time.sleep(self._config.get_daemon_sleep_time())
                 self.load_status_file()
 
-
```
ankilosaurus, line 183: revert white space.

deploy-agent/deployd/agent.py
```diff
@@ -207,7 +206,6 @@ def _resolve_deleted_env_name(self, envName, envId):
                 return name
         return None
 
-
```
ankilosaurus, line 210: revert white space.

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+def create_sc_timing(name, value, sample_rate=1.0, tags=None):
     if IS_PINTEREST:
-        try:
-            from pinstatsd.statsd import sc
-            tags = _add_default_tags(tags)
-            return sc.timing(stat, value, sample_rate=sample_rate, tags=tags)
-        except Exception as e:
-            error_msg = str(e)
-            log.error('unable to send metric: {}'.format(error_msg))
+        mtype = 'timing'
```
ankilosaurus, line 72: all 3 create_  methods have basically same code. Can we parameterize mtype and cleanup?

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+        client.send(mtype=mtype,
+                    name=name,
+                    value=value,
+                    sample_rate=sample_rate,
+                    tags=tags)
     else:
         return
 
 
+class Cache:
```
ankilosaurus, line 108: rename is to MetricsCache.

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+    """ local cache for metrics
+        creates empty cache file
+    """
+    def __init__(self, path):
+        self.path = path
```
ankilosaurus, line 113: what's default path ?

deploy-agent/deployd/types/deploy_goal.py
```diff
@@ -50,6 +49,26 @@ def __init__(self, jsonValue=None):
             self.firstDeploy = jsonValue.get('firstDeploy')
             self.isDocker = jsonValue.get('isDocker')
 
+    def __eq__(self, other):
```
ankilosaurus, line 52: why is this change?

deploy-agent/deployd/common/stats.py
kynanlalone: I tried this change but felt the resulting code was not as readable as keeping the original structure.  It also changed all the calls.  Let me know if you really want it!

deploy-agent/deployd/common/stats.py
kynanlalone: Great suggestion.  Done!  That makes it more readable.

deploy-agent/deployd/common/stats.py
kynanlalone: It is set by the MetricClient.  Added a commit to this PR to set a default in the MetricCache class to make it more obvious.

deploy-agent/deployd/types/deploy_goal.py
kynanlalone: Change is so retry loop timers are squashed instead of reported separately.
https://github.com/pinterest/teletraan/blob/master/deploy-agent/deployd/agent.py#L369

deploy-agent/deployd/common/stats.py
ankilosaurus: default will be none as per `METRIC_CACHE_PATH = os.getenv('METRIC_CACHE_PATH', None)`.

deploy-agent/deployd/common/stats.py
ankilosaurus: I don't understand why this methods should even exist after introducing MetricsClient.

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+                    tags=tags)
+
+    def _send_mtype(self):
+        """ send metric to sc using corrected
+            calls per metric type
+        """
+        func = getattr(sc, self.stat.mtype)
+        func_v2 = getattr(sc_v2, self.stat.mtype)
+
+        if self.stat.mtype == 'increment':
```
ankilosaurus, line 265: use enum for mtype.

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+
+    def _flush_cache(self):
+        """ read from cache, send every metric, truncate cache """
+        log.warning('flushing metrics from cache')
+        for stat in self.cache.read():
+            if not stat.deserialize():
+                # unable to parse stat, skip
+                log.error('unable to parse stat: {}'.format(stat))
+                continue
+            self.stat = stat
```
ankilosaurus, line 287: pass stat as parameter to seld._send(stat) instead of defining member variable.

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+        """ add default tags, send metric, write to, or flush cache
+            depending on health check """
+        tags = self._add_default_tags(tags)
+        if self.is_healthy():
+            # trigger flush cache first
+            if self.cache and (self.cache.exists() and not self.cache.is_empty()):
+                self._flush_cache()
+            # send metric
+            self.stat = self._parse_stat(mtype=mtype, name=name, value=value, sample_rate=sample_rate, tags=tags)
+            self._send()
```
ankilosaurus, line 308: pass stat as parameter to seld._send(stat) instead of defining member variable.

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+            return False
+
+
+class MetricClient:
+    """ metrics client wrapper, enables disk cache """
+
+    def __init__(self, port=METRIC_PORT_HEALTH, cache_path=METRIC_CACHE_PATH):
+        self.port = port
+        self.cache = MetricCache(path=cache_path)
+        self.stat = None
```
ankilosaurus, line 233: what's the use of stat if it's lifetype is limited to _send()

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
+        self.sample_rate = sample_rate
+        self.tags = tags
+        self.ins = ins
+        try:
+            self.JSONDecodeError = json.decoder.JSONDecodeError
+        except AttributeError:
+            # python2 support
+            self.JSONDecodeError = ValueError
+
+    def __repr__(self):
```
jogo, line 185: repr should return a string https://docs.python.org/3/reference/datamodel.html#object.__repr__

deploy-agent/deployd/common/stats.py
```diff
@@ -28,69 +49,287 @@ def __exit__(self, stats, sample_rate=1.0, tags=None):
...
-        return tags
-    if __version__ and not tags:
-        tags = dict()
-    tags['deploy_agent_version'] = __version__
-    return tags
+def create_stats_timer(name, sample_rate=1.0, tags=None):
+    if IS_PINTEREST:
+        from pinstatsd.statsd import statsd_context_timer
 
+        class TimingOnlyStatClient:
```
jogo, line 56: why define a class inside a function?

deploy-agent/deployd/types/build.py
```diff
@@ -43,6 +43,26 @@ def __init__(self, jsonValue=None):
             self.publishInfo = jsonValue.get('publishInfo')
             self.publishDate = jsonValue.get('publishDate')
 
+    def __eq__(self, other):
```
jogo, line 46: FYI see the warning about __hash__ here https://docs.python.org/3/reference/datamodel.html#object.__hash__

Also usually good to have an `isinstance` check so that comparing different types will return False and not an exception.

deploy-agent/deployd/common/stats.py
kynanlalone: In this case it was done for readability.  In retrospect, that's bad form.  Moved it out.  Thanks!

deploy-agent/deployd/common/stats.py
kynanlalone: These are functions that are called by the existing code.  I kept the entry points in place to avoid refactoring too much of the downstream code.

deploy-agent/deployd/common/stats.py
kynanlalone: Valid point!  Added exceptions to raise when MetricClient or MetricCache encounter a configuration error in a later commit.  Thanks!

deploy-agent/deployd/common/stats.py
kynanlalone: Huge oversight.  Thanks.  I refactored the code to not even use __repr__ which can be problematic for this use case to begin with.

deploy-agent/deployd/common/stats.py
kynanlalone: the data time enum was introduced to the standard library in python 3.4

deploy-agent/deployd/types/build.py
kynanlalone: Refactored __eq__ and added a __hash__ for both DeployGoal and Build.  Thanks.

deploy-agent/deployd/common/stats.py
kynanlalone: Stat is used by MetricClient and MetricCache (MetricClient.parse_stat and MetricCache.read) and this member variable is used by both flushing cache, reading, and sending metrics.

deploy-agent/deployd/common/stats.py
kynanlalone: The existing logic also reads from this member variable in other places as well, so changing this to a param means a lot of passing it around where it's needed

deploy-agent/deployd/common/stats.py
kynanlalone: The existing logic also reads from this member variable in other places as well, so changing this to a param means a lot of passing it around where it's needed

deploy-agent/tests/unit/deploy/common/test_stats.py
```diff
@@ -119,13 +126,23 @@ def test_deserialize(self):
...
 
 
+class TestMetricClientExceptions(unittest.TestCase):
+    def test__MetricClientConfigurationError(self):
+        with self.assertRaises(MetricClientConfigurationError):
+            MetricClient(port=None, cache_path=None)
+
+
 class TestMetricClient(unittest.TestCase):
     cache_path = 'tests/unit/deploy/common/test_stats.cache'
```
jogo, line 136: https://docs.python.org/3/library/tempfile.html may be useful

deploy-agent/deployd/types/deploy_goal.py
```diff
@@ -49,25 +49,30 @@ def __init__(self, jsonValue=None):
...
-            return True
-        return False
+        """ compare DeployGoals """
+        return isinstance(other, DeployGoal) \
+            and self.__key() == other.__key()
 
     def __ne__(self, other):
-        return not self.__eq__(other)
+        """ compare DeployGoals """
+        return not self.__key() == other.__key()
```
jogo, line 75: should this have an isinstance check as well?

deploy-agent/tests/unit/deploy/common/test_stats.py
```diff
@@ -146,53 +163,49 @@ def test_is_healthy(self, mock_connect_ex):
...
-    def tearDown(self):
-        MetricCache(self.cache_path)
-        os.remove(self.cache_path)
-
 
 class TestTimeElapsed(unittest.TestCase):
+    def setUp(self):
+        self.elapsed = TimeElapsed()
+
+    def tearDown(self):
```
jogo, line 171: is this teardown needed?

deploy-agent/tests/unit/deploy/common/test_stats.py
kynanlalone: although a good suggestion skipping here to match expected inputs that this test case covers

deploy-agent/tests/unit/deploy/common/test_stats.py
kynanlalone: no it's not.  removed.

deploy-agent/deployd/types/deploy_goal.py
kynanlalone: yes it should.  added.  thanks!

ankilosaurus (approved):

robbintt:
PR Approved including changes up to and including bef82e0.

PR #928, opened 2022-02-05
Remove constraint allowing only one of alarm autoscaling or time based autoscaling

robbintt:
AWS autoscaling supports scheduled scaling and alarm autoscaling together. Remove the restrictions preventing the concurrent use of these features. Fix any issues.

DONE:
* FIXED - fix alarm "Save" form submit button. it does not allow update of existing alarms. New alarms "Add Alarm" button is not affected.
* CAN NOT REPRO - test that alarms and time based scaling can be put on the ASG together by deploy-board form submit.  (Can not repro: this was possibly, but not definitely, related to custom metric ~25 minute scaling latency at that time.)

Test plan:
* manual testing

### DIFF
--- a/deploy-board/deploy_board/templates/groups/asg_config.tmpl
+++ b/deploy-board/deploy_board/templates/groups/asg_config.tmpl
@@ -130,7 +130,7 @@ function loadAsgAlarms() {
 }
 
 function getAdvancedSetting() {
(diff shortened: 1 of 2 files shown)


robbintt:
There is an issue with the alarm form submit button.

ankilosaurus (approved):

PR #929, opened 2022-02-08
modify Stat parsing to be more defensive with bad data

kynanlalone:
* modify Stat parsing to be more defensive with bad data
* Account for a rare possible case where cache has been modified outside of deploy-agent and includes valid json structure with invalid properties
* Add unit tests for control characters and invalid types

### DIFF
--- a/deploy-agent/deployd/common/stats.py
+++ b/deploy-agent/deployd/common/stats.py
@@ -205,14 +205,17 @@ class Stat:
 
     def _deserialize(self):
         """ read in json, setting defaults for a stat
-            return: None
+            return: bool
         """
         obj = json.loads(self.ins)
-        self.mtype = obj.get('mtype', None)
-        self.name = obj.get('name', None)
-        self.value = obj.get('value', None)
-        self.sample_rate = obj.get('sample_rate', None)
-        self.tags = obj.get('tags', None)
+        if isinstance(obj, dict):
+            self.mtype = obj.get('mtype', None)
+            self.name = obj.get('name', None)
+            self.value = obj.get('value', None)
+            self.sample_rate = obj.get('sample_rate', None)
+            self.tags = obj.get('tags', None)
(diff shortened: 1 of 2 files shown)


deploy-agent/deployd/common/stats.py
```diff
@@ -205,14 +205,17 @@ def serialize(self):
 
     def _deserialize(self):
         """ read in json, setting defaults for a stat
-            return: None
+            return: bool
         """
         obj = json.loads(self.ins)
```
robbintt, line 210: better to be explicit here, pass `ins` in the function signature, since this function is internal you can pass self.ins whenever you invoke it. this will give the reader an easier time in the future

deploy-agent/deployd/common/stats.py
```diff
@@ -222,12 +225,14 @@ def deserialize(self, ins=None):
         if ins:
             self.ins = ins
         try:
-            self._deserialize()
-            return True
+            valid = self._deserialize()
+            if valid:
```
robbintt, line 229: since valid is not reused, you can simply `if self._deserialize():`

deploy-agent/deployd/common/stats.py
```diff
@@ -205,14 +205,17 @@ def serialize(self):
 
     def _deserialize(self):
         """ read in json, setting defaults for a stat
-            return: None
+            return: bool
```
robbintt, line 208: I think there is a better way to architect this by using try/except, but it's more of a python nit.

robbintt (approved):
Approved with some improvement ideas for the future.

deploy-agent/deployd/common/stats.py
kynanlalone: I agree, I actually visited this idea but it's a lot of refactoring.  I'll make a note to revisit.

PR #932, opened 2022-02-10
First run stickiness

kynanlalone:
Stickiness for first_run.  Only re-evaluate first_run when it is False (or None).  This ensures first_run carries through for the entire first run.  Re-evaluates in the highly unlikely event envs are lost during a run.

### DIFF
--- a/deploy-agent/deployd/agent.py
+++ b/deploy-agent/deployd/agent.py
@@ -69,7 +69,7 @@ class DeployAgent(object):
         self.stat_time_elapsed_total = TimeElapsed()
         self.stat_stage_time_elapsed = None
         self.deploy_goal_previous = None
-        self._first_run = None
+        self._first_run = False
         self._helper = helper or Helper(self._config)
         self._STATUS_FILE = self._config.get_env_status_fn()
         self._client = client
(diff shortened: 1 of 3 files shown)


deploy-agent/deployd/agent.py
```diff
@@ -87,10 +88,22 @@ def load_status_file(self):
         self._config.update_variables(self._curr_report)
 
     def first_run(self):
-        """ check if this the very first run of agent on this instance """
+        """ check if this the very first run of agent on this instance
+            supports sticky Truth: re-evaluate first run
+            only if self._first_run is None or False
+            return: bool self._first_run
+        """
+        if self._first_run:
```
robbintt, line 96: i think you can set the `__init__` value to false and collapse these 2 if cases into 
```
if self._first_run or not self._envs:
    self._first_run = True
```
and get rid of the else

robbintt: This also gets rid of the possibility of the `None` value which is confusing since it's never an evaluated state.

deploy-agent/deployd/agent.py
```diff
@@ -87,10 +88,22 @@ def load_status_file(self):
         self._config.update_variables(self._curr_report)
 
     def first_run(self):
```
robbintt, line 90: you can make this an `@property` for clarity for the user, although the practical value is dubious without a setter

robbintt: this change is definitely recommended for working with `_first_run` private state, then you can drop the `()` off all references to `first_run` in the codebase. I think this is a good time to do it since you are testing all the cases anyways.

deploy-agent/deployd/agent.py
kynanlalone: Both great suggestions.  Added.  Thank you!

deploy-agent/deployd/agent.py
kynanlalone: Both great suggestions.  Added.  Thank you!

deploy-agent/deployd/agent.py
```diff
@@ -86,11 +87,16 @@ def load_status_file(self):
...
 
+    @property
     def first_run(self):
-        """ check if this the very first run of agent on this instance """
-        if not self._envs:
-            return True
-        return False
+        """ check if this the very first run of agent on this instance
+            supports sticky Truth: re-evaluate first run
+            only if self._first_run is None or False
```
robbintt, line 94: comment fix

tylerwowen (commented):
This PR is approved.

PR #937, opened 2022-02-18
Replaced the default Metrics Dashboard link with Observability Hub

tylerwowen:
## Changes

The UI now populates the metrics dashboard button with the generated hub url (in the format of `https://statsboard.pinadmin.com/hub?host_type=<host_type>`) if there is no existing configuration and the environment is not a sidecar. The `host_type` is acquired using statsboard's API (`https://statsboard.pinadmin.com/api/v1/teletraan_host_types?env=<envname>&stage=<stagename>`) based on the env and stage. 

**Note**: I observed that the existing code has bug when determine the existence of a dashboard. Since dashboard no longer presents _dashboard does not exist_, the Metrics Dashboard button always shows. 

## Testing
Started a local instance and tested manually:

- [x] When hostType API returns empty list, no button is presented
   none
- [x] When hostType API returns 1 result, the buttonUrl is properly populated
   1result
- [ ] When hostType API returns multiple results, only the first result is used (didn't find an example to perform this test yet)
- [x] When a metrics dashboard is configured, the configured one is used.
   configured
- [x] Sidecar env: button shows when there is metrics dashboard configured. 
- [x] Sidecar env: button doesn't show when there is NO metrics dashboard configured.


ankilosaurus (approved):

ankilosaurus (changes requested):
If service is a sidecar we don't want to generate link based on host type

deploy-board/deploy_board/webapp/service_add_ons.py
```diff
@@ -430,18 +438,20 @@ def getLogHealthReport(configStr, report):
...
+    hostType = None
+    if metricsDashboardUrl is None:
+        if isSidecar:
+            state = ServiceAddOn.UNKNOWN
+        else:
+            try:
+                hostType = getStatsboardHostType(env, stage)
+            except:
+                state = ServiceAddOn.UNKNOWN
+            if hostType is None:
```
ankilosaurus, line 452: This can be in try block.

ankilosaurus (approved):

deploy-board/deploy_board/webapp/service_add_ons.py
tylerwowen: Done

tylerwowen (commented):
This PR is approved.

osoriano (commented):
This PR is approved.

PR #941, opened 2022-03-07
use NotNull to allow empty list

liyaqin1:
We found a bug when removing the last remaining host/group. Fixed it via using NotNull instead of NotEmpty.

### DIFF
--- a/deploy-service/teletraanservice/pom.xml
+++ b/deploy-service/teletraanservice/pom.xml
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>5.4.1.Final</version>
+            <version>6.0.23.Final</version>
         </dependency>
         <dependency>
             <groupId>com.pinterest.teletraan</groupId>
--- a/deploy-service/teletraanservice/pom.xml
+++ b/deploy-service/teletraanservice/pom.xml
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>6.0.23.Final</version>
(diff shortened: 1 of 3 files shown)


robbintt:
Can you share some more context about the bug this solves in the PR description? I want to understand it better. Thanks!

liyaqin1:
> Can you share some more context about the bug this solves in the PR description? I want to understand it better. Thanks!

After bumping dropwizard, I only replaced @NotEmpty with @NotNull for enum types.
In this PR, I also replaced @NotEmpty with @NotNull for collection types.
For now, all @NotEmpty are used for int or String which is good.

Also, I revert this commit https://github.com/pinterest/teletraan/commit/8c0979c1e6277935cee6c6da337d5b34f4a017a1 to make sure the instance could be launched on devapp.

deploy-service/teletraanservice/pom.xml
```diff
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>6.0.23.Final</version>
+            <version>5.4.1.Final</version>
```
robbintt, line 54: Why reduce the version of the `hibernate-validator`?

liyaqin1: I used this version 5.4.1.Final at the very beginning and then github just bumped it for me here automatically: https://github.com/pinterest/teletraan/commit/8c0979c1e6277935cee6c6da337d5b34f4a017a1
After I approved and test locally, I found it could not launch the instance on my devapp. That's why I revert the version again here. ^^

liyaqin1: ok, let me do it

robbintt (approved):

PR #943, opened 2022-03-12
Switch executor sleep to subprocess.wait

jogo:
When using time.sleep we are forced to wait for the full sleep_time, but
if we use subprocess.Popen.wait we will wait _up_ to sleep_time.
(https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait)

With PROCESS_POLL_INTERVAL defaulting to 2 seconds, switching to this
logic can save up to 2 seconds for very fast executors, which adds up
with many stages and services on a single box.

Update test cases to use full paths for shell commands as we aren't
using shell=True

### DIFF
--- a/deploy-agent/deployd/common/executor.py
+++ b/deploy-agent/deployd/common/executor.py
@@ -21,6 +21,8 @@ import stat
 import time
 import traceback
 
+from future.utils import PY3
+
 from deployd.common.types import DeployReport, PingStatus, PRE_STAGE_STEPS, AgentStatus
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


ankilosaurus (approved):

deploy-agent/deployd/common/executor.py
```diff
@@ -108,7 +110,14 @@ def run_cmd(self, cmd, **kw):
 
                         # sleep some seconds before next poll
                         sleep_time = self._get_sleep_interval(start, self.PROCESS_POLL_INTERVAL)
-                        time.sleep(sleep_time)
+                        if PY3:
```
ankilosaurus, line 113: Test w/ py3 before merging.

robbintt:
This PR is approved.

PR #945, opened 2022-03-15
Update _read_host_info to call facter less

jogo:
After a Client that uses facter is initialized we don't want to keep
calling facter every time send_reports runs. If facter fails for any
reason and a field is missing the next call will try and load the
missing field again (related to #341).

* If keys_to_fetch is empty don't call facter at all
* Since stage_type might not exist, only try fetching it once.

### DIFF
--- a/deploy-agent/deployd/__init__.py
+++ b/deploy-agent/deployd/__init__.py
@@ -19,4 +19,4 @@ IS_PINTEREST = True if os.getenv("IS_PINTEREST", "false") == "true" else False
 METRIC_PORT_HEALTH = int(os.getenv('METRIC_PORT_HEALTH')) if os.getenv('METRIC_PORT_HEALTH', False) else None
 METRIC_CACHE_PATH = os.getenv('METRIC_CACHE_PATH', None)
 
-__version__ = '1.2.38'
+__version__ = '1.2.39'
--- a/deploy-agent/deployd/client/client.py
+++ b/deploy-agent/deployd/client/client.py
@@ -44,6 +44,9 @@ class Client(BaseClient):
         self._autoscaling_group = None
         self._availability_zone = None
         self._stage_type = None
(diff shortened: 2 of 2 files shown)


ankilosaurus (changes requested):
I think we don't want to hit facter at all if it was previously fetched (for all keys) since host data should not change once initialized in facter.

Can you use https://github.com/pinterest/teletraan/pull/945/files plz?

jogo:
> I think we don't want to hit facter at all if it was previously fetched (for all keys) since host data should not change once initialized in facter.
> 
> Can you use https://github.com/pinterest/teletraan/pull/945/files plz?

If you are referring to https://github.com/pinterest/teletraan/pull/893, in case facter returns partial results or some other unexpected error I figured more fine grained logic would allow for retrying those cases. That being said I am not tied to any specific implementation of the fix here, happy with any solution.

ankilosaurus:
> > I think we don't want to hit facter at all if it was previously fetched (for all keys) since host data should not change once initialized in facter.
> > Can you use https://github.com/pinterest/teletraan/pull/945/files plz?
> 
> If you are referring to #893, in case facter returns partial results or some other unexpected error I figured more fine grained logic would allow for retrying those cases. That being said I am not tied to any specific implementation of the fix here, happy with any solution.

Oops, yes, I was referring to #893. 

however, here you are just checking for stage_type and not others? 

I am ok if we want to be granular in that case call to facter can be wrapped with 

`            if keys_to_fetch:
                facter_data = utils.get_info_from_facter(keys_to_fetch)
`

ankilosaurus:
In fact i think, this is a better fix
`
if keys_to_fetch: 
      facter_data = utils.get_info_from_facter(keys_to_fetch)
       # read facter_data #
`

ankilosaurus (approved):

tylerwowen (commented):
This PR is approved.

PR #948, opened 2022-03-16
remove button to show commits between builds if it is the single buil…

liyaqin1:
…d page

### DIFF
--- a/deploy-board/deploy_board/templates/deploys/deploy_build.html
+++ b/deploy-board/deploy_board/templates/deploys/deploy_build.html
@@ -39,7 +39,7 @@
         <h4 class="panel-title pull-left">Build Details</h4>
     </div>
     <div class="panel-body table-responsive">
-        {% include "builds/pick_a_build.tmpl" with deployState=deployState overridePolicy=overridePolicy %}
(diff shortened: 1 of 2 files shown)


liyaqin1:
tested on integ teletraan.
before this pr, the build page shows 3 buttons and the builds page also shows 3 buttons.
after this pr, the build. page shows only 2 buttons which is expected and the builds page shows 3 buttons which is also expected:
Screen Shot 2022-03-16 at 11 56 28 AM
Screen Shot 2022-03-16 at 11 56 43 AM

deploy-board/deploy_board/templates/deploys/deploy_build.html
```diff
@@ -39,7 +39,7 @@ <h4 class="panel-title pull-left">Deploy a Build</h4>
         <h4 class="panel-title pull-left">Build Details</h4>
     </div>
     <div class="panel-body table-responsive">
-        {% include "builds/pick_a_build.tmpl" with deployState=deployState overridePolicy=overridePolicy %}
+        {% include "builds/pick_a_build.tmpl" with deployState=deployState overridePolicy=overridePolicy singleBuildPage=1 %}
```
tylerwowen, line 42: nit: It would be more explicit to use `True`

robbintt (approved):

PR #949, opened 2022-03-16
use True instead of 1

liyaqin1:
according to https://stackoverflow.com/questions/11804315/how-to-pass-boolean-keyword-argument-along-with-the-use-of-include-template-ta, seems like we should use True instead of 1 since our python requirements shows Django==1.11.29


liyaqin1:
according to https://stackoverflow.com/questions/11804315/how-to-pass-boolean-keyword-argument-along-with-the-use-of-include-template-ta, seems like we should use True instead of 1 since our python requirements shows Django==1.11.29

tylerwowen (approved):

PR #950, opened 2022-03-18
Use radio buttons to separate list filter and cluster configuration 

tylerwowen:
Previously in the UI, 'Assign Public IP' checkbox also determined what type of subnets (public or private) to show in the drop down list. In other words, if 'Assign Public IP' is checked, users are only supposed to choose from a list of public subnets. This guarantee however was broken by a bug `var arr = assignPublicIp == 'true' ? this.allPublic : this.allPrivate`. `assignPublicIp` can be a boolean and thus the expression is always evaluated to false and only private subnets are listed. As a consequence, people are allowed to choose private subnet with public IP and public subnet without public IP. Although we don't think these 2 possible combinations are practical for user scenarios, we still keep them available to minimize the impacts of this change.

Here is a description of existing and changed behaviors:
1. [Existing] During cluster creation, `Assign Public IP` checkbox is bond to the display of private or public subnets. new_adv
2. [Changed] When users try to modify an existing cluster, there are 2 separate controls: (1) A radio button that controls the display of public or private subnets. (2) `Assign Public IP` checkbox remains.
    1. When the page is first loaded, the ratio button's state is determined by the cluster subnet configuration. config_simple_init
    2. When user selects public subnet, only public subnets show up in the drop down menu and `Assign Public IP` is checked. config_adv_public_drop
    3. When user selects private subnet, only private subnets show up in the drop down menu and `Assign Public IP` is unchecked. config_adv_private_drop
    4. When user changes existing subnet type, an inline warning shows. warning_change_type
    5. When user checks `Assign Public IP` with private subnets or unchecks with public subnets, a banner shows on top of the screen to warn user to double check and ask the user to report their scenario to #teletraan. Note that users are still able to proceed with the warning. config_adv_warning

### DIFF
--- a/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
+++ b/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
@@ -326,10 +326,10 @@ Vue.component("placements-select", {
     </span></div>\
     </div>\
     <div class="col-xs-2" v-if="showsubnettype">\
-        <input type="radio" id="public" value="public" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
+        <input type="radio" id="public-subnet" value="public" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
         <label for="public">Public subnets</label>\
         <br>\
-        <input type="radio" id="private" value="private" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
+        <input type="radio" id="private-subnet" value="private" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
         <label for="private">Private subnets</label>\
         <br>\
     </div>\
--- a/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
+++ b/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
@@ -323,9 +323,9 @@ Vue.component("placements-select", {
     <button class="deployToolTip btn btn-default" type="button" data-toggle="tooltip" v-on:click="helpClick" title="click to see more information">\
         <span class="glyphicon glyphicon-question-sign"></span>\
     </button>\
-   </span></div>\
+    </span></div>\
     </div>\
-    <div class="col-xs-2">\
+    <div class="col-xs-2" v-if="showsubnettype">\
         <input type="radio" id="public" value="public" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
         <label for="public">Public subnets</label>\
         <br>\
@@ -337,7 +337,7 @@ Vue.component("placements-select", {
         <input type="checkbox" id="checkbox" v-bind:checked="assignpublicip" v-on:click="assignipchange($event.target.checked)">\
         <label for="checkbox">Assign Public IP</label>\
     </div></div>',
-    props: ['label', 'title', 'selectoptions', 'showhelp', 'assignpublicip', 'subnettype'],
+    props: ['label', 'title', 'selectoptions', 'showhelp', 'assignpublicip', 'subnettype', 'showsubnettype'],
     data: function () {
         return {
             groupStyle: this.showhelp ? 'input-group' : ''
--- a/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
+++ b/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
@@ -315,7 +315,7 @@ Vue.component("placements-select", {
     <label class="deployToolTip control-label col-xs-2" data-toggle="tooltip" v-bind:title="title">\
       {{label}}\
     </label>\
-    <div class="col-xs-7"><div v-bind:class="groupStyle">\
+    <div class="col-xs-6"><div v-bind:class="groupStyle">\
           <select class="form-control chosen-select"  v-on:change="updateValue($event.target.value)" multiple>\
             <option v-for="option in selectoptions" v-bind:value="option.value" v-bind:selected="option.isSelected">{{option.text}}</option>\
           </select>\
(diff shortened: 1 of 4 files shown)


robbintt:
This is looking great! I am curious what happens if a user is currently using a mix of private and public subnets. If they try to load the form, what are the outcomes of the following cases:
* will the form show all selected subnets?
* if the user saves the page, will public and private subnets still show as selected and save?
* will there be any disruption for any users if they deselect and reselect (it seems like they will just need to pick a new subnet and rotate old hosts
* do the users need to audit their current clusters and remove any hosts in the wrong subnet category? (private clusters with public subnets or vice versa)?

tylerwowen:
> will the form show all selected subnets

Yes. The filtering only applies to presentation. The selected subnets are loaded from cluster configs. 
> if the user saves the page, will public and private subnets still show as selected and save?

I don't quite follow this question. 
> will there be any disruption for any users if they deselect and reselect (it seems like they will just need to pick a new subnet and rotate old hosts

If a user doesn't change the subnet type, then there is no disruption. If a user changes the subnet type, the placements field is cleared and the user has to reselect from the corresponding list. However if the user change the subnet type back to the original value, the currently selected subnets will be restored and unsaved changes before changing the subnet type will not be preserved. 
> do the users need to audit their current clusters and remove any hosts in the wrong subnet category? (private clusters with public subnets or vice versa)?

I'd recommend all users perform an audit. Or we can query the database to find out and inform them. 
> Any other concerns or existing users or form selection?

Since a cluster can be either public or private, if a user checks "assign public IP" when rotating a subnet, the cluster will be updated to either public or private, regardless whether the rest subnets are private or public. This leads to undefined behaviors and thus why I created this weak bond between public subnet and assign public IP by setting the default values. This way users have the minimal possibility to choose the wrong cluster x subnet combination.

deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
```diff
@@ -315,21 +315,29 @@ Vue.component("placements-select", {
...
-   </span></div>\
+    </span></div>\
     </div>\
-    <div class="col-xs-3">\
-      <input type="checkbox" id="checkbox" v-bind:checked="assignpublicip" v-on:click="assignipchange($event.target.checked)">\
-      <label for="checkbox">Assign Public IP</label>\
-    <div></div>',
-    props: ['label', 'title', 'selectoptions', 'showhelp', 'assignpublicip'],
+    <div class="col-xs-2" v-if="showsubnettype">\
+        <input type="radio" id="public" value="public" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
```
robbintt, line 329: nit: ID is unique on the page right? It might be too general if you just call it "public" and "private".

deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
```diff
@@ -315,21 +315,29 @@ Vue.component("placements-select", {
...
+        <label for="public">Public subnets</label>\
+        <br>\
+        <input type="radio" id="private" value="private" v-model="subnettype" v-on:click="filterclick($event.target.value)">\
```
robbintt, line 332: nit: ID is unique on the page right? It might be too general if you just call it "public" and "private".

deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
```diff
@@ -315,21 +315,29 @@ Vue.component("placements-select", {
...
+        <label for="private">Private subnets</label>\
+        <br>\
+    </div>\
+    <div class="col-xs-2">\
+        <input type="checkbox" id="checkbox" v-bind:checked="assignpublicip" v-on:click="assignipchange($event.target.checked)">\
```
robbintt, line 337: nit: id is unique on the page right? It might be too general if you just call it "checkbox". (looks like this one is just a formatting change

robbintt (approved):
Approved, some small changes needed.

deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
tylerwowen: Yes, so I'm gonna leave it as-is.

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -840,13 +842,23 @@ <h4 class="panel-title pull-left pointer-cursor">
                 },
                 selectpublicip: function(value){
                     this.assignPublicIP = value
+                    if (this.assignPublicIP && this.subnetType == 'private' ||
+                        !this.assignPublicIP && this.subnetType == 'public') {
+                        globalNotificationBanner.error =
+                            `Warning: You ${this.assignPublicIP ? 'selected' :'unselected'} 'Assign Public IP' for ${this.subnetType} subnet type. ` +
+                            `Please double check if this is desiered. If you DO require this configuration, please report your scenario to the CDP team.`
```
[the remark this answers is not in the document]
tylerwowen: Updated.

ankilosaurus:
what's the reasoning behind providing 2.iv ?

ankilosaurus:
when user toggles from public to private and vice-e-versa, do we populate warning? Is it clear to user that existing subnets will be removed? 
what's the user experience when user saves subnet configs and then toggles public<->private?

note that, this introduces different behavior and could be prone to user mistakes.

ankilosaurus:
error message can be more generic for Teletraan users.

tylerwowen:
> what's the reasoning behind providing 2.iv ?

Because a public subnet without public IP can cause many problems. I'm also thinking of collecting user scenarios to better understand the usage. 
> what's the user experience when user saves subnet configs and then toggles public<->private?
 
If a user changes the subnet type, the placements field is cleared and the user has to reselect from the corresponding list. However there is a bug that the saved subnets are not preserved if it is toggled back to the original setting (private -> public -> private).


> when user toggles from public to private and vice-e-versa, do we populate a warning? Is it clear to user that existing subnets will be removed?

No we don't show a warning if a user simply changes the subnet type. The whole field is cleared so I think it's quite clear that existing subnets will be removed.

robbintt:
Would it be easy to clear the warning message when the right setting is returned to? This would give the user clear guidance that they are going back to a "expected state".

ankilosaurus:
warning looks good. if you toggle back, will it restore config?

tylerwowen:
> warning looks good. if you toggle back, will it restore config?

Yes

ankilosaurus (approved):

PR #952, opened 2022-03-21
Generate new Nimbus id when cloning cluster

osoriano:
The "Capacity" view in the Teletraan stage has a "clone cluster" button
to clone the stage and cluster.

However, it also clones the nimbus uuid which gets the teletraan stage/nimbus id out of sync.

Instead, generate a new Nimbus uuid for the cloned cluster

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -635,6 +635,7 @@ def delete_cluster(request, name, stage):
 
 
 def clone_cluster(request, src_name, src_stage):
+    external_id = None
     try:
         params = request.POST
         dest_name = params.get('new_environment', src_name)
@@ -646,7 +647,7 @@ def clone_cluster(request, src_name, src_stage):
         ##0. teletraan service get src env buildName
         src_env = environs_helper.get_env_by_stage(request, src_name, src_stage)
         build_name = src_env.get('buildName', None)
-        external_id = src_env.get('externalId', None)
+        external_id = environs_helper.create_identifier_for_new_stage(request, dest_name, dest_stage)
 
         ##1. teletraan service create a new env
         dest_env = environs_helper.create_env(request, {
@@ -709,10 +710,15 @@ def clone_cluster(request, src_name, src_stage):
         return HttpResponse(json.dumps(src_cluster_info), content_type="application/json")
     except NotAuthorizedException as e:
         log.error("Have an NotAuthorizedException error {}".format(e))
+        if external_id is not None:
+            environs_helper.delete_nimbus_identifier(request, external_id)
+
         return HttpResponse(e, status=403, content_type="application/json")
     except Exception as e:
         log.error("Failed to clone cluster env_name: %s, stage_name: %s" % (src_name, src_stage))
         log.error(traceback.format_exc())
+        if external_id is not None:
+            environs_helper.delete_nimbus_identifier(request, external_id)
         return HttpResponse(e, status=500, content_type="application/json")
 
 
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -818,38 +818,6 @@ class EnvNewDeployView(View):
         common.deploy(request, name, stage)
         return redirect('/env/%s/%s/deploy' % (name, stage))
 
-def create_identifier_for_new_stage(request, env_name, stage_name):
-    """ Create a Nimbus Identifier for the new stage. Assumes that the environment has at least one stage with externalId set.
-        This is needed so that the method knows which project to associate the new stage to.
-
-        If the environment has no stage with externalId set, this method will not attempt to create an Identifier.
-    """
-
-    # get all stages within this environment
-    all_env_stages = environs_helper.get_all_env_stages(request, env_name)
-    stage_with_external_id = None
-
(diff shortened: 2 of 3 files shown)


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -646,7 +646,9 @@ def clone_cluster(request, src_name, src_stage):
         ##0. teletraan service get src env buildName
         src_env = environs_helper.get_env_by_stage(request, src_name, src_stage)
         build_name = src_env.get('buildName', None)
-        external_id = src_env.get('externalId', None)
+        external_id = environs_helper.create_identifier_for_new_stage(request, dest_name, dest_stage)
+        if external_id is None:
+            raise TeletraanException("Unable to find Nimbus identifier for enviornment %s" % dest_name)
```
ankilosaurus, line 651: this is a different behavior from existing flow which is not failing on nimbus id creation.

Also, if stage creation in Teletraan fails, do we clean up generated nimbus id?

ankilosaurus: ^ can be wrapped in helper method to avoid repeated code.

osoriano: Thanks, updated
* Added `IS_PINTEREST` and `identifier.get('uuid')` check to helper method
* Added cleanup for nimbus id

deploy-board/deploy_board/webapp/helpers/environs_helper.py
```diff
@@ -65,6 +65,45 @@
...
 
 
+def create_identifier_for_new_stage(request, env_name, stage_name):
+    """ Create a Nimbus Identifier for the new stage. Assumes that the environment has at least one stage with externalId set.
+        This is needed so that the method knows which project to associate the new stage to.
+
+        If the environment has no stage with externalId set, this method will not attempt to create an Identifier.
+    """
+    # Only create identifier for Pinterest environments
+    if not IS_PINTEREST:
```
ankilosaurus, line 75: could do same for delete_nimbus_identifier. 

This methods could be part of interface to avoid `if IS_PINTEREST` everywhere but not in the scope of this PR.

ankilosaurus (approved):

deploy-board/deploy_board/webapp/helpers/environs_helper.py
osoriano: Moved `IS_PINTEREST` into `delete_nimbus_identifier`.

yeah would be nice to avoid `IS_PINTEREST`

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -709,10 +710,15 @@ def clone_cluster(request, src_name, src_stage):
         return HttpResponse(json.dumps(src_cluster_info), content_type="application/json")
     except NotAuthorizedException as e:
         log.error("Have an NotAuthorizedException error {}".format(e))
+        if external_id is not None:
+            environs_helper.delete_nimbus_identifier(request, external_id)
```
robbintt, line 714: If there is any bad data, will this delete the nimbus identifier causing an incident for old stages?

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -709,10 +710,15 @@ def clone_cluster(request, src_name, src_stage):
...
+
         return HttpResponse(e, status=403, content_type="application/json")
     except Exception as e:
         log.error("Failed to clone cluster env_name: %s, stage_name: %s" % (src_name, src_stage))
         log.error(traceback.format_exc())
+        if external_id is not None:
+            environs_helper.delete_nimbus_identifier(request, external_id)
```
robbintt, line 721: If there is any bad data, will this delete the nimbus identifier causing an incident for old stages?

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -862,20 +830,16 @@ def post_add_stage(request, name):
...
 
     try:
         if from_stage:
             common.clone_from_stage_name(request, name, stage, name, from_stage, description, external_id)
         else:
             common.create_simple_stage(request,name, stage, description, external_id)
     except:
-        if IS_PINTEREST:
-            environs_helper.delete_nimbus_identifier(request, external_id)
+        environs_helper.delete_nimbus_identifier(request, external_id)
```
robbintt, line 842: If there is any bad data, will this delete the nimbus identifier causing an incident for old stages?

robbintt (changes requested):
I think this could cause an incident, added some comments inline about deleting nimbus identifiers in the except blocks.

deploy-board/deploy_board/webapp/cluster_view.py
robbintt: looks like external ID can only be set if new, so i think this one is OK

deploy-board/deploy_board/webapp/cluster_view.py
robbintt: looks like external ID can only be set if new, so i think this one is OK

deploy-board/deploy_board/webapp/env_views.py
robbintt: looks like external ID can only be set if new, so i think this one is OK

robbintt (approved):
All cases look good.

PR #954, opened 2022-03-23
register keys

aagxxi:
Adding register of keys to knox

### DIFF
--- a/deploy-sentinel/teletraan/RESTARTING
+++ b/deploy-sentinel/teletraan/RESTARTING
@@ -4,5 +4,9 @@ set -e
 
 echo Running RESTARTING script.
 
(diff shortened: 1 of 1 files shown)


deploy-sentinel/teletraan/RESTARTING
```diff
@@ -4,5 +4,9 @@ set -e
 
 echo Running RESTARTING script.
 
+# Register knox keys
+knox register -k teletraan:prod:rodimusapitoken
```
krockpot, line 8: `knox register -g -k <key>`  will also fetch it, `-k` only adds it to the local config file (fetches every 10 minutes)

krockpot (approved):

PR #958, opened 2022-04-02
Structured logging

kynanlalone:
Logging overhaul.  Changes standard logs to JSON logs and adds visibility for requests handled, requests made, and templates.

* add JSON logging helper
* fix request handler(s),
* add template error logging

relevant python documentation:
https://docs.python.org/3/library/logging.html#logging.Formatter
https://docs.python.org/3/howto/logging-cookbook.html#customized-exception-formatting

### DIFF
--- a/deploy-board/deploy_board/webapp/helpers/settings_logging.py
+++ b/deploy-board/deploy_board/webapp/helpers/settings_logging.py
@@ -1,0 +1,37 @@
+import json
+import logging
+
+
+class StructuredMessage(logging.Formatter):
+    def formatException(self, exc_info):
+        """
+        format an exception so that it prints on a single line.
+        return: str
+        """
+        result = super(StructuredMessage, self).formatException(exc_info)
+        return repr(result)
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/webapp/helpers/settings_logging.py
```diff
@@ -0,0 +1,37 @@
+import json
+import logging
+
+
+class StructuredMessage(logging.Formatter):
```
robbintt, line 5: It might be possible to instantiate a formatter without inheritance. Do you want to meet up and discuss soon? https://docs.python.org/3/library/logging.html#logging.Formatter

robbintt:
I think I traced how you got here looking through some online documentation and python cookbook, can you put more discussion in the PR description so the approach has the correct context? I think the use of subclass and super here has a code smell, when using some formatting strings would make more sense, but it's recommended in the Python logging cookbook, so we can go with it.

robbintt (approved):

robbintt (approved):

PR #962, opened 2022-04-11
rodimus token from knox - recommended way

aagxxi:
knox key - rodimus token from knox, recommended way


aagxxi:
superseded by new PR

PR #980, opened 2022-06-02
Fetch acceptance data to populate base image Acceptance column

tylerwowen:
`GET /base_images/` does not return acceptance information. Added separate calls to Rodimus to get the acceptance data.

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -296,12 +296,12 @@ def create_base_image(request):
 def get_base_images(request):
     index = int(request.GET.get('page_index', '1'))
     size = int(request.GET.get('page_size', DEFAULT_PAGE_SIZE))
-    base_images = baseimages_helper.get_all(request, index, size)
+    base_images = baseimages_helper.get_all_with_acceptance(request, index, size)
     provider_list = baseimages_helper.get_all_providers(request)
     cells_list = cells_helper.get_by_provider(request, DEFAULT_PROVIDER)
 
     return render(request, 'clusters/base_images.html', {
-        'base_images': add_acceptance(request, base_images),
+        'base_images': base_images,
         'provider_list': provider_list,
         'cells_list': cells_list,
         'pageIndex': index,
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -311,6 +311,23 @@ def get_base_images(request):
     })
 
 
+def add_acceptance(request, base_images):
```
ankilosaurus, line 314: this looks like a helper method and should not be visible outside of class?
Also, might want to rename it to reflect it's not updating any record ex: populate_acceptance_status

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -311,6 +311,23 @@ def get_base_images(request):
...
+    fetched_names = set()
+    name_acceptance_map = {}
+    for img in base_images:
+        name = img['abstract_name']
```
ankilosaurus, line 318: is abstract_name unique? I think it's user entered field and does not have uniqueness contraint. In that case, can we populate status based on unique key ex: provider_name?

deploy-board/deploy_board/webapp/cluster_view.py
ankilosaurus: actually I think this can be part of baseimages_helper

deploy-board/deploy_board/webapp/cluster_view.py
tylerwowen: `get_acceptance_by_name` takes abstract_name and returns all images with that abstrac_name. That's why I used the dict `name_acceptance_map` to store the mapping from provider_name to acceptance.

deploy-board/deploy_board/webapp/cluster_view.py
tylerwowen: Good idea, moved.

deploy-board/deploy_board/webapp/helpers/baseimages_helper.py
```diff
@@ -28,6 +28,24 @@ def get_all(request, index, size):
     return rodimus_client.get("/base_images", request.teletraan_user_id.token, params=params)
 
 
+def get_all_with_acceptance(request, index, size):
+    base_images = get_all(request, index, size)
+    fetched_names = set()
+    name_acceptance_map = {}
+    for img in base_images:
+        name = img['abstract_name']
```
ankilosaurus, line 36: Is abstract_name unique ?

tylerwowen: It isn't.

ankilosaurus (approved):

PR #983, opened 2022-06-04
When an existing ASG is disabled, still launch is ASG by increasing the capacity.

tylerwowen:
After users disabled auto scaling, they still have the need to increase the capacity manually via “Launch Instance”. Therefore Teletraan should increase the ASG size like the ASG is still enabled. 

When clusters have associated ASGs, the only way to launch outside the ASG is to specify a subnet and a placement group.

### DIFF
--- a/deploy-board/deploy_board/webapp/group_view.py
+++ b/deploy-board/deploy_board/webapp/group_view.py
@@ -1046,7 +1046,7 @@ def add_instance(request, group_name):
                 use_placement_group = True
 
         # When a group has an associated ASG, its status is either ENABLED or DISABLED.
-        # We should always launch in ASG when there is an ASG setup regardless its status
+        # We should always launch in ASG when an ASG is set up regardless its status,
         # unless placement group is specified
         if (asg_status == "ENABLED" or asg_status == "DISABLED") and not use_placement_group:
             launch_in_asg = True
--- a/deploy-board/deploy_board/webapp/group_view.py
+++ b/deploy-board/deploy_board/webapp/group_view.py
@@ -1032,7 +1032,6 @@ def add_instance(request, group_name):
     subnet = None
     placement_group = None
     asg_status = str(params['asgStatus']).upper()
-    launch_in_asg = False
     use_placement_group = False
 
     try:
@@ -1049,9 +1048,6 @@ def add_instance(request, group_name):
         # We should always launch in ASG when an ASG is set up regardless its status,
         # unless placement group is specified.
         if (asg_status == 'ENABLED' or asg_status == 'DISABLED') and not use_placement_group:
-            launch_in_asg = True
-
-        if launch_in_asg:
             # Launch hosts inside ASG / Bump ASG size by required instances
             autoscaling_groups_helper.launch_hosts(request, group_name, num, None)
             content = 'Capacity increased by {} for Auto Scaling Group {}. Please go to ' \
--- a/deploy-board/deploy_board/webapp/group_view.py
+++ b/deploy-board/deploy_board/webapp/group_view.py
@@ -194,7 +194,7 @@ def update_launch_config(request, group_name):
         launch_config = group_info.get("launchInfo")
         if 'userData' in launch_config:
             actual_spiffe_id = get_spiffe_id(launch_config['userData'].splitlines())
-        
+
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -1045,16 +1045,20 @@ def add_instance(request, group_name):
...
-        if str(asg_status).upper() == "ENABLED" and not use_placement_group:
+        # When a group has an associated ASG, its status is either ENABLED or DISABLED.
+        # We should always launch in ASG when an ASG is set up regardless its status,
+        # unless placement group is specified.
+        if (asg_status == 'ENABLED' or asg_status == 'DISABLED') and not use_placement_group:
             launch_in_asg = True
 
         if launch_in_asg:
             # Launch hosts inside ASG / Bump ASG size by required instances
             autoscaling_groups_helper.launch_hosts(request, group_name, num, None)
```
robbintt, line 1056: Below there is a message in the condition hosts are launched in the ASG, maybe the message should also apply here -- if so it could be moved out of the conditional logic and use host_ids as the test, or a flag: `content = '{} hosts have been launched to group {} (host ids: {})'.format(num, group_name, host_ids)` currently line 1081 in this diff.

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -1045,16 +1045,20 @@ def add_instance(request, group_name):
...
 
-        # Check if asg is enabled and does not use placement group. Then launch in asg
-        if str(asg_status).upper() == "ENABLED" and not use_placement_group:
+        # When a group has an associated ASG, its status is either ENABLED or DISABLED.
+        # We should always launch in ASG when an ASG is set up regardless its status,
+        # unless placement group is specified.
+        if (asg_status == 'ENABLED' or asg_status == 'DISABLED') and not use_placement_group:
             launch_in_asg = True
 
         if launch_in_asg:
```
ankilosaurus, line 1054: looks like we don't need extra launch_in_asg variable. lines 1051-1054 can be simplified with single if condition.

ankilosaurus: also, do we still need to check for asg_status given it's launching instance for either ENABLED or DISABLED ?

can be just "if not use_placement_group:" ?

robbintt (commented):
I noticed there is also an env_view route to another `add_instance` implementation. It might be out of scope or it might be better to also disable this method:
* https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/webapp/urls.py#L188
* https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/webapp/env_views.py#L1623
* https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/webapp/helpers/autoscaling_groups_helper.py#L53

It looks like `/<env>/<stage>/addinstance` may not be referenced, but it still seems functional: https://github.com/pinterest/teletraan/search?q=addinstance

deploy-board/deploy_board/webapp/group_view.py
tylerwowen: Unfortunately, `autoscaling_groups_helper.launch_hosts` doesn't return anything.

robbintt (dismissed):

deploy-board/deploy_board/webapp/group_view.py
tylerwowen: > can be just "if not use_placement_group:" ?
No, because if `asg_status == 'UNKNOWN'` and `use_placement_group == False`, the current condition won't be met, but condition`if not use_placement_group` would be metand the code would increase the capacity of a non-exist ASG.

tylerwowen: Removed `launch_in_asg`.

tylerwowen:
> hi @tylerwowen thanks for resolving comments. diff looks good. one question: what will happen when ASG at MAX capacity and someone tries to launch host ? Will it surface error to user ? Might want to add list of cases you have covered.

@ankilosaurus yes, any error message will be sent back to the user. #982 ensures that. All of the `messages.add_message()` calls in this function are setting the user-facing messages.

ankilosaurus (approved):

PR #987, opened 2022-06-10
Added inline form warning when users try to over-resize ASGs

tylerwowen:
Added inline form warning when users try to over-resize ASGs. Specifically, 
1. Unified env view and group view min/max size UI components by extending `asg-capacity-config`. 
2. Updated `asg-capacity-config` and `static-capacity-config` to support displaying inline form warning and error.
3. Extended `EnvCapacityConfigView.get()` and `get_asg_config()` to return placements so the frontend can check remaining capacity.

### Warning shown when max size increase exceeds the remaining capacity
resize-warning

group_resieze_warning

### Warning shown when min size increase exceeds the remaining capacity
resize-error
group_resize_error

### Warning shown when potential AZ imbalance is detected.
imbalance

### DIFF
--- a/deploy-board/deploy_board/static/js/components/capacitycomponents.js
+++ b/deploy-board/deploy_board/static/js/components/capacitycomponents.js
@@ -205,6 +205,20 @@ function getCapacityAlertMessage(isWarning, remainingCapacity, placements, incre
     }
 }
 
+function checkImbalance(placements, threshold) {
+    var insufficientPlacements = [];
+    var showImbalanceWarning = false;
+    for (var p of placements) {
+        if (p.capacity < threshold) {
+            showImbalanceWarning = true;
+            insufficientPlacements.push(p)
+        }
+    }
+    return showImbalanceWarning ? `The ASG might have AZ imbalance based on a potential increase of ${threshold} hosts per placement. `+
+        `Please pay attention to the following placements:\n` +
+        `${JSON.stringify(insufficientPlacements, ['capacity', 'provider_name', 'abstract_name'], 2)}` : '';
+}
+
 /**
  * The capacity button. This is shown when autoscaling is disabled or min\max size are the same.
  * In this case, only one capacity box is shown
@@ -220,7 +234,8 @@ Vue.component("static-capacity-config", {
                 :value="capacity" @input="onCapacityChange($event.target.value)" @keydown.enter.prevent="">
         </div>
     </div>
-    <form-error v-show="showSizeError" :alert-text="sizeError"></form-error>
+    <form-danger v-show="showSizeError" :alert-text="sizeError"></form-danger>
+    <form-warning v-show="showImbalanceWarning" :alert-text="imbalanceWarning"></form-warning>
     </div>`,
     props: {
         originalCapacity: Number,
@@ -231,7 +246,9 @@ Vue.component("static-capacity-config", {
         return {
             capacity: this.originalCapacity,
             showSizeError: false,
+            showImbalanceWarning: false,
             sizeError: '',
+            imbalanceWarning: '',
         }
     },
     methods: {
@@ -248,6 +265,8 @@ Vue.component("static-capacity-config", {
             } else {
                 this.showSizeError = false;
             }
+            this.imbalanceWarning = checkImbalance(this.placements, (sizeIncrease / this.placements.length).toFixed());
+            this.showImbalanceWarning = this.imbalanceWarning != '';
         }
     }
 });
@@ -277,8 +296,9 @@ Vue.component("asg-capacity-config", {
             </div>
         </div>
     </div>
+    <form-danger v-show="showSizeError" :alert-text="sizeError"></form-danger>
     <form-warning v-show="showSizeWarning" :alert-text="sizeWarning"></form-warning>
-    <form-error v-show="showSizeError" :alert-text="sizeError"></form-error>
+    <form-warning v-show="showImbalanceWarning" :alert-text="imbalanceWarning"></form-warning>
     </div>`,
     props: {
         labelBootstrapClass: {
@@ -297,6 +317,7 @@ Vue.component("asg-capacity-config", {
             type: String,
             default: 'Capacity'
         },
+        currentSize: Number,
         originalMinSize: Number,
         originalMaxSize: Number,
         remainingCapacity: Number,
@@ -308,8 +329,10 @@ Vue.component("asg-capacity-config", {
             maxSize: this.originalMaxSize,
             showSizeError: false,
             showSizeWarning: false,
+            showImbalanceWarning: false,
             sizeError: '',
             sizeWarning: '',
+            imbalanceWarning: '',
         }
     },
     methods: {
@@ -340,12 +363,19 @@ Vue.component("asg-capacity-config", {
                 this.showSizeError = false;
             }
 
-            if(!this.showSizeError && maxIncrease >= this.remainingCapacity){
+            if (!this.showSizeError && maxIncrease >= this.remainingCapacity) {
                 this.sizeWarning = getCapacityAlertMessage(true, this.remainingCapacity, this.placements, maxIncrease);
                 this.showSizeWarning = true;
             } else {
                 this.showSizeWarning = false;
             }
+
+            const avgSizeIncreasePerPlacement = ((this.maxSize - this.currentSize) / this.placements.length).toFixed();
+            this.imbalanceWarning = checkImbalance(this.placements, avgSizeIncreasePerPlacement);
+            this.showImbalanceWarning = this.imbalanceWarning != '';
         },
+    },
+    attached: function() {
+        this.validateSize();
     }
 });
--- a/deploy-board/deploy_board/static/js/components/capacitycomponents.js
+++ b/deploy-board/deploy_board/static/js/components/capacitycomponents.js
@@ -188,23 +188,67 @@ Vue.component('in-rolling-alert', {
     props:['actionlink']
 });
 
+function getCapacityAlertMessage(isWarning, remainingCapacity, placements, increase) {
+    const errorMessage = `Insufficient combined remaining capacity in this cluster/auto scaling group. `;
+    const instruction = `You can attach additional placements to the corresponding clutter to increase` +
+                        ` total potential capacity at Cluster Configuration -> Advanced Settings.\n`;
+    const status = `Combined remaining capacity: ${remainingCapacity}\n` +
+                   `Current placement(s): ${JSON.stringify(placements, null, 2)}`;
+
+    if (isWarning) {
+        return errorMessage + `You shouldn't save this configuration because the cluster/ASG will run into capacity issues. ` +
+                instruction + `Requested size increase: ${increase}\n` + status
+    } else {
+        return errorMessage + `You can save this configuration but the cluster/ASG might run into capacity issues in the future. ` +
+                instruction + `Requested size increase: ${increase}\n` + status;
+    }
+}
 
 /**
  * The capacity button. This is shown when autoscaling is disabled or min\max size are the same.
  * In this case, only one capacity box is shown
  */
 Vue.component("static-capacity-config", {
-    template: '<div class="form-group">\
-    <label for="capacity" class="deployToolTip control-label col-xs-4" title="Number of hosts for this service">\
-        Capacity\
-    </label>\
-    <div class="col-xs-2" >\
-    <input class="form-control" v-bind:value="capacity" v-on:change="updateValue($event.target.value)" @keydown.enter.prevent="" type="number" min="0"></input>\
-    </div></div>',
-    props: ['capacity'],
+    template: `<div>
+    <div class="form-group">
+        <label for="capacity" class="deployToolTip control-label col-xs-4" title="Number of hosts for this service">
+            Capacity
+        </label>
+        <div class="col-xs-2" >
+            <input name='"capacity" class="form-control" type="number" min="0" required
+                v-model="capacity" @keydown.enter.prevent="">
+        </div>
+    </div>
+    <form-error v-show="showSizeError" :alert-text="sizeError"></form-error>
+    </div>`,
+    props: {
+        originalCapacity: 0,
+        remainingCapacity: Infinity,
+        placements: {},
+    },
+    data: function() {
+        return {
+            capacity: this.originalCapacity,
+            showSizeError: false,
+            sizeError: '',
+        }
+    },
+    watch: {
+        capacity: function (value) {
+            this.capacity = Number(capacity);
+            this.validateSize();
+            this.$emit('change', value);
+        },
+    },
     methods: {
-        updateValue: function (value) {
-            this.$emit("change", value)
+        validateSize: function () {
+            const sizeIncrease = this.capacity - this.originalCapacity;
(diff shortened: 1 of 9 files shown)


deploy-board/deploy_board/templates/clusters/cluster_capacity.tmpl
```diff
@@ -55,43 +63,28 @@
...
                     if (value){
                         this.save();
                     }
                 },
                 save:function(){
                     if (!this.autoscalingEnabled){
-                        if (this.desiredCapacity<0){
-                            globalNotificationBanner.error = "Capacity cannot be less than 0"
-                            return
-                        }
```
tylerwowen, line 86: Note the input box of the form now has `min="0"` to ensure this property.

deploy-board/deploy_board/templates/clusters/cluster_capacity.tmpl
```diff
@@ -55,45 +63,25 @@
...
                         //Set min and max size directly to group. Right now, cluster itself has no autoscaling info
                         this.minsize = this.desiredCapacity
                         this.maxsize = this.desiredCapacity
                     }
 
-                    if (this.minsize<0 || this.maxsize<0 || this.minsize > this.maxsize){
-                            globalNotificationBanner.error = "Min and Max size cannot be less than 0 and Max size cannot be less than Min size"
```
tylerwowen, line 93: This is ensured by the vue component, no need to check again here.

deploy-board/deploy_board/static/js/components/capacitycomponents.js
```diff
@@ -214,30 +257,95 @@ Vue.component("static-capacity-config", {
...
+        },
+        onMaxSizeChange: function (value) {
+            this.maxSize = Number(value);
+            if (this.maxSize < this.minSize) {
+                this.minSize = this.maxSize;
+            }
+            this.validateSize();
+            this.$emit('max-change', this.maxSize);
+        },
+        validateSize: function() {
```
ankilosaurus, line 332: Can we simplify this logic? If we are not preventing users from making this modification, we can just warn for these cases.

tylerwowen: I'm not sure if I follow. Here it is trying to warn users inside `validateSize`.

ankilosaurus:
I think it's nice to show capacity for each subnet in the warning message. However, subnet <-> capacity is sufficient and useful. Should avoid showing other fields specially generated ones.

deploy-board/deploy_board/webapp/capacity_views.py
```diff
@@ -92,22 +99,22 @@ def get(self, request, name, stage):
...
         data['info'] = json.dumps(data)
         return render(request, 'configs/capacity.html', data)
 
     def post(self, request, name, stage):
         logger.info("Post to capacity with data {0}".format(request.body))
         data = json.loads(request.body)
-        hosts =  data.get('hosts')
-        if hosts != None:
+        hosts = data.get('hosts')
+        if hosts is not None:
```
robbintt, line 112: you can use `if hosts:` here

robbintt: (probably)

deploy-board/deploy_board/webapp/capacity_views.py
```diff
@@ -92,22 +99,22 @@ def get(self, request, name, stage):
...
             environs_helper.update_env_capacity(request, name, stage, capacity_type="HOST", data=hosts)
 
         groups = data.get("groups")
-        if groups != None:
+        if groups is not None:
```
robbintt, line 116: same as above: `if groups:`

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -24,10 +24,9 @@
 import logging
 import traceback
 
-from helpers import environs_helper, clusters_helper
-from helpers import groups_helper, baseimages_helper
-from helpers import specs_helper, autoscaling_groups_helper
-from helpers import autoscaling_metrics_helper
+from helpers import environs_helper, clusters_helper, groups_helper,\
```
robbintt, line 27: nit: prefer black style or just leave as is: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#parentheses

robbintt: we should probably choose a standard as a project 😃

deploy-board/deploy_board/webapp/helpers/placements_helper.py
```diff
@@ -36,3 +36,13 @@ def get_by_provider_and_cell_name(request, provider, cell_name):
 
 def get_by_id(request, placement_id):
     return rodimus_client.get("/placements/%s" % placement_id, request.teletraan_user_id.token)
+
+
+def get_simplified_by_ids(request, placement_str, provider, cell):
+    current_placement_ids = set(placement_str.split(','))
+    all_placements = get_by_provider_and_cell_name(request, provider, cell)
+    current_placements = filter(lambda p: p['id'] in current_placement_ids, all_placements)
```
robbintt, line 44: nit: list comprehension is probably more idiomatic and easy to read unless you are running into memory constraints where iterables shine (not this case 😄 )

* [item for item in iterable if function(item)]
* https://python-reference.readthedocs.io/en/latest/docs/functions/filter.html

deploy-board/deploy_board/webapp/helpers/placements_helper.py
```diff
@@ -36,3 +36,13 @@ def get_by_provider_and_cell_name(request, provider, cell_name):
...
+    return map(lambda p: {'id': p['id'],
```
robbintt, line 45: I think this is fine, but you might be better served by a list comprehension here that pares extra keys from the original list of dicts, here is a similar example without list comprehension: https://stackoverflow.com/a/12387627

robbintt: you would basically provide the desired keys as a list and make the check `if k not in [a, b, c, d]: del d[k]`

deploy-board/deploy_board/webapp/helpers/placements_helper.py
```diff
@@ -36,3 +36,13 @@ def get_by_provider_and_cell_name(request, provider, cell_name):
...
+                          'capacity': p['capacity'],
+                          'abstract_name': p['abstract_name'],
```
robbintt, line 47: might prefer `p.get(k, '')` since this is just for the user. then missing keys will be populated ''

deploy-board/deploy_board/templates/groups/asg_config.tmpl
```diff
@@ -105,6 +81,19 @@ $(function() {
...
 
+var asgConfigVue = new Vue({
+    el: '#autoscalingConfigId',
+    data: {
+        {% if placements %}
+        placements: {{ placements | safe }},
+        remainingCapacity: {{placements| safe }}.reduce((s, e) => s + e.capacity, 0),
+        {% else %}
+        placements: {},
+        remainingCapacity: Infinity,
```
robbintt, line 92: probably `n/a` rather than infinity

tylerwowen: `n/a` won't be evaluated properly in this expression `if (sizeIncrease >= this.remainingCapacity) {`

tylerwowen:
> I think it's nice to show capacity for each subnet in the warning message. However, subnet <-> capacity is sufficient and useful. Should avoid showing other fields specially generated ones.

> removed `id`, kept `provider_name` so users know which subnet it is, and `abstract_name` which suggests AZ. The AZ information can be later used to find additional subnets to attach to the cluster.

deploy-board/deploy_board/webapp/helpers/placements_helper.py
```diff
@@ -36,3 +36,10 @@ def get_by_provider_and_cell_name(request, provider, cell_name):
 
 def get_by_id(request, placement_id):
     return rodimus_client.get("/placements/%s" % placement_id, request.teletraan_user_id.token)
+
+
+def get_simplified_by_ids(request, placement_str, provider, cell):
+    current_placement_ids = set(placement_str.split(','))
+    all_placements = get_by_provider_and_cell_name(request, provider, cell)
+    return [{k: placement[k] for k in placement if k in ['id', 'capacity', 'abstract_name', 'provider_name']}
```
robbintt, line 44: nit - might still prefer `.get(k)` for user facing in case a key is missing. it seems a key should never be missing though...

deploy-board/deploy_board/static/js/components/capacitycomponents.js
```diff
@@ -214,30 +257,95 @@ Vue.component("static-capacity-config", {
...
-                    <span class="input-group-addon">Max Size</span>\
-                    <input class="form-control" v-bind:value="maxsize" v-on:change="updateMaxSize($event.target.value)" @keydown.enter.prevent=""></input>\
-                </div>\
-            </div>\
-    </div>',
-    props: ['minsize', 'maxsize'],
-    methods: {
-        updateMinSize: function (value) {
-            this.$emit('minchange', value)
+    template: `<div>
```
robbintt, line 260: is it an extra div? doesn't seem used, maybe consumed somewhere else?

deploy-board/deploy_board/webapp/helpers/placements_helper.py
tylerwowen: Line 44 `for k in placement` so k must exist.
Line 45 `placement['id']``id` is a DB id so it must exist.

deploy-board/deploy_board/static/js/components/capacitycomponents.js
tylerwowen: No, this groups
```html
 <form-warning v-show="showSizeWarning" :alert-text="sizeWarning"></form-warning>
 <form-error v-show="showSizeError" :alert-text="sizeError"></form-error>
```
Otherwise weird layout shows.

deploy-board/deploy_board/static/js/components/capacitycomponents.js
```diff
@@ -188,23 +188,66 @@ Vue.component('in-rolling-alert', {
...
 Vue.component("static-capacity-config", {
-    template: '<div class="form-group">\
-    <label for="capacity" class="deployToolTip control-label col-xs-4" title="Number of hosts for this service">\
-        Capacity\
-    </label>\
-    <div class="col-xs-2" >\
-    <input class="form-control" v-bind:value="capacity" v-on:change="updateValue($event.target.value)" @keydown.enter.prevent=""></input>\
-    </div></div>',
-    props: ['capacity'],
+    template: `<div>
```
robbintt, line 213: is this an extra div? doesn't seem used, not sure if it helps rendering somewhere else? (same message as below)

deploy-board/deploy_board/static/js/components/capacitycomponents.js
ankilosaurus: I meant we can have generic warning message when capacity is changed such that ASG could have imbalance to simplify logic that's in generateCapacityAlertMsg. 

Also, ASG could run into capacity issue without this config change. So, I think checking this condition on capacity page load would be great.

deploy-board/deploy_board/static/js/components/capacitycomponents.js
```diff
@@ -214,30 +257,95 @@ Vue.component("static-capacity-config", {
...
+            const maxIncrease = this.maxSize - this.originalMaxSize;
+
+            if (minIncrease >= this.remainingCapacity) {
+                this.sizeError = getCapacityAlertMessage(false, this.remainingCapacity, this.placements, minIncrease);
+                this.showSizeError = true;
+            } else {
+                this.showSizeError = false;
+            }
+
+            if(!this.showSizeError && maxIncrease >= this.remainingCapacity){
```
robbintt, line 343: can't both conditions need warnings? it seems they are independent, e.g. if both are 1 million, we should display both messages

robbintt: Or do both have the same display but different severity? I will check

deploy-board/deploy_board/static/js/components/capacitycomponents.js
tylerwowen: No, it groups up `form-error `

deploy-board/deploy_board/static/js/components/capacitycomponents.js
tylerwowen: The idea is to show the worse problem. When the scenario you mentioned happens, an error shows first. When the error (min size) is corrected, the warning shows util the max size is corrected.

robbintt (approved):

deploy-board/deploy_board/static/js/components/capacitycomponents.js
ankilosaurus: also, it seems it would catch total capacity issues in ASG and not the AZ imbalance ? Is that the intention?

deploy-board/deploy_board/static/js/components/capacitycomponents.js
robbintt: it would be better to show all problems upfront if possible, otherwise the user is waiting on one error to see the next error, sort of like when we have a compile failure, then find more compile failures are waiting after. we can avoid that by showing all problems

deploy-board/deploy_board/static/js/components/capacitycomponents.js
tylerwowen: That is the intention. Regarding AZ imbalance, I think we can check `if (maxSize - current size) / numPlacements >= any remaining capacity`. However this will inevitably result in a even complicated logic.

deploy-board/deploy_board/static/js/components/capacitycomponents.js
tylerwowen: This has some subtle differences. A user has to over-resize once, sees and ignores the first alert and then over-resizes again to trigger the second alert (hiding if it's a warning or overriding if it's an error). I will however remove the check to show both.

deploy-board/deploy_board/static/js/components/capacitycomponents.js
tylerwowen: Done

deploy-board/deploy_board/static/js/components/capacitycomponents.js
```diff
@@ -248,6 +265,8 @@ Vue.component("static-capacity-config", {
             } else {
                 this.showSizeError = false;
             }
+            this.imbalanceWarning = checkImbalance(this.placements, (sizeIncrease / this.placements.length).toFixed());
```
robbintt, line 268: This seems good.

I am also wondering if we could/should sum the remaining capacity in the placements from the information available "remaining capacity" for each placement? 

On one hand, the capacity is available, OTOH it can become unbalanced. In this sense, maybe the increase should follow a different calculation.

What if instead we use min() of list of remaining capacity per AZ, so basically sum remaining capacity of subnets in each AZ. This could be a followup ticket or something, the PR is already pretty large.

robbintt: What if we just move this calculation into a function for now and ticket the improvement?

robbintt (approved):

PR #988, opened 2022-06-15
fix the github app token usage

yongwen:
as github app token will expires, we need to set the request header every time with the newly created github token.  this change support both permant personal token (from this.token) and the github app generated temporary token.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
@@ -85,15 +85,19 @@ public class GithubManager extends BaseManager {
                 GitHub gitHubApp = new GitHubBuilder().withJwtToken(jwtToken).build();
                 GHAppInstallation appInstallation = gitHubApp.getApp().getInstallationByOrganization(this.githubAppOrganization);
                 GHAppInstallationToken appInstallationToken = appInstallation.createToken().create();    
-                this.token = appInstallationToken.getToken();
+
+                // always use the newly created github app token as the token will expire
+                this.headers.put("Authorization", String.format("Token %s", appInstallationToken.getToken()));
             } catch (Exception e) {
                 // e.printStackTrace();
                 LOG.error("Exception when getting Github token: ", e);
                 throw e;
             }
         }
-
-        this.headers.put("Authorization", String.format("Token %s", this.token));
+        else {
+            // this is the token that does not expire
+            this.headers.put("Authorization", String.format("Token %s", this.token));
+        }
     }
 
     private String getSha(Map<String, Object> jsonMap) {


deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
```diff
@@ -85,15 +85,19 @@ private void setHeaders() throws Exception {
...
+                // always use the newly created github app token as the token will expire
+                this.headers.put("Authorization", String.format("Token %s", appInstallationToken.getToken()));
             } catch (Exception e) {
                 // e.printStackTrace();
                 LOG.error("Exception when getting Github token: ", e);
                 throw e;
             }
         }
-
-        this.headers.put("Authorization", String.format("Token %s", this.token));
```
akspocket, line 96: what's the issue with original code? 

Before:
--------
if (token empty):
  get token

set header

After:
------
if (token empty):
  get token
  set header
else
  set header

ankilosaurus, line 96: what's fixed here ?

what's the issue with original code?

Before:
--------
if (token empty):
  get token

set header

After:
------
if (token empty):
  get token
  set header
else
  set header

robbintt: i think this is just a minor fix putting headers once, the main fix is removing the extra getToken()

robbintt: now auth header won't be put if there is no token. which seems fine? :)

robbintt (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
akspocket: with change, will it ever cache token?

ankilosaurus: with change, will it ever cache token?

yongwen: We no longer cache token as the original problem is that the GitHub token was cached and expired

ankilosaurus: so do we still need else case and token variable ?

ankilosaurus: this.token is populated from json. Are we putting token in config yaml?

yongwen: The current config does not use personal token in config yaml but the code supports that, which is the reason for the else case.

PR #989, opened 2022-06-16
Increased log level for deploy_board.webapp.security

tylerwowen:
99% of production log is generated by `logger.debug("Redirect oauth for authentication!, url = " + url)`.
Increased the log level to INFO and bumped other log points to INFO.


kynanlalone-pinterest (approved):

tylerwowen:
Verified that the log was gone.

PR #1002, opened 2022-07-09
Added a buffer window for AutoPromoter

tylerwowen:
#392 changed the way AutoPromoter handles cron schedule. It made builds possible to be promoted every time AutoPromoter runs after their schedule time. For example, if a stage's auto promotion is scheduled at 10AM everyday, a build is possible to be promoted after the first succeeding 10AM. This doesn't meet user expectation. 

To minimize the change of the whole implementation, I here propose a solution to add a small buffer window only in which promotion is possible. Currently the buffer window is set to 2 minutes. Given the fact that we run the AutoPromoter every 30 seconds, 2 minutes should be long enough to process the schedule. This is also configurable so service owners can define their own buffer window size.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AutoPromoter.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AutoPromoter.java
@@ -48,7 +48,7 @@ import java.util.function.Function;
 public class AutoPromoter implements Runnable {
 
     public static final String AUTO_PROMOTER_NAME = "AutoPromoter";
-    public static final int BUFFER_TIME_MINUTE = 2;
+    public static final int DEFAULT_BUFFER_TIME_MINUTE = 2;
     private static final Logger LOG = LoggerFactory.getLogger(AutoPromoter.class);
     public BuildDAO buildDAO;
     private EnvironDAO environDAO;
@@ -57,6 +57,7 @@ public class AutoPromoter implements Runnable {
     private UtilDAO utilDAO;
     private DeployHandler deployHandler;
     private BuildTagsManager buildTagsManager;
+    private int bufferTimeMinutes;
     private final int maxCheckBuildsOrDeploys = 100;
 
 
@@ -68,6 +69,12 @@ public class AutoPromoter implements Runnable {
         buildDAO = serviceContext.getBuildDAO();
         buildTagsManager = new BuildTagsManagerImpl(serviceContext.getTagDAO());
         deployHandler = new DeployHandler(serviceContext);
+        bufferTimeMinutes = DEFAULT_BUFFER_TIME_MINUTE;
+    }
+
+    public AutoPromoter withBufferTimeMinutes(int bufferTime) {
+        bufferTimeMinutes = bufferTime;
+        return this;
     }
 
     void processBatch() throws Exception {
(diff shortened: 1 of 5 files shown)


deploy-service/teletraanservice/src/test/java/com/pinterest/teletraan/worker/AutoPromoteBuildTest.java
```diff
@@ -24,26 +24,47 @@
...
 import org.mockito.Mockito;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
+import java.util.function.Function;
 
 
-public class AutoPromotBuildTest {
+public class AutoPromoteBuildTest {
```
ntascii, line 40: good eye :)

deploy-service/teletraanservice/src/test/java/com/pinterest/teletraan/worker/AutoPromoteBuildTest.java
```diff
@@ -54,72 +75,37 @@ public void setUp() throws Exception {
...
+        t10AMPromoteBean = new PromoteBean();
+        t10AMPromoteBean.setEnv_id(environBean.getEnv_id());
+        t10AMPromoteBean.setSchedule(CronTenAMPerDay);
+        t10AMPromoteBean.setLast_update(0L);
+        t10AMPromoteBean.setDelay(0);
     }
 
-
-    @Test
-    public void testGetScheduledCheckDueTime() throws Exception {
```
ntascii, line 65: why removing this test?

deploy-service/teletraanservice/src/test/java/com/pinterest/teletraan/worker/AutoPromoteBuildTest.java
```diff
@@ -54,72 +75,37 @@ public void setUp() throws Exception {
...
-          new DateTime(now.getYear(), now.getMonthOfYear(), now.getDayOfMonth(), 10, 0, 0);
-      Assert.assertEquals(result.getRight().longValue(), start.getMillis());*/
-        //}
+    @After
+    public void tearDown() throws Exception {
+        DateTimeUtils.setCurrentMillisSystem();
     }
 
-    /* Autopromote enabled for any new build. But no builds and no prev deploys*/
+    /* AutoPromote enabled for any new build.
```
ntascii, line 106: not sure if I understand the new comment.

deploy-service/teletraanservice/src/test/java/com/pinterest/teletraan/worker/AutoPromoteBuildTest.java
tylerwowen: This is removed along with the no-longer-referenced function `GetScheduledCheckDueTime`.

ntascii:
Is there any way to put 2min in a config? so if we want to adjust the number later on we don't need to change the code.

tylerwowen:
> Is there any way to put 2min in a config? so if we want to adjust the number later on we don't need to change the code. Not sure if it's possible for OSS but worth double checking.

Good suggestion, done

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/AutoPromoter.java
```diff
@@ -229,6 +229,15 @@ public PromoteResult computePromoteBuildResult(EnvironBean currEnvBean,
             return new PromoteResult().withResultCode(PromoteResult.ResultCode.NoAvailableBuild);
         }
 
+        String schedule = promoteBean.getSchedule();
+        if (!StringUtils.isEmpty(schedule)) {
+            CronExpression cronExpression = new CronExpression(schedule);
+            DateTime nextDueTime = new DateTime(
+                    cronExpression.getTimeAfter(DateTime.now().minusMinutes(BUFFER_TIME_MINUTE).toDate()));
+            if (nextDueTime.isAfterNow()) {
```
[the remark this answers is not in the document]
tylerwowen: Done

ntascii (approved):

PR #1005, opened 2022-07-12
Filter base image help and host type help by cell

tylerwowen:
We should pass `cell` to requests since these 2 APIs are cell-aware.

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -571,7 +571,7 @@
                             })
                             capacitySetting.baseImageHelpData = data
                             capacitySetting.showBaseImageHelp = true
-                        })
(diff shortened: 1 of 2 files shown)


ntascii:
Just curious how you tested this change?

tylerwowen:
> Just curious how you tested this change?

I have a dump of the integ DB and just applied to my local container so I was able to get to the cluster creation UI.

ntascii:
Do we need to update this file as well https://sourcegraph.pinadmin.com/github.com/pinterest/teletraan/-/blob/deploy-board/deploy_board/templates/clusters/cluster_configuration.html?

tylerwowen:
> Do we need to update this file as well https://sourcegraph.pinadmin.com/github.com/pinterest/teletraan/-/blob/deploy-board/deploy_board/templates/clusters/cluster_configuration.html?

Thanks, updated.

ntascii (approved):

PR #563, opened 2018-02-15
change script value text area in script page from single line to multiple-lines

WeiCheng1992:
screen shot 2018-02-14 at 4 11 39 pm
screen shot 2018-02-14 at 4 08 26 pm

PR #1007, opened 2022-07-12
Remember scroll position so page doesn't jump after resizing textarea

tylerwowen:
This fixes a bug that causes UI Auto scrolls to top when editing a Teletraan stage "Script Config", or any text area. Introduced by #563 .

`$('textarea')` was removed because we should not autogrow heights of all textareas in the page.

Bug:

https://user-images.githubusercontent.com/8442875/179045697-e55e26bd-7e8a-4411-9a77-94178274264c.mov

With fix:

https://user-images.githubusercontent.com/8442875/179045728-bef9536e-6ee5-4d4b-994e-948b71976bf0.mov

### DIFF
--- a/deploy-board/deploy_board/templates/configs/config_map.tmpl
+++ b/deploy-board/deploy_board/templates/configs/config_map.tmpl
@@ -176,7 +176,7 @@
             $('#resetConfigMapBtnId').removeAttr('disabled');
         });
 
-        $('#mapConfigFieldSetId').on('change click keyup keydown', 'textarea', function () {
+        $('#mapConfigFieldSetId, #newEntryFormId').on('change click keyup keydown', 'textarea', function () {
             var scrollLeft = window.pageXOffset ||
                 (document.documentElement || document.body.parentNode || document.body).scrollLeft;
 


ntascii:
I'd suggest to attach a screenshot or screen capture video for these types of UI changes to help reviewers get how they look afterward.

deploy-board/deploy_board/templates/configs/config_map.tmpl
```diff
@@ -176,15 +176,18 @@
             $('#resetConfigMapBtnId').removeAttr('disabled');
         });
 
-        $('textarea').on('change click keyup keydown', function() {
-           this.style.height = "32px";
-           this.style.height = (this.scrollHeight + 2)+"px";
-        });
+        $('#mapConfigFieldSetId').on('hange click keyup keydown', 'textarea', function () {
```
osoriano, line 179:
```suggestion
        $('#mapConfigFieldSetId').on('change click keyup keydown', 'textarea', function () {
```

tylerwowen: Good eye, thanks

tylerwowen:
> I'd suggest to attach a screenshot or screen capture video for these types of UI changes to help reviewers get how they look afterward.

I usually would, but this is fixing a bug that otherwise causes page scrolls to top. Also added bug description in the top comment.

deploy-board/deploy_board/templates/configs/config_map.tmpl
```diff
@@ -176,15 +176,18 @@
...
 
-        $('#mapConfigFieldSetId').on('change click keyup keydown', 'textarea', function () {
-           this.style.height = "32px";
-           this.style.height = (this.scrollHeight + 2)+"px";
-        });
+            var scrollTop  = window.pageYOffset ||
+                (document.documentElement || document.body.parentNode || document.body).scrollTop;
+
+            this.style.height = 'auto';
+            this.style.height = this.scrollHeight + 'px';
```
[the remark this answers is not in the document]
tylerwowen: This is for sizing down. Without it, it only grows but doesn't shrink.

ntascii (dismissed):

tylerwowen:
> Thanks for uploading the videos, I can visualize the change now. How do you think if we make the textarea is scrollable with a fixed threshold size (like this comment box) instead of keep increasing its size per its content? It might simplify the way we handle the textarea. Your call to choose which one to go.

This seems to be a low use area. Also I think showing everything is helpful to our users so they don't have to scroll up or down.

ntascii (approved):

PR #1008, opened 2022-07-13
Fix teletraan displays incorrect AMI

ntascii:
Teletraan UI shows latest AMI selected, however, new instances will launch on an older AMI. When the AMI is older, the UI shows the first (latest) AMI selected, when in reality the cluster does not have this configuration. This confused users. Fix by including the current ami in the selection dropdown.

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -426,8 +426,18 @@
         });
         const initialAssignPublicIP = currentCluster.configs.assign_public_ip === "true" ;
         const initialCell = currentCluster.cellName;
+        var baseImageAccepted = info.baseImages.filter(function (o) {
+            return o.acceptance == 'ACCEPTED';
+        });
 
-        var sorted = info.baseImages.sort(function(item1, item2) {
+        if (currentCluster.baseImage != null) {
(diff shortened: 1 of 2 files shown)


ntascii:
Tested locally with an existing cluster from integ. 
Base image info. 
Screen Shot 2022-07-13 at 11 15 02 AM
AMI dropdown
Screen Shot 2022-07-13 at 11 15 17 AM

liyaqin1 (approved):

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -426,8 +426,18 @@ <h4 class="panel-title pull-left pointer-cursor">
         });
         const initialAssignPublicIP = currentCluster.configs.assign_public_ip === "true" ;
         const initialCell = currentCluster.cellName;
+        var baseImageAccepted = info.baseImages.filter(function (o) {
+            return o.acceptance == 'ACCEPTED';
+        });
+
+        if (currentCluster.baseImage != null) {
+            // Make sure in the base image dropdown the current base image exists.
+            if (!baseImageAccepted.some(image => image.id === currentCluster.baseImage.id)) {
```
tylerwowen, line 435: Maybe `Array.prototype.find()` is more readable here.

PR #1009, opened 2022-07-13
Show only accepted AMI in new capacity advance

ntascii:
Currently we show both denied and accepted amis. We had a logic to filter out but looks like we use it wrongly. This change filtered out to only accepted AMIs when creating a new capacity. 
Tested locally at env/canary/config/newcapacity/advanced/
before: 
Screen Shot 2022-07-13 at 12 56 12 PM
after:
Screen Shot 2022-07-13 at 12 10 05 PM

### DIFF
--- a/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
@@ -90,12 +90,13 @@ var capacityView = new Vue({
 
 var info = {{capacity_creation_info|safe}}
 var environment = info.environment;
-var accepted = info.baseImages.filter(function (e) {
+var baseImagesAccepted = info.baseImages.filter(function (e) {
         return e.acceptance == 'ACCEPTED';
     });
-var sorted = accepted.sort(function(item1,item2){
+var baseImagesAcceptedSorted = baseImagesAccepted.sort(function(item1,item2){
         return item2.publish_date - item1.publish_date;
 });
+const firstBaseImageId = baseImagesAcceptedSorted[0].id;
 var placements = getDefaultPlacement(info);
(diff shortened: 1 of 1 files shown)


tylerwowen:
I'm not sure if I can build a connection between the description and the changed code.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -109,16 +110,16 @@ <h4 class="panel-title pull-left">Capacity</h4>
             }).filter(function(item){return item.name!='cmp_group'}),
         assignPublicIP: false,
         baseImageHelpData: [],
-        baseimages: info.baseImages.map(
+        baseimages: sorted.map(
```
ntascii, line 113: main change

tylerwowen: I see, thanks.

tylerwowen: Maybe rename it to `sortedAccepted`

ntascii: sounds good.

liyaqin1 (approved):

osoriano (approved):

tylerwowen:
If we only show accepted, there are unknown and denied images. Should we not allow users to choose from those?

ntascii:
yeah let's keep this PR open and we can discuss more about that. We'll need to gather more context around that. But my current thought is AMI should be validated properly before being available to users. Note that this view is for creating a new capacity so using a declined AMI would be strange. Showing unknown and declined amis to users would introduce confusion instead (at least from not a newbie user like me).

PR #1010, opened 2022-07-19
updated logic to handle unsupported SCMs and SCM name origin from teletraanconfigs

rfleur01:


deploy-board/deploy_board/webapp/helpers/systems_helper.py
```diff
@@ -21,7 +21,7 @@
 deployclient = DeployClient()
 
 
-def get_url_pattern(request, type=""):
+def get_url_pattern(request, type=""):
```
robbintt, line 24: some extra whitespace

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/PhabricatorManager.java
```diff
@@ -41,7 +41,7 @@
 public class PhabricatorManager extends BaseManager {
 
     private static final Logger LOG = LoggerFactory.getLogger(PhabricatorManager.class);
-    public static final String TYPE = "Phabricator";
+    public static String TYPE;
```
robbintt, line 44: is TYPE still needed in this class?

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/SourceControlManagerProxy.java
```diff
@@ -31,7 +33,21 @@ private SourceControlManager getSourceControlManager(String scmType) throws Exce
...
     }
 
-    public String getType() {
+    public List<String> getSCMs() throws Exception {
+        validSCMs = new ArrayList<String> (this.managers.keySet());
+        return validSCMs;
+    }
+
+    public Boolean approved(String scmName) {
+        for (String validSCM : validSCMs) {
```
robbintt, line 42: try to use a function designed to check membership, e.g. https://stackoverflow.com/questions/37330551/how-to-check-whether-a-liststring-contains-a-specific-string

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
```diff
@@ -140,8 +140,8 @@ public static TeletraanServiceContext setupContext(TeletraanServiceConfiguration
         HashMap<String, SourceControlManager> managers = new HashMap<String, SourceControlManager>();
         for(SourceControlFactory scf : sourceControlConfigs) {
             SourceControlManager scm = scf.create();
-            String type = scm.getType();
-            managers.put(type, scm);
+            String type = scm.getTypeName();
+            managers.put(type, scm);
```
robbintt, line 144: remove trailing whitespace (might be an option for this in your IDE)

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
```diff
@@ -210,6 +210,15 @@ public Response publish(
                 "Artifact URL points to unapproved location.");
         }
 
+        // Initialize the list of Source Control Managers
+        sourceControlManagerProxy.getSCMs();
+
+        // Check if build SCM is approved via allow list of SCMs
+        if (!sourceControlManagerProxy.approved(buildBean.getScm())) {
```
robbintt, line 217: maybe `approved` function name could be clearer from the perspective of the reader in this module, e.g. `hasScmType`

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/PhabricatorManager.java
rfleur01: I kept it to keep the static reference to it in SourceControlManagerProxy. Since Phabricator is the default SCM and I did not want to use a hardcoded string, I assigned TYPE to the value of typeName (in this case "Phabricator" from the config file) and created a static method that returns TYPE since getTypeName() returns a non static variable (and shouldn't because the value of typeName comes from the config file).

deploy-board/deploy_board/webapp/helpers/systems_helper.py
rfleur01: ack

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
rfleur01: ack

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
rfleur01: Agreed. I will use` hasSCMType()`

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/PhabricatorManager.java
robbintt: Thanks, is it only consumed in SourceControlMangerProxy? If so we can change that to support the new non-static usage and reduce the total lines of code

rfleur01: Yes and I'm trying to find a way to do that, but it looks like I have to return a static variable for that static method referenced in SourceControlManagerProxy. The getTypeName() method wouldn't work because that's a method that can only be called by instances of the PhabricatorManager class. My understanding of a static method is that it acts as a class method, so if we want to default type to be "Phabricator" we need a way of passing the value of typeName (non-static) to TYPE (static) which I did by initializing TYPE and assigning it to typeName and returning it in the getStaticTypeName() method which is used in SourceControlManagerProxy

rfleur01:
The PR contains:

- removal of hard coded values in PhabricatorManager and GIthubManager classes and updated logic and methods to support the change. SCM typeName is pulled as a Json property from the config file instead.
- typeName property now allows us to create a list that is used to validate builds before publishing
- Creation of getter method and validation method for SCM typeName values in SourceControlManagerProxy  
- defaultTypeName property removes the hardcoded "Phabricator" value used as a default when the scm is ""
- Updates to scm unit test
- updates to sample config file so open source users can follow changes made in teletraanconfigs
- updates to existing method names to be more descriptive

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/SourceControlManagerProxy.java
```diff
@@ -31,8 +34,20 @@ private SourceControlManager getSourceControlManager(String scmType) throws Exce
         return manager;
     }
 
-    public String getType() {
-        return DEFAULT_TYPE;
+    public List<String> getSCMs() throws Exception {
+        validSCMs = new ArrayList<String> (this.managers.keySet());
```
yongwen, line 38: you can move the init of validSCM to the constructor above so this get will be just returning a variable.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
```diff
@@ -210,11 +210,20 @@ public Response publish(
                 "Artifact URL points to unapproved location.");
         }
 
+        // Initialize the list of Source Control Managers
+        sourceControlManagerProxy.getSCMs();
```
yongwen, line 214: see the above comment in the "getSCM" function. then there is no need for this line. as the initialization is done in constructor.  In general, a getX function would expect the return value is being used, so just a getX statement here is odd.

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
```diff
@@ -55,7 +55,8 @@ public class GithubManager extends BaseManager {
 
     public Map<String, String> headers = new HashMap<String, String>();
 
-    public GithubManager(String token, String appId, String appPrivateKeyKnox, String appOrganization, String apiPrefix, String urlPrefix) throws Exception {
```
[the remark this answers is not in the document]
rfleur01: I did the same thing for PhabricatorManager so that the typeName from configs gets populated when getTypeName() is called in ConfigHelper.java on line 144

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/SourceControlManagerProxy.java
rfleur01: Makes sense. I will update accordingly.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
rfleur01: Yeah I see why the use of it is unclear. I will update accordingly.

yongwen (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
```diff
@@ -167,8 +168,8 @@ public String getUrlPrefix() {
     }
 
     @Override
-    public String getType() {
-        return TYPE;
+    public String getTypeName() {
+        return typeName;
     }
```
tylerwowen, line 173: Can this be implemented in the BaseManager?

tylerwowen (changes requested):

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/SourceControlManagerProxy.java
```diff
@@ -31,8 +35,19 @@ private SourceControlManager getSourceControlManager(String scmType) throws Exce
...
+    }
+
+    public Boolean hasSCMType(String scmName) {
+        if (validSCMs.contains(scmName)) {
+            return true;
+        }
+        return false;
+    }
+
+    public String getTypeName() {
```
tylerwowen, line 49: I'd use `getDefualtTypeName`.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
```diff
@@ -210,11 +210,17 @@ public Response publish(
                 "Artifact URL points to unapproved location.");
         }
 
+        // Check if build SCM is approved via allow list of SCMs
+        if (!sourceControlManagerProxy.hasSCMType(buildBean.getScm())) {
```
tylerwowen, line 214: It looks to me that the if the configured type name is "Github" and "github" is received, then the service throws an error. Is it the intention?

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/SourceControlManagerProxy.java
```diff
@@ -31,8 +35,19 @@ private SourceControlManager getSourceControlManager(String scmType) throws Exce
...
     }
 
-    public String getType() {
-        return DEFAULT_TYPE;
+    public List<String> getSCMs() throws Exception {
+        return validSCMs;
+    }
+
+    public Boolean hasSCMType(String scmName) {
+        if (validSCMs.contains(scmName)) {
```
tylerwowen, line 43: `return validSCMs.contains(scmName)`

tylerwowen:
I'm still not persuaded that the introduction of a configurable `typeName` is needed for fixing the bug caused case sensitivity. Can you elaborate it more? 
I'd only change the `SourceControlManager:getSourceControlManager` to convert `scmType` to the same case as the `*Manager.Type`.

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
rfleur01: There's no way (to my knowledge) of getting the value of typeName into the class as an argument if a BaseManager object were to be instantiated, which is not possible since it is an abstract class

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/PhabricatorManager.java
```diff
@@ -181,8 +182,8 @@ public String getUrlPrefix() {
     }
 
     @Override
-    public String getType() {
-        return TYPE;
+    public String getTypeName() {
+        return typeName;
     }
```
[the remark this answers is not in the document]
rfleur01: There's no way (to my knowledge) of getting the value of typeName into the class as an argument if a BaseManager object were to be instantiated, which is not possible since it is an abstract class

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/PhabricatorFactory.java
```diff
@@ -24,6 +24,10 @@
 
 @JsonTypeName("phabricator")
 public class PhabricatorFactory implements SourceControlFactory {
+    @NotNull
+    @JsonProperty
+    private String typeName;
```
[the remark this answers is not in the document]
rfleur01: Yes, when a GithubManager or PhabricatorManager object is instantiated it will need to have access to this Json value from the config file since it holds a specific value

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/GithubFactory.java
```diff
@@ -37,6 +37,10 @@ public class GithubFactory implements SourceControlFactory {
     @JsonProperty
     private String appOrganization;
 
+    @NotNull
+    @JsonProperty
+    private String typeName;
```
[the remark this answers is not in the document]
rfleur01: Yes, when a GithubManager or PhabricatorManager object is instantiated it will need to have access to this Json value from the config file since it holds a specific value

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
rfleur01: Correct, that is the intention.

rfleur01:
> 

- Refactoring the SCM logic so that it no longer pulls the valid SCM values from the hardcoded values and pulls from configs was needed, as discussed and reviewed with  @robbintt 
- Blocking the publishing of builds that do not have valid SCM types is what addresses the bug. If we prevent the publishing of builds with invalid SCMs then those builds will never make it to Teletraan and which will prevent future 500 errors. We also have a follow up task to update the database to fix any existing unsupported SCMs. Those two steps will eliminate the possibility of causing the bug.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/GithubFactory.java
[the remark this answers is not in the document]
tylerwowen: Missed that part, thanks!

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/PhabricatorFactory.java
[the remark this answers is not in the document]
tylerwowen: Missed that part, thanks!

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
tylerwowen: I think we should actually allow this by attempt to convert user input to the same case. The conversion is very trivial. There are 2 types of invalid SCM types - invalid case and unsupported type. We do want to avoid unsupported type but should do our best to handle invalid case.

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
tylerwowen: I was thinking
```java
public class BaseManager {
    protected String typeName;
    public String getTypeName() {
        return typeName;
    }
} 

public class GirhubManager extends BaseManager {
      public GithubManager(String token, String appId, String appPrivateKeyKnox, String appOrganization, String 
        typeName, String apiPrefix, String urlPrefix) throws Exception {
            this.typeName = typeName;
  }
}
```

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
rfleur01: Okay then that change would need to be made to the BuildBean class that accepts the scm json value. I would have to reassign the scm initialized here:
`@JsonProperty("type")
    private String scm;`
to a capitalized version.

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
tylerwowen: Did you also remove the @Override implementation in GithubManager?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java
tylerwowen: Yes updating the setter of `private String scm` to capitalize all setter input is a good point. That ensures only capitalized SCM types are stored in the DB from now on. We should also update the config so `typeName` of each manager is capitalized. Last we want to capitalize user input. This may be an incomplete list, please ensure all other necessary scenarios are updated accordingly.

rfleur01: Sounds good. I already have a PR in Phabricator for the config updates you mentioned. I will implement what you mentioned here.

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
rfleur01: Removing private field typeName in GithubManager did the trick. I will implement and commit the solution

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/BuildBean.java
```diff
@@ -194,7 +195,7 @@ public SetClause genSetClause() {
         clause.addColumn("build_id", build_id);
         clause.addColumn("build_name", build_name);
         clause.addColumn("artifact_url", artifact_url);
-        clause.addColumn("scm", scm);
+        clause.addColumn("scm", StringUtils.capitalize(scm));
```
rfleur01, line 198: modified this to unsure the capitalized value is what gets committed to the database

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/BuildBean.java
```diff
@@ -117,11 +118,11 @@ public void setArtifact_url(String artifact_url) {
     }
 
     public String getScm() {
-        return scm;
+       return StringUtils.capitalize(scm);
```
rfleur01, line 121: Updated the getter and setter since I could not modify the @JsonProperty value directly

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
```diff
@@ -136,14 +136,15 @@ public static TeletraanServiceContext setupContext(TeletraanServiceConfiguration
...
         context.setHostGroupDAO(configuration.getHostGroupFactory().createHostGroupDAO());
 
+        String defaultScmTypeName = configuration.getDefaultScmTypeName();
         List<SourceControlFactory> sourceControlConfigs = configuration.getSourceControlConfigs();
         HashMap<String, SourceControlManager> managers = new HashMap<String, SourceControlManager>();
         for(SourceControlFactory scf : sourceControlConfigs) {
             SourceControlManager scm = scf.create();
-            String type = scm.getType();
+            String type = scm.getTypeName();
             managers.put(type, scm);
```
tylerwowen, line 145: I think this becomes a bug based on your implementation. `scm.getTypeName()` returns 'Github' and then `buildBean.getScm()` returns 'GITHUB`, `hasSCMType()` will always return false.

rfleur01: `StringUtils.capitalize(arg)` returns a String with only the first letter capitalized so `buildBean.getScm()` returns 'Github'

tylerwowen: I see. I thought it did all cap. Then could you use either all cap or all lower case? Having only the first letter capitalized is not a very good practice.

rfleur01: I see what you mean. We would be modifying each scm from the Json to be all caps. I will update the typeName in teletraanconfigs to reflect that as well so that when the scm is being checked against the valid scm list it doesn't result in an error

rfleur01: I'm getting a few test fails because when I call `toUpperCase()` on `scm` within the `genSetClause()` method in those tests it returns a nulPointerException because scm is not populated in those tests with`toUpperCase()` still being called on the `scm` variable. I also tested adding the scm value to the broken tests and that seemed to break other tests related to other beans like `TagBean`

rfleur01: I committed the updates along with some unit test updates that were needed

rfleur01:
As discussed in Slack with @tylerwowen, I
Updated the following function, since this is the interface where the internal representation meets the outside information.
`public Boolean hasSCMType(String scmName) {
    return validSCMs.contains(scmName); // validSCMs.stream().anyMatch(scmName::equalsIgnoreCase);
} `

Update the SourceControlManagerProxy to have a case-insensitve map

tylerwowen (approved):

tylerwowen (approved):

PR #1011, opened 2022-07-20
Updated teletraan service yaml

tylerwowen:
Typo


deploy-service/teletraanservice/bin/server-no-db.yaml
```diff
@@ -50,7 +50,7 @@ workers:
     properties:
       initialDelay: 20
       period: 30
-      bufferTimeMinutess: 2
+      bufferTimeMinutes: 2
```
ntascii, line 53: good catch

ntascii (approved):

PR #1022, opened 2022-08-09
updating message

rashmi59:
Teletraan customers frequently change their cluster settings (ami, instance type, etc) and wonder why their changes don't do anything. Hence updating the message to

Are you sure to save the changes?

Please note that the saved settings will only apply for newly launched hosts. If you have existing hosts in your cluster, you can use the "Replace Cluster" feature on the left to perform a rolling replacement of all existing hosts in your cluster.

Can see the updated popup on clicking save in the integ build associated with my changes
https://deploy-integ.pinadmin.com/env/helloworlddummyservice-server/yaqin-test-5/config/cluster/config/

Screen Shot 2022-08-09 at 1 52 38 PM

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -87,7 +87,7 @@
         </div>
     </div>
     <modal v-bind:title="confirmDialogTitle" v-bind:id="confirmDialogId" v-on:input="clickDialog">
-        <div slot="body">Are you sure to save the changes?</div>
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -87,7 +87,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
         </div>
     </div>
     <modal v-bind:title="confirmDialogTitle" v-bind:id="confirmDialogId" v-on:input="clickDialog">
-        <div slot="body">Are you sure to save the changes?</div>
+        <div slot="body">Are you sure to save the changes? <br><br>Please note that the saved settings will only apply for newly launched hosts. If you have existing hosts in your cluster, you can use the "Replace Cluster" feature on the left to perform a rolling replacement of all existing hosts in your cluster.</div>
```
tylerwowen, line 90: It would be nice if this is break into multiple lines.

tylerwowen (approved):

ntascii:
nice~

PR #1024, opened 2022-08-10
Modifications to render error message bar for Scheduled Actions and Scaling Alarms

rfleur01:
emit useful logs and display a failure page when scheduled action is below min cluster size or greater than max cluster size and when invalid Scaling Alarm is saved.

Test Plan:

1. Navigate to Autoscaling Settings page
2. Create a a Scheduled Action with a valid scheduled (i.e 0 1 * * *) and a Capacity > the Max Size of the Auto Scaling Group or < the Min Size of the Auto Scaling Group
Screen Shot 2022-08-26 at 4 49 05 PM
3. Confirm that you are taken to an error page upon saving
Screen Shot 2022-08-26 at 4 49 18 PM
4. Return to the Autoscaling Settings Page
5. Creating a Scaling Alarm with a negative values under "When the metric:"
Screen Shot 2022-08-26 at 4 50 42 PM
6. Confirm that you are taken to an error page upon saving
Screen Shot 2022-08-26 at 4 50 58 PM


deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -543,7 +543,7 @@ def add_alarms(request, group_name):
         autoscaling_groups_helper.add_alarm(request, group_name, [alarm_info])
     except:
         log.error(traceback.format_exc())
```
tylerwowen, line 545: Is the exception logged twice in this case? If it does, we don't need a try-catch here.

tylerwowen: I noticed the pattern in this file. This is probably fine then.

rfleur01: It technically logs it twice to the console. If I move if statement and its variables to before the try statement it logs once and still renders the error page, while still maintaining the original functionality.

tylerwowen: Looks like the try-catch was to log the error, and fail silently. If the purpose of this PR is to not fail silently, then you can simply remove the try-catch. I think that way any exception will be logged and surface up to the user.

tylerwowen: When you put the `raise` in the catch, it will propagate the exception and thus not fail silently.

rfleur01: I see and now that the IF statement is outside of the try-catch, I will remove the `raise` statement that I put in originally. That will allow `autoscaling_groups_helper.add_scheduled_actions()` to fail silently

deploy-board/deploy_board/webapp/group_view.py
```diff
@@ -521,28 +521,27 @@ def delete_alarms(request, group_name):
...
+        if "metricUrl" in params:
+            alarm_info["metricSource"] = params["metricUrl"]
+    else:
+        alarm_info["fromAwsMetric"] = True
+        if "awsMetrics" in params:
+            alarm_info["metricSource"] = params["awsMetrics"]
+    alarm_info["groupName"] = group_name
+    autoscaling_groups_helper.add_alarm(request, group_name, [alarm_info])
+   
+    log.error(traceback.format_exc())
```
osoriano, line 544:
```suggestion
```
Seems this line no longer applies since no exception is being handled

rfleur01: You're right. I will remove it

tylerwowen (approved):
LGTM

PR #1025, opened 2022-08-11
Fixing site health metrics

rashmi59:
Fixing site health metrics. 

Already have teletraan config PR up.

Corresponding image in local - 
Screen Shot 2022-08-19 at 1 54 24 PM

### DIFF
--- a/deploy-board/deploy_board/webapp/util_views.py
+++ b/deploy-board/deploy_board/webapp/util_views.py
@@ -42,6 +42,8 @@ def _get_latest_metrics(url):
     data = json.loads(response.read())
     # Return the first datapoint in the datapoints list
     if data:
+        if 'data' in data:
+            data = data['data']
         if 'datapoints' in data[0] and len(data[0]['datapoints']) != 0:
             return data[0]['datapoints']
             # Check for TSDB response
--- a/deploy-board/deploy_board/webapp/util_views.py
+++ b/deploy-board/deploy_board/webapp/util_views.py
@@ -39,18 +39,16 @@ def _convert_opentsdb_data(dps):
 
 def _get_latest_metrics(url):
     response = urllib2.urlopen(url)
-    decoded_response = str(response.read()).decode('string_escape')
-    data = json.loads(decoded_response)
+    data = json.load(response)
 
     # Return the first datapoint in the datapoints list
     if data:
-        if 'data' in data:
-            data = data['data']
-        if 'datapoints' in data[0] and len(data[0]['datapoints']) != 0:
-            datapoints = data[0]['datapoints']
-            datapoints_without_none = [datapoint for datapoint in datapoints if datapoint[1] != None]
-            return datapoints_without_none
-            # Check for TSDB response
+        try:
+            return [datapoint for datapoint in data['data'][0]['datapoints'] if datapoint[1] != None]
+        except:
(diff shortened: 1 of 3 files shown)


ntascii:
can you share a screenshot of the new change?

rashmi59:
> can you share a screenshot of the new change?

Done

tylerwowen (changes requested):
Commenting out is not a fix. Even if you have to comment it out, a TODO with the associated bug number is still strongly desired. If you do think this is not needed for pinterest, it should be conditionally disabled. Last, what prevents you "figure out the meaning and purpose of those metrics"?

rashmi59:
> Commenting out is not a fix. Even if you have to comment it out, a TODO with the associated bug number is still strongly desired. If you do think this is not needed for Pinterest, it should be conditionally disabled. Last, what prevents you from "figure out the meaning and purpose of those metrics"?

I know that commenting out is not a fix and hence I mentioned that I am figuring out in the meanwhile. Nothing is preventing me from doing so, only that it is taking me a while to understand the systems and mechanisms of how things work over here and it seems that there are inconsistencies across dev, local, integ and prod versions of the application adding to the time. I believe in not having things which do not work live as it is not providing any value to the user while being broken and new users who are onboarding might be confused. If this is unnecessary and not how it works here, I can shelf this PR.

deploy-board/deploy_board/webapp/util_views.py
```diff
@@ -39,11 +39,17 @@ def _convert_opentsdb_data(dps):
 
 def _get_latest_metrics(url):
     response = urllib2.urlopen(url)
-    data = json.loads(response.read())
+    decoded_response = str(response.read()).decode('string_escape')
```
tylerwowen, line 42: Can you explain this a little more? This only works python2, we should try to avoid tech debt.

rashmi59: Without this line it is unable to parse json object. I saw on removing escape sequences, it works. I can look for a python 3 alternative. Although even urllib2 is a python 2 package.

tylerwowen: `data = json.load(response)` worked for me.

deploy-board/deploy_board/webapp/util_views.py
```diff
@@ -39,11 +39,17 @@ def _convert_opentsdb_data(dps):
...
+    data = json.loads(decoded_response)
+
     # Return the first datapoint in the datapoints list
     if data:
+        if 'data' in data:
```
tylerwowen, line 47: This put the TSDB case in risk. I'd avoid this and do something like 
`if 'data' in data and 'datapoints' in data['data'][0] and len(data['data'][0]['datapoints']) > 0:`

tylerwowen: Or even better
```python
try:
  return [dp for dp in data['data][0]['datapoints'] if dp[1] != None]
except:
  pass
```

deploy-board/deploy_board/webapp/util_views.py
rashmi59: thanks, apologies for doing it in a convoluted manner

tylerwowen (approved):
LGTM

PR #1026, opened 2022-08-16
Use `is not None` so empty list can pass if check.

tylerwowen:
Use `is not None` so empty list can pass if check. This is required when use tries to delete all hosts or groups.

Regression introduced in #987 

Tests:
Test in local setup that the last existing capacity can be deleted.

### DIFF
--- a/deploy-board/deploy_board/webapp/capacity_views.py
+++ b/deploy-board/deploy_board/webapp/capacity_views.py
@@ -109,11 +109,11 @@ class EnvCapacityConfigView(View):
         logger.info("Post to capacity with data {0}".format(request.body))
         data = json.loads(request.body)
         hosts = data.get('hosts')
-        if hosts:
+        if hosts is not None:
             environs_helper.update_env_capacity(request, name, stage, capacity_type="HOST", data=hosts)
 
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/capacity_views.py
```diff
@@ -109,11 +109,11 @@ def post(self, request, name, stage):
         logger.info("Post to capacity with data {0}".format(request.body))
         data = json.loads(request.body)
         hosts = data.get('hosts')
-        if hosts:
+        if hosts is not None:
```
robbintt, line 112: can request.body ever be None? yay python...

robbintt (approved):

deploy-board/deploy_board/webapp/capacity_views.py
tylerwowen: Not in my test cases. I was thinking of removing the check, but it doesn't hurt to be extra cautious.

tylerwowen: Actually if `request.body` is `{}`, then `hsots` is `None`.

PR #1027, opened 2022-08-17
Groups config page needs subnets and public IP UI updates

rfleur01:
In order render the UI components from the Cluster settings, the cluster needs to exist. The Group Launch Configuration page is a "create" page if the Group does not exist, but the create button is greyed out  even if you populate the fields. 

Solution

- render a message to the user to create a cluster in their environment rather than rendering the Group Launch Configuration. Even if they were able to create a group it would be associated with any environment or stage. 

Test Plan:

- Navigate to the following URL path /groups/{{group that does not exist}}/config/
- Confirm that an error message page is rendered
- Screen Shot 2022-08-25 at 2 39 12 PM
- Navigate to the following URL path /groups/{{group that does exist}}/config/
- Confirm that the Group Configuration page renders
- Screen Shot 2022-08-25 at 2 38 58 PM


yongwen:
thanks. I was confused by the test plan:   the step 3 is the same text as the step 1 and why the result is different. probably you meant step 3 is going to an existing groups' config page.

yongwen (approved):
code looks good to me.

rfleur01:
> thanks. I was confused by the test plan: the step 3 is the same text as the step 1 and why the result is different. probably you meant step 3 is going to an existing groups' config page.

Formatting issue. I corrected it

PR #1029, opened 2022-08-23
Remove PY3 check in deploy agent

ntascii:
Removed related py2 code as we are using py3. 
Removed tox ut for py2. 
UTs + built and ran locally with `python setup.py install`

### DIFF
--- a/deploy-agent/deployd/common/caller.py
+++ b/deploy-agent/deployd/common/caller.py
@@ -17,8 +17,6 @@ import traceback
 import logging
 import time
 
-from future.utils import PY3
-
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


robbintt:
Did you inspect the dependencies? The one I know of is `lockfile`, a deprecated library that only lists itself as valid til python 3.3: https://pypi.org/project/lockfile/0.10.2/

3.3 is not guaranteed forwards compatible in Python. Also need to look at remaining libraries.

robbintt (changes requested):
(request changes requires comment)

osoriano:
Closing outdated PR :sparkles:

PR #1030, opened 2022-08-26
adding stopping script 

rashmi59:
Deploy sentinel does not contain stopping script, which makes it difficult to test whether stopping script is being called. Hence adding the stopping script.

Also adding a new state to rodimus named TELETRAAN_STOP_REQUESTED to call teletraan API from Rodimus to enable graceful termination.

Removing knox authentication, as it prevents users from deploying to deploy-open-sentinel.

Tested and verified in integ using deploy_open_sentinel to see that stopping script is called and host is correctly marked for termination. 

Screen Shot 2022-09-09 at 4 43 15 PM
Screen Shot 2022-09-09 at 4 44 17 PM
Screen Shot 2022-09-02 at 12 13 09 PM


deploy-sentinel/teletraan/STOPPING
```diff
@@ -0,0 +1,2 @@
+
+echo Running STOPPING script.
```
yongwen, line 2: you may want to add a long sleep here so you can ssh into the host to see the log.   as the log will only available on the host.

rashmi59: done.

tylerwowen: I think having the echo before sleep makes more sense.

deploy-sentinel/teletraan/STOPPING
```diff
@@ -0,0 +1,4 @@
+
```
tylerwowen, line 1: shebang?

deploy-sentinel/teletraan/STOPPING
rashmi59: I believe the purpose of having sleep is to give time to ssh into the system, else we might never see the echo in the console. Correct me if I am wrong

yongwen: agree with Tyler that sleep after echo is better.  the output of the script will be stored in /var/log/deployd/<service>.log, so after you ssh in, you can check the output during the sleep period.  otherwise, the script will exit and the host will be terminated after the script exit.    another suggestion is to increase the 5min to 10min so you have enough time.

deploy-board/deploy_board/webapp/templatetags/utils.py
```diff
@@ -143,6 +143,7 @@
     "UNKNOWN": "fa fa-spinner fa-spin",
     "SUCCEEDED": "fa fa-spinner fa-spin color-green",
     "QUALIFIED": "fa fa-check-circle color-green",
+    "TELETRAAN_STOP_REQUESTED": "fa fa-circle-stop fa-blink color-yellow",
```
tylerwowen, line 146: I'd stick with green.

tylerwowen (approved):

PR #1032, opened 2022-08-26
Anhn/telefig unavailable metrics

ntascii:


deploy-agent/deployd/agent.py
```diff
@@ -111,8 +112,12 @@ def _send_deploy_status_stats(self, deploy_report):
             tags['stage_name'] = self._response.deployGoal.stageName
         if deploy_report.status_code:
             tags['status_code'] = deploy_report.status_code
-        if deploy_report.output_msg and deploy_report.output_msg.find("teletraan_config_manager") != -1:
-            tags['error_source'] = DeployErrorSource.TELEFIG
+        if deploy_report.output_msg: 
+            if deploy_report.output_msg.find("teletraan_config_manager") != -1:
+                tags['error_source'] = DeployErrorSource.TELEFIG
+            if check_telefig_unavaialble_error(deploy_report.output_msg):
```
robbintt, line 118: nit: typo

deploy-agent/deployd/common/utils.py
```diff
@@ -147,3 +148,6 @@ def check_not_none(arg, msg=None):
         raise ValueError(msg)
     return arg
 
+def check_telefig_unavaialble_error(msg):
+    telefig_unavaialble_error = "{}: No such file or directory".format(TELEFIG_BINARY)
```
robbintt, line 152: nit: same typo

deploy-agent/deployd/common/utils.py
```diff
@@ -29,6 +29,7 @@
 
 log = logging.getLogger(__name__)
 
+TELEFIG_BINARY = "configure-serviceset"
```
robbintt, line 32: ideally this can go in some configuration file to collect hardcoded values

robbintt: otherwise there will be a mysterious error years down the road when this changes

deploy-agent/tests/unit/deploy/server/test_agent.py
```diff
@@ -545,5 +545,31 @@ def test_send_deploy_status_with_errors(self, mock_create_sc):
...
+            PingResponse(jsonValue=self.ping_response1),
+            PingResponse(jsonValue=self.ping_noop_response)
+        ]
+        mock_create_sc.return_value = None
+
+        client = mock.Mock()
+        client.send_reports = mock.Mock(side_effect=ping_response_list)
+        agent = DeployAgent(client=client, estatus=estatus, conf=self.config,
+                            executor=self.executor, helper=self.helper)
+        report = DeployReport(status_code=AgentStatus.UNKNOWN, error_code=1, output_msg="/usr/local/bin/configure-serviceset: No such file or directory", retry_times=3)
```
robbintt, line 567: if you choose to put TELEFIG_BINARY in config, then you can reference it here, to avoid 2 hardcoded values relying on each other.

deploy-agent/deployd/agent.py
yongwen, line 118: maybe we could do "else" so the string compare does not happen twice which might be expensive if the output msg is large.  maybe check "unavailable error" first , else check others.

deploy-agent/deployd/common/utils.py
ntascii: Curious how putting in a config file could avoid errors? Basically, we need to sync the current telefig bin with the code in this repo. Whether it stays here or in a config file, I don't see much difference unless there is an auto sync mechanism

robbintt: The point is not config, the point is having in a discoverable location. Hardcoded values strewn throughout the repo causes headaches especially from external references, in this case the telefig repo.  Better to have all in one place, the natural place is the configuration.

robbintt: by the way, this is an internal service, so it might be good to consider not having any reference to it here in the open source repo

deploy-agent/tests/unit/deploy/server/test_agent.py
ntascii: We can refer the const in common utils here as well but it's a mocking value of an error message. We probably would not want to build a mock message based on the logic of the tested function.

deploy-agent/deployd/agent.py
ntascii: yeah true, output msg can be large

deploy-agent/tests/unit/deploy/server/test_agent.py
robbintt: yeah, i have seen what you are referring to before, it's called "ugly mirror".  I am just referring to the telefig value, not the whole message.

deploy-agent/deployd/common/utils.py
ntascii: yeah great point. let me double check on that.

ntascii:
Due to the weird config between my folked repo and main repo. This introduced too many commit. Let's continue the review in the another PR. https://github.com/pinterest/teletraan/pull/1033

PR #1037, opened 2022-08-31
[Configurable Replacement Timeout] UI changes

tylerwowen:
Added the text input fields with validation for replacement timeout configuration

replacement_timeout

### New cluster
1. Navigate to advanced new capacity page /env/helloworlddummyservice-server/test-integ-1/config/newcapacity/advanced/.
2. Observed that 
   1. the field _Replacement Timeout_ exist **only** in the advanced view.
   2. can't save a value smaller than 5 or larger than 1440.
   3. the default 45 is populated.
   4. hovering the title will show the help message.
   5. can save the new value successfully.
 ### Existing cluster
1. Navigate to advanced cluster config page /env/helloworlddummyservice-server/test-tyler/config/cluster/config/.
2. Observed that 
   1. the field _Replacement Timeout_ exist **only** in the advanced view.
   2. can't save a value smaller than 5 or larger than 1440.
   3. the correct current value populated.
   4. hovering the title will show the help message.
   5. can save the new value successfully.

### DIFF
--- a/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
@@ -52,6 +52,7 @@
                     <placements-select label="Placements" title="Placements" v-bind:selectoptions="placements" v-bind:assignpublicip="assignPublicIP"
                         showhelp="true" v-on:helpclick="placementsHelpClick" v-on:assignpublicipclick="selectpublicip"></placements-select>
                     <placements-help v-show="showPlacementsHelp" v-bind:data="placementsHelpData"></placements-help>
+                    <label-input label="Replacement Timeout" placeholder="2700" v-model="replacementTimeout"></label-input>
                     <div>
                         <div class="form-group"></div>
                         <div class="form-group">
@@ -167,6 +168,7 @@ var capacitySetting = new Vue({
                     isSelected: item.name == info.defaultCell
                 }
             }),
+        replacementTimeout: 2700,
         securityZoneHelpData: [],
         securityZones: info.securityZones.map(
             function (item) {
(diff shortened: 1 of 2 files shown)


yongwen (dismissed):
r+.  just a comment that there is a cluster replacement configuration "Maximum Replaceable Host" is configured in the environment/stage config UI, not in the cluster config UI. in the future, we could think about consolidate them in one place.

ankilosaurus (dismissed):
can you add help string? 
For ex: 
1. what timeout means? 
2. Unit of timeout?
3. what's default if user did not provide timeout?

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -52,6 +52,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
                     <placements-select label="Placements" title="Placements" v-bind:selectoptions="placements" v-bind:assignpublicip="assignPublicIP"
                         showhelp="true" v-on:helpclick="placementsHelpClick" v-on:assignpublicipclick="selectpublicip"></placements-select>
                     <placements-help v-show="showPlacementsHelp" v-bind:data="placementsHelpData"></placements-help>
+                    <label-input label="Replacement Timeout" placeholder="2700" v-model="replacementTimeout" title="Time out for cluster replacement in seconds"></label-input>
```
ankilosaurus, line 55: be consistent with unit in help and error message?

ankilosaurus: I think unit should be in minutes for cluster replacement timeout.

tylerwowen: done

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -472,6 +474,12 @@ <h4 class="panel-title pull-left">Capacity</h4>
                 return false
             }
 
+            const replacementTimeout = clusterInfo['replacementTimeout']
+            if (replacementTimeout == null || isNaN(replacementTimeout) || replacementTimeout < 5 || replacementTimeout > 24 * 60) {
```
ankilosaurus, line 478: looks like this check can be bypassed using API.

tylerwowen: The API has its own check

PR #1038, opened 2022-09-01
Vuln update io dropwizard

aagxxi:
update io.dropwizard 1.3.29 to 2.1.1
update swagger 1.5.5 to 1.6.6


aagxxi:
This update not to be applied before updating io.dropwizard in Rodimus.

PR #1043, opened 2022-09-09
Fix slow db query for host tags

osoriano:
While looking into CDP-4769, this database query was very slow:

```
SELECT host_id, tag_value, tag_name, host_name
FROM (
  SELECT DISTINCT(host_tags.host_id) AS host_id, host_tags.tag_value AS tag_value, host_tags.tag_name AS tag_name, agents.host_name AS host_name, host_tags.env_id AS env_id
  FROM agents INNER JOIN host_tags ON agents.host_id = host_tags.host_id
) AS p1
WHERE p1.tag_name = 'availability_zone' AND p1.env_id = 'REDACTED';

... rows in set (3.74 sec)
```

Removing the unnecessary subquery greatly improved the performance:

```
SELECT DISTINCT(host_tags.host_id) AS host_id, host_tags.tag_value AS tag_value, host_tags.tag_name AS tag_name, agents.host_name AS host_name
FROM agents INNER JOIN host_tags ON agents.host_id = host_tags.host_id
WHERE host_tags.tag_name = 'availability_zone' AND agents.env_id = 'REDACTED';

... rows in set (0.01 sec)
```


osoriano:
Tested this change on a local setup. Will also verify in integ

robbintt (approved):

PR #1059, opened 2022-10-12
Teletraan UI (deployboard) does not gracefully handle nimbus failure

rfleur01:
Background:
In deployboard if nimbus returns a 5xx http status then deployboard page fails. This means that browsing some pages will fail to load if nimbus is unavailable.

See https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/webapp/helpers/nimbusclient.py#L25

Solution:
Surface the Nimbus error message to the user as a banner rather than a whole page that restrict access to deploy-board 

Test Plan:

- Modify the code to force an Nimbus error
- Navigate to an environment and confirm that the error banner is displayed and that the environment page renders properly
- image (6)

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -698,14 +698,22 @@ def clone_cluster(request, src_name, src_stage):
     except NotAuthorizedException as e:
         log.error("Have an NotAuthorizedException error {}".format(e))
         if external_id is not None:
-            environs_helper.delete_nimbus_identifier(request, external_id)
+            try:
+                environs_helper.delete_nimbus_identifier(request, external_id)
+            except TeletraanException as detail:
+                log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+                messages.add_message(request, messages.ERROR, detail) 
 
         return HttpResponse(e, status=403, content_type="application/json")
     except Exception as e:
         log.error("Failed to clone cluster env_name: %s, stage_name: %s" % (src_name, src_stage))
         log.error(traceback.format_exc())
         if external_id is not None:
-            environs_helper.delete_nimbus_identifier(request, external_id)
+            try:
+                environs_helper.delete_nimbus_identifier(request, external_id)
+            except TeletraanException as detail:
+                log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+                messages.add_message(request, messages.ERROR, detail) 
         return HttpResponse(e, status=500, content_type="application/json")
 
 
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -703,7 +703,6 @@ def clone_cluster(request, src_name, src_stage):
             except Exception as detail:
                 message = 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
                 log.error(message)
-                return HttpResponse(message, status=403, content_type="application/json")
 
         return HttpResponse(e, status=403, content_type="application/json")
     except Exception as e:
@@ -715,7 +714,6 @@ def clone_cluster(request, src_name, src_stage):
             except Exception as detail:
                 message = 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
                 log.error(message)
-                return HttpResponse(message, status=500, content_type="application/json")
                 
         return HttpResponse(e, status=500, content_type="application/json")
 
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -700,9 +700,10 @@ def clone_cluster(request, src_name, src_stage):
         if external_id is not None:
             try:
                 environs_helper.delete_nimbus_identifier(request, external_id)
-            except TeletraanException as detail:
-                log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
-                messages.add_message(request, messages.ERROR, detail) 
+            except Exception as detail:
+                message = 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                log.error(message)
+                return HttpResponse(message, status=403, content_type="application/json")
 
         return HttpResponse(e, status=403, content_type="application/json")
     except Exception as e:
@@ -711,9 +712,11 @@ def clone_cluster(request, src_name, src_stage):
         if external_id is not None:
             try:
                 environs_helper.delete_nimbus_identifier(request, external_id)
-            except TeletraanException as detail:
-                log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
-                messages.add_message(request, messages.ERROR, detail) 
+            except Exception as detail:
+                message = 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                log.error(message)
+                return HttpResponse(message, status=500, content_type="application/json")
+                
         return HttpResponse(e, status=500, content_type="application/json")
 
 
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -331,17 +331,16 @@ class EnvLandingView(View):
         stage_with_external_id = None
         existing_stage_identifier = None
         for env_stage in envs:
-            if env_stage['externalId'] is not None and env_stage['stageName'] == stage:
+            if env_stage['externalId'] is None and env_stage['stageName'] == stage:
                 stage_with_external_id = env_stage
                 break
 
-        if stage_with_external_id is not None and stage_with_external_id['externalId'] is not None:
-            try:
-                existing_stage_identifier = environs_helper.get_nimbus_identifier(request, stage_with_external_id['externalId'])
-                project_name_is_default = True if existing_stage_identifier is not None and existing_stage_identifier['projectName'] == "default" else False
-            except TeletraanException as detail:
-                log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
-                messages.add_message(request, messages.ERROR, detail)
+        try:
+            existing_stage_identifier = environs_helper.get_nimbus_identifier(request, stage_with_external_id['externalId'])
+            project_name_is_default = True if existing_stage_identifier is not None and existing_stage_identifier['projectName'] == "default" else False
+        except TeletraanException as detail:
+            log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+            messages.add_message(request, messages.ERROR, detail)
 
         project_info = None
         if existing_stage_identifier:
@@ -349,7 +348,11 @@ class EnvLandingView(View):
             if project_name:
                 project_info = {}
                 project_info['project_name'] = project_name
+            try:    
                 project_info['project_url'] = environs_helper.get_nimbus_project_console_url(project_name)
+            except TeletraanException as detail:
+                log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+                messages.add_message(request, messages.ERROR, detail)    
 
         if IS_PINTEREST:
             basic_cluster_info = clusters_helper.get_cluster(request, env.get('clusterName'))
@@ -840,7 +843,11 @@ def post_add_stage(request, name):
         else:
             common.create_simple_stage(request,name, stage, description, external_id)
     except:
-        environs_helper.delete_nimbus_identifier(request, external_id)
+        try:
+            environs_helper.delete_nimbus_identifier(request, external_id)
+        except TeletraanException as detail:
+            log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+            messages.add_message(request, messages.ERROR, detail)     
         raise
 
     return redirect('/env/' + name + '/' + stage + '/config/')
@@ -855,7 +862,11 @@ def remove_stage(request, name, stage):
             break
 
     if current_env_stage_with_external_id is not None and current_env_stage_with_external_id['externalId'] is not None:
-        environs_helper.delete_nimbus_identifier(request, current_env_stage_with_external_id['externalId'])
+        try:
+            environs_helper.delete_nimbus_identifier(request, current_env_stage_with_external_id['externalId'])
+        except TeletraanException as detail:
+            log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+            messages.add_message(request, messages.ERROR, detail)    
 
     environs_helper.delete_env(request, name, stage)
     envs = environs_helper.get_all_env_stages(request, name)
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -795,7 +795,7 @@ def post_create_env(request):
             external_id = environs_helper.create_identifier_for_new_stage(request, env_name, stage_name)
             common.clone_from_stage_name(request, env_name, stage_name, clone_env_name,
                                         clone_stage_name, description, external_id)
-        except TeletraanException as detail:
+        except Exception as detail:
             if external_id:
                 try:
                     environs_helper.delete_nimbus_identifier(request, external_id)
@@ -805,7 +805,7 @@ def post_create_env(request):
             else:
                 message = 'Failed to create identifier for {}/{}: {}'.format(env_name, stage_name, detail)
                 messages.add_message(request, messages.ERROR, message)
-                raise detail
+        raise detail
     else:
         data = {}
         data['envName'] = env_name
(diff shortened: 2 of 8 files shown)


tylerwowen (changes requested):
This doesn't seem like a right fix to me. 

Please find all references of the following methods, and make sure their callers handle exceptions properly.
```py
def get_nimbus_identifier(request, name):
    return nimbusclient.get_one_identifier(name, token=request.teletraan_user_id.token)


def create_nimbus_identifier(request, data):
    return nimbusclient.create_one_identifier(data, token=request.teletraan_user_id.token)


def delete_nimbus_identifier(request, name):
    # Nimbus identifiers are only for Pinterest environments
    if not IS_PINTEREST:
        return None
    return nimbusclient.delete_one_identifier(name, token=request.teletraan_user_id.token)


def get_nimbus_project_console_url(project_name):
    return nimbusclient.get_one_project_console_url(project_name)


```

rfleur01:
@tylerwowen by "handle exceptions properly" are you saying that these methods when called should be surrounded by try-except statement? For my understanding, can you elaborate on why the fix doesn't seem right?

tylerwowen:
Yes. The one you modified is an example. Since it is handled, it actually show the expected page instead of the error page. The problem is when the exception is not handled and thus user can't see the expected page.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -854,7 +863,11 @@ def remove_stage(request, name, stage):
             break
 
     if current_env_stage_with_external_id is not None and current_env_stage_with_external_id['externalId'] is not None:
-        environs_helper.delete_nimbus_identifier(request, current_env_stage_with_external_id['externalId'])
+        try:
+            environs_helper.delete_nimbus_identifier(request, current_env_stage_with_external_id['externalId'])
+        except TeletraanException as detail:
+            log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+            messages.add_message(request, messages.ERROR, detail)
```
tylerwowen, line 870: Same clean up attempt here.

tylerwowen (changes requested):
You will need the template in all pages involved so they can show the error. 
```
{% if messages %}
...
```
And because this is used multiple place, you probably want to extract it to a one place, and reference it else where.

Also I think we should use this as an opportunity to improve the nimbus failure handling. We should try to improve the error message to be more instructive so users know how to clean up nimbus resources manually.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -839,7 +844,11 @@ def post_add_stage(request, name):
         else:
             common.create_simple_stage(request,name, stage, description, external_id)
     except:
-        environs_helper.delete_nimbus_identifier(request, external_id)
+        try:
+            environs_helper.delete_nimbus_identifier(request, external_id)
+        except TeletraanException as detail:
+            log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+            messages.add_message(request, messages.ERROR, detail)
```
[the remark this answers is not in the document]
rfleur01: So are you saying that we do or do not want to display the message to the user? if the nimbus identifier was not created and `delete_nimbus_identifier` fails shouldn't we just log the error and process with the operation? Since there is nothing to delete  and you said that we shouldn't show that kind of error to the user we should get rid of the raise statements.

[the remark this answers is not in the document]
tylerwowen: Sorry for the confusion. My second thought is to show the instructive error message. Note that the inner exception is based on the outer exception, so both errors are returned to users.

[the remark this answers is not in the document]
rfleur01: Got it. Thanks for clarifying and this should probably surfaced as a banner since we still want to maintain render the page

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -839,8 +844,12 @@ def post_add_stage(request, name):
...
             common.create_simple_stage(request,name, stage, description, external_id)
     except:
-        environs_helper.delete_nimbus_identifier(request, external_id)
-        raise
+        try:
+            environs_helper.delete_nimbus_identifier(request, external_id)
+        except TeletraanException as detail:
+            message = 'Failed to delete Nimbus identifier {}, Error Message: {}'.format(external_id, detail)
+            log.error(message)
+            messages.add_message(request, messages.ERROR, message)
```
tylerwowen, line 852: This implementation doesn't propagate the original issue to the user. For example, if `clone_from_stage_name` fails before it created a nimbus identifier and results in an exception `originalException`, this call `delete_nimbus_identifier` is expected to fail as well. In this case `originalException` is more important than `detail`.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -854,7 +863,12 @@ def remove_stage(request, name, stage):
             break
 
     if current_env_stage_with_external_id is not None and current_env_stage_with_external_id['externalId'] is not None:
-        environs_helper.delete_nimbus_identifier(request, current_env_stage_with_external_id['externalId'])
+        try:
+            environs_helper.delete_nimbus_identifier(request, current_env_stage_with_external_id['externalId'])
+        except TeletraanException as detail:
+            message = 'Failed to delete Nimbus identifier {}, Error Message: {}'.format(current_env_stage_with_external_id['externalId'], detail)
```
tylerwowen, line 869: Here we can suggest user to verify if the identifier no longer exists. We often see users report that nimbus identifier already exists when they try to create a new stage.

deploy-board/deploy_board/webapp/helpers/environs_helper.py
```diff
@@ -89,14 +91,22 @@ def create_identifier_for_new_stage(request, env_name, stage_name):
         return None
 
     # retrieve Nimbus identifier for existing_stage
-    existing_stage_identifier = get_nimbus_identifier(request, stage_with_external_id['externalId'])
+    try:
+        existing_stage_identifier = get_nimbus_identifier(request, stage_with_external_id['externalId'])
+    except TeletraanException as detail:
+        log.error('Handling TeletraanException when trying to access nimbus API, error message {}'.format(detail))
+        messages.add_message(request, messages.ERROR, detail)
```
tylerwowen, line 98: This change will cause the function to continue execute. Is it desired?

deploy-board/deploy_board/webapp/env_views.py
rfleur01: I see what you mean. I'll update it.

deploy-board/deploy_board/webapp/env_views.py
rfleur01: I'll update it to: 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'

deploy-board/deploy_board/webapp/helpers/environs_helper.py
rfleur01: Yes, in the previous state if `get_nimbus_identifier` failed it would surface an error, but not render the page. We want to render the page with a message, which this accomplishes. I looked at the references to this method and it seems safe to allow it to execute

tylerwowen: Can you double check with @osoriano? He added 
`external_id = environs_helper.create_identifier_for_new_stage(request, name, stage)`

osoriano: This request is in the write path. I see it's used for creating new stages.

I believe the Nimbus id is a requirement for creating a new stage, so I don't think we could continue here. Otherwise, it would be possible to create a new stage without a Nimbus ID.

Same applies to the `create_nimbus_identifier` below

rfleur01: Gotcha, the original ticket wanted to prevent Nimbus error from blocking certain pages from rendering, but this seems like maybe this can be the exception since it’s a write path vs rendering something like the environment landing page. @tylerwowen I will revert this change as well as the one on line 105

deploy-board/deploy_board/webapp/helpers/environs_helper.py
```diff
@@ -18,6 +18,11 @@
 import logging
 from deploy_board.webapp.helpers.deployclient import DeployClient
 from deploy_board.settings import IS_PINTEREST
```
tylerwowen, line 20: I think you can just revert all changes in this file and keep it as-is.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -341,14 +341,19 @@ def get(self, request, name, stage=None):
                 project_name_is_default = True if existing_stage_identifier is not None and existing_stage_identifier['projectName'] == "default" else False
```
tylerwowen, line 341:
```diff
diff --git a/deploy-board/deploy_board/webapp/env_views.py b/deploy-board/deploy_board/webapp/env_views.py
index 391f22a1..939a22af 100644
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -784,9 +784,24 @@ def post_create_env(request):
     clone_env_name = data.get("clone_env_name")
     clone_stage_name = data.get("clone_stage_name")
     description = data.get('description')
+
     if clone_env_name and clone_stage_name:
-        common.clone_from_stage_name(request, env_name, stage_name, clone_env_name,
-                                     clone_stage_name, description)
+        try:
+            external_id = environs_helper.create_identifier_for_new_stage(request, env_name, stage_name)
+            common.clone_from_stage_name(request, env_name, stage_name, clone_env_name,
+                                         clone_stage_name, description, external_id)
+        except TeletraanException as detail:
+            if external_id:
+                try:
+                    environs_helper.delete_nimbus_identifier(request, external_id)
+                except:
+                    message = 'Also failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(
+                        external_id, detail)
+                    log.error(message)
+            else:
+                message = 'Failed to create identifier for {}/{}: {}'.format(env_name, stage_name, detail)
+            messages.add_message(request, messages.ERROR, message)
+            raise detail
     else:
         data = {}
         data['envName'] = env_name
@@ -831,16 +846,24 @@ def post_add_stage(request, name):
     if from_stage and from_stage not in stages:
         raise Exception("Can not clone from non-existing stage!")
 
-    external_id = environs_helper.create_identifier_for_new_stage(request, name, stage)
-
-    try:
-        if from_stage:
+    if from_stage:
+        try:
+            external_id = environs_helper.create_identifier_for_new_stage(request, name, stage)
             common.clone_from_stage_name(request, name, stage, name, from_stage, description, external_id)
-        else:
-            common.create_simple_stage(request,name, stage, description, external_id)
-    except:
-        environs_helper.delete_nimbus_identifier(request, external_id)
-        raise
+        except TeletraanException as detail:
+            if external_id:
+                try:
+                    environs_helper.delete_nimbus_identifier(request, external_id)
+                except:
+                    message = 'Also failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(
+                        external_id, detail)
+                    log.error(message)
+            else:
+                message = 'Failed to create identifier for {}/{}: {}'.format(name, stage, detail)
+            messages.add_message(request, messages.ERROR, message)
+            raise detail
+    else:
+        common.create_simple_stage(request, name, stage, description, external_id)
 
     return redirect('/env/' + name + '/' + stage + '/config/')
 
```

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -790,19 +790,22 @@ def post_create_env(request):
...
+            if external_id:
+                try:
+                    environs_helper.delete_nimbus_identifier(request, external_id)
+                except:
+                    message = 'Also failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                    log.error(message)
+            else:
+                message = 'Failed to create identifier for {}/{}: {}'.format(env_name, stage_name, detail)
                 messages.add_message(request, messages.ERROR, message)
+                raise detail
```
tylerwowen, line 808: Should we raise exception in both cases?

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -847,17 +850,27 @@ def post_add_stage(request, name):
...
+            if external_id:
+                try:
+                    environs_helper.delete_nimbus_identifier(request, external_id)
+                except:
+                    message = 'Also failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                    log.error(message)
+            else:
+                message = 'Failed to create identifier for {}/{}: {}'.format(name, stage, detail)
+                messages.add_message(request, messages.ERROR, message)
+                raise detail
```
tylerwowen, line 867: Should we raise exception in both cases?

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -698,14 +698,25 @@ def clone_cluster(request, src_name, src_stage):
     except NotAuthorizedException as e:
         log.error("Have an NotAuthorizedException error {}".format(e))
         if external_id is not None:
-            environs_helper.delete_nimbus_identifier(request, external_id)
+            try:
+                environs_helper.delete_nimbus_identifier(request, external_id)
+            except Exception as detail:
+                message = 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                log.error(message)
+                return HttpResponse(message, status=403, content_type="application/json")
```
tylerwowen, line 706: I wouldn't do this. The original exception `NotAuthorizedException` is not returned to users. Also, 403 is misleading for the message `Failed to delete Nimbus identifier`.

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -698,14 +698,25 @@ def clone_cluster(request, src_name, src_stage):
...
         log.error("Failed to clone cluster env_name: %s, stage_name: %s" % (src_name, src_stage))
         log.error(traceback.format_exc())
         if external_id is not None:
-            environs_helper.delete_nimbus_identifier(request, external_id)
+            try:
+                environs_helper.delete_nimbus_identifier(request, external_id)
+            except Exception as detail:
+                message = 'Failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                log.error(message)
+                return HttpResponse(message, status=500, content_type="application/json")
```
tylerwowen, line 718: Similar here. We are only try to delete nimbus as best effort. It should not cover the original exception.

deploy-board/deploy_board/webapp/env_views.py
tylerwowen: Outside the else block

rfleur01: This method is actually not even being used since the Create Environment button brings you to Ninbus instead of the modal that uses this method. I do imagine if for some reason it is used in the future we would want to raise the Exception to that the env is not created without a Nimbus Id.

deploy-board/deploy_board/webapp/env_views.py
rfleur01: Without raising the execution continues and still creates the stage. That might not be an issue, in which case it does snot need to be raised.

deploy-board/deploy_board/webapp/cluster_view.py
rfleur01: I can remove this. I think it makes more sense to just log the message. The stage still get created but the cluster is not cloned.

deploy-board/deploy_board/webapp/cluster_view.py
rfleur01: same as above. I can remove this. I think it makes more sense to just log the message. The stage still get created but the cluster is not cloned.

deploy-board/deploy_board/webapp/cluster_view.py
tylerwowen: If the user intention is to clone a cluster, I don't think it's useful to create a stage without the cloned cluster.

deploy-board/deploy_board/webapp/env_views.py
tylerwowen: It is used when a cluster is cloned. i.e. config/capacity/ -> Clone Cluster.

deploy-board/deploy_board/webapp/env_views.py
tylerwowen: The `Exception` could be anything. We should only declare a success when all expectations met. Otherwise it will leave potential bad states that are hard to diagnose and maintain.

tylerwowen (approved):

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -784,9 +789,23 @@ def post_create_env(request):
...
+            if external_id:
+                try:
+                    environs_helper.delete_nimbus_identifier(request, external_id)
+                except:
+                    message = 'Also failed to delete Nimbus identifier {}. Please verify that identifier no longer exists, Error Message: {}'.format(external_id, detail)
+                    log.error(message)
+            else:
+                message = 'Failed to create identifier for {}/{}: {}'.format(env_name, stage_name, detail)
+                messages.add_message(request, messages.ERROR, message)
+        raise detail
```
osoriano, line 808: Hi @rfleur01 @tylerwowen, looks like this `raise` is used outside of the `except` block. Is this intentional or am I missing something?

rfleur01: fix is here: https://github.com/pinterest/teletraan/pull/1098

PR #1064, opened 2022-10-18
put rodimus call to create the cluster at the end to avoid asg creation before teletraan env update

liyaqin1:
We send Rodimus request from Teletraan to create ASG even when the cluster creation is failed on Teletraan side.
So, when user tried to create it again, it always got the error like: the ASG exists.
We have to do all Teletraan side updates and then send Rodimus request to create ASG.

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -82,8 +82,6 @@ class EnvCapacityBasicCreateView(View):
                         log.error("Teletraan does not support user to change %s %s" % (field, cluster_info[field]))
                         raise TeletraanException("Teletraan does not support user to create %s" % s)
 
-            clusters_helper.create_cluster_with_env(request, cluster_name, name, stage, cluster_info)
-
             log.info("Associate cluster_name to environment")
             # Update cluster info
             environs_helper.update_env_basic_config(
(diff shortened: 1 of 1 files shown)


tylerwowen:
I don't think there is a perfect order to fix the problem. We can still ran into situations that Teletraan has all information but ASG is missing in AWS. What we really need is _transaction_.

liyaqin1:
> I don't think there is a perfect order to fix the problem. We can still ran into situations that Teletraan has all information but ASG is missing in AWS. What we really need is _transaction_.

I agree with you. This is a better way before we have transaction, right?
If the user failed in the steps before ASG creation, the users can delete the stage by themselves and then the issue will be solved.
If we have ASG created like in the current solution, we need to delete that on AWS side. And they also need to delete the stage.
If they forget to delete the stage and create the cluster again immediately, we have to delete the ASG again for them and at the same time delete the stage.

liyaqin1:
tested on integ:
when I changed the stage type and then created the cluster, I got error like:
Screen Shot 2022-10-18 at 4 01 13 PM
And on AWS side, the ASG is not created which is expected:
Screen Shot 2022-10-18 at 4 10 33 PM

tylerwowen (approved):
Yeah it's fine for short term work around. Can you also include some Wiki instructions so users know what to expect and how to deal with any failures?

liyaqin1:
> Yeah it's fine for short term work around. Can you also include some Wiki instructions so users know what to expect and how to deal with any failures?

Yes, I wrote a wiki and update the jira ticket with the wiki, thanks.

PR #1066, opened 2022-10-19
Expose maximum remaining capacity of clusters in Teletraan UI

rfleur01:
Background:
We already have capacity info for each placement(subnet). However the information is not easily accessible.

Solution:
create a UI component that shows remaining capacity
show the UI component in some or all of the following places:
Cluster capacity config page. i.e. https://deploy.pinadmin.com/env/rodimus-worker/integ/config/capacity/
- Drop “max” from max remaining capacity and add it as help text under Capacity that is shown by default
Cluster config page
- Drop “max” from max remaining capacity and add it as help text under Placements that is shown by default
Environment/stage page
- On the Capacity button, show it conditionally. When it’s <100, show it and mark it as yellow or red
capacity_config
capacity_view
env


rfleur01:
@yongwen in reference to our slack conversation I added a conditional that check to see if there is a a cluster and conditionally updated placements. Here is the screenshot of my validation: 
image

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -65,7 +65,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
                     showhelp="true" v-on:helpclick="securityZoneHelpClick"></label-select>
                     <securityzone-help v-show="showSecurityZoneHelp" v-bind:data="securityZoneHelpData"></securityzone-help>
                     <placements-select label="Placements" title="Placements" showsubnettype="true" v-bind:selectoptions="placements" v-bind:assignpublicip="assignPublicIP" v-bind:subnettype="subnetType"
-                    showhelp="true" v-on:helpclick="placementsHelpClick" v-on:assignpublicipclick="selectpublicip" v-on:subnetfilterclick="changeSubnetType"></placements-select>
+                    v-bind:remainingcapacity="remainingCapacity" showhelp="true" v-on:helpclick="placementsHelpClick" v-on:assignpublicipclick="selectpublicip" v-on:subnetfilterclick="changeSubnetType"></placements-select>
```
tylerwowen, line 68: nit: line too long

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -418,6 +418,7 @@ <h4 class="panel-title pull-left pointer-cursor">
     var currentCluster = info.currentCluster;
     if (currentCluster!=null){
         var currentPlacements = currentCluster.placement.split(',');
+        var currentPlacementObjs = info.placements.filter(function(placement) {return currentPlacements.includes(placement.id)});
```
tylerwowen, line 421: This doesn't seem to handle placement changes. 
I'd recommend handle the changes in 
```
                $(".chosen-select").chosen({
                        "width": "100%"
                    })
                    .change(function() {
                        capacitySetting.selectedPlacements = $('.chosen-select').val()
                    });
```
Besides, I think the calculation logic can move to `getDefaultPlacement()`, similar to the implementation of `getSimpleList()`.

deploy-board/deploy_board/templates/environs/env_landing.html
```diff
@@ -209,6 +209,7 @@ <h4 class="modal-title">Resume Confirm</h4>
 <script>
 
 var capacityInfo = {{capacity_info | safe}}
+var placements = {{placements | safe}}
```
tylerwowen, line 212: Since the only needed value is the remaining size, it's better to calculate the remaining capacity in the Django App so the page can load faster with less contents to download and compute.

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
rfleur01: To clarify: 
-You recommend moving the calculation to `getDefaultPlacement()`
-Handle changes for remaining capacity in the place changes for selected placements are made

tylerwowen: Yes.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -385,6 +392,7 @@ def get(self, request, name, stage=None):
                 "display_stopping_hosts": DISPLAY_STOPPING_HOSTS,
                 "project_name_is_default": project_name_is_default,
                 "project_info": project_info,
+                "remaining_capacity": json.dumps(remaining_capacity),
```
tylerwowen, line 395: Is `json.dumps()` needed?

rfleur01: Yes, otherwise the Cluster Component does not render properly

tylerwowen: `remaining_capacity` is just a number. Why would it not work?

rfleur01: This issue occurs when there is no Capacity. Since the calculation for remainingCapacity  is done Python it returns None when there is no Cluster. Javascript does not recognize None which then prevents the component from rendering correctly. `json.dumps()` converts None to null.

tylerwowen (approved):

deploy-board/deploy_board/templates/environs/env_landing.html
```diff
@@ -231,7 +231,7 @@ <h4 class="modal-title">Resume Confirm</h4>
             },
             computed: {
                 capacityText: function() {
-                    if (this.remainingCapacity < 100) {
+                    if (this.remainingCapacity < 100 && this.remainingCapacity != null) {
```
tylerwowen, line 234: `this.remainingCapacity < 100 && this.remainingCapacity != null` works in JavaScript but it's a good practice to do the null check first.

rfleur01: updated

tylerwowen (approved):

PR #1068, opened 2022-10-19
Fix `stage_type` handling issue when updating `EnvStage`

tylerwowen:
# Problem
```java
if (origBean.getStage_type() != EnvType.DEFAULT && origBean.getStage_type() != environBean.getStage_type()) {
```
The if statement results in true when the original environment stage has a non-default `stage_type` and the request doesn't intend to change the `stage_type`, which is `null`. 

# Fix
When a request has no intention to change stage type, we should set `stage_type` to the current value. Otherwise the default value will be used.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
@@ -111,8 +111,8 @@ public class EnvStages {
         }
 
         if (environBean.getStage_type() == null) {
-            // Request has no intention to change stage type, so set it to current value to
-            // avoid the default value being used.
+            // Request has no intention to change stage type, so set it to the current value
+            // to avoid the default value being used.
             environBean.setStage_type(origBean.getStage_type());
         } else if (origBean.getStage_type() != Constants.DEFAULT_STAGE_TYPE) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST,
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
@@ -109,9 +109,16 @@ public class EnvStages {
         } catch (Exception e) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, e.toString());
         }
-        if (origBean.getStage_type() != EnvType.DEFAULT && origBean.getStage_type() != environBean.getStage_type()) {
-            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, "Modification of stage type is not allowed!");
+
+        if (environBean.getStage_type() == null) {
+            // Request has no intention to change stage type, so set it to current value to
+            // avoid the default value being used.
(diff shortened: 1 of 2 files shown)


liyaqin1 (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
```diff
@@ -109,9 +109,16 @@ public void update(@Context SecurityContext sc,
         } catch (Exception e) {
             throw new TeletaanInternalException(Response.Status.BAD_REQUEST, e.toString());
         }
-        if (origBean.getStage_type() != EnvType.DEFAULT && origBean.getStage_type() != environBean.getStage_type()) {
-            throw new TeletaanInternalException(Response.Status.BAD_REQUEST, "Modification of stage type is not allowed!");
+
+        if (environBean.getStage_type() == null) {
```
ankilosaurus, line 113: if (environBean.getStage_type() != null && origBean.getStage_type() != EnvType.DEFAULT && origBean.getStage_type() != environBean.getStage_type())

ankilosaurus (changes requested):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
tylerwowen: If you leave environBean's stage_type to null, `normalizeEnvRequest` called in `updateStage` will fill the value with the default value, unintentionally change the value.

ankilosaurus: ah i see. makes sense. so you are implicitly backfilling. can you do at the same place where the default is getting set?
I also think the problem is with the caller. since it's sending PATCH request using PUT API

tylerwowen: Yeah I agree it's the problem with the caller. But unfortunately we don't have a single PATCH and been using PUT APIs as PATCH. It will involve too many changes to do the PATCH conversion. 
I can't do at the same place where the default is getting, `public void updateStage(EnvironBean updateBean, String operator)` as it's a "PUT" style API. 

This is inside `normalizeEnvRequest()`
```java
        if (envBean.getStage_type() == null) {
            envBean.setStage_type(Constants.DEFAULT_STAGE_TYPE);
        }
```

ankilosaurus: ok. I think we can still go ahead with the fix just to mitigate this issue incase someone misuses PUT for PATCH but should submit another fix in Rodimus(caller) to use PUT api in a correct way not necessarily replace with PATCH.

ankilosaurus (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
tylerwowen: Thanks. Just a clarification, all callers in deploy_board use this API in the PATCH fashion. This is not used in Rodimus.

tylerwowen:
Verified in integ.

PR #1077, opened 2022-10-31
Rsinha/bad build notify

rashmi59:
Notify if running on bad build after deployment. Searching by meta info as that seems to be the only unique property in the Tags Database. Tested on integ

Screen Shot 2022-11-03 at 1 25 54 PM
Screen Shot 2022-11-03 at 1 26 17 PM


tylerwowen:
Please revisit `BuildTagsManagerImpl`, especially `getEffectiveBuildTag(List<BuildTagBean> tags, BuildBean build)`. We don't have a 1:1 mapping between builds and tags. This change only depends the exact tag.

tylerwowen (changes requested):
Also please rebase your branch from origin/master

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -152,33 +156,50 @@ String generateMessage(String buildId, EnvironBean envBean, DeployState state, D
         String webLink = deployBoardUrlPrefix + String.format("/env/%s/%s/deploy/",
             envBean.getEnv_name(),
             envBean.getStage_name());
+        
+        BuildTagsManager buildTagsManager = new BuildTagsManagerImpl(this.tagDAO);
```
tylerwowen, line 160: `buildTagsManager` is available in `ServiceContext`. Therefore you don't need `tagDAO` either.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -152,33 +156,50 @@ String generateMessage(String buildId, EnvironBean envBean, DeployState state, D
...
                 webLink);
+            } else {
+                return String.format("%s/%s: %s %s/%s completed successfully. See details <%s>",
+                envBean.getEnv_name(),
+                envBean.getStage_name(),
+                action,
+                buildBean.getScm_branch(),
+                buildBean.getScm_commit_7(),
+                webLink);
+            }
```
tylerwowen, line 183: Can we use different templates so the code is not duplicated?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
tylerwowen: And probably make `buildTagsManager` a member of the class.

rashmi59: I tried using it that way and the code doesnt work when doing that. It doesnt pick the tagBean. I saw elsewhere in the code also being done this way. Hence did it this way.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
rashmi59: do you mean changing the strings

tylerwowen: Like
```java
String template = (tagBean != null && tagBean.getValue() == TagValue.BAD_BUILD) ?
"WARNING: %s/%s: %s %s/%s completed successfully, but running on bad build. See details <%s>" :
"%s/%s: %s %s/%s completed successfully. See details <%s>";
```

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
tylerwowen: Can you add the initialization in `TeletraanServiceContext::setupContext`? It's generally a better pattern and it also makes testing easier.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -46,6 +47,7 @@ public class CommonHandler {
     private EnvironDAO environDAO;
     private BuildDAO buildDAO;
     private AgentDAO agentDAO;
+    private TagDAO tagDAO;
```
tylerwowen, line 50: not needed?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/CommonHandler.java
```diff
@@ -121,10 +124,12 @@ public CommonHandler(ServiceContext serviceContext) {
         buildDAO = serviceContext.getBuildDAO();
         agentDAO = serviceContext.getAgentDAO();
         utilDAO = serviceContext.getUtilDAO();
+        tagDAO = serviceContext.getTagDAO();
```
tylerwowen, line 127: Not needed?

tylerwowen (approved):

PR #1084, opened 2022-11-08
Enforce unique abstract_name and provider_name for Placements

rfleur01:
Definition of Done:

Only unique abstract_names and provider_names are able to be published
Duplicate placements are removed from database
Background:

Currently there is no logic ensuring unique placement provider or abstract names. Therefore people can create multiple placement records for the same subnet with different or same names. In fact, there is no difference choosing any duplicate placements as long as their provider names are properly configured. 
According to cloud-arch there isn't a good reason to allow duplicate subnets references in Teletraan. We are clear to proceed with enforcing uniqueness


rfleur01:
Closing to focus on DB validation.

PR #1085, opened 2022-11-08
Update metric url validation

osoriano:
Users will most likely provide metric urls for
the configured metric store.

Add configuration for specifying the metric url
prefix which will be used to validate metric
urls


deploy-board/deploy_board/settings.py
```diff
@@ -83,6 +83,7 @@
 # Serviceframework add-on vars
 SERVICE_RATELIMIT_CONFIG_URL = os.getenv("SERVICE_RATELIMIT_CONFIG_URL")
 STATSBOARD_API_FORMAT = os.getenv("STATSBOARD_API_FORMAT", "OFF")
+STATSBOARD_API_PREFIX = os.getenv("STATSBOARD_API_PREFIX", None)
```
tylerwowen, line 86:
```python
>>> url =''
>>> url.startswith(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: startswith first arg must be str or a tuple of str, not NoneType
```

osoriano: Yeah may not be great for the default to fail. Updated the default to `https://`

tylerwowen (approved):

PR #1086, opened 2022-11-08
Fix uuid not getting populated in envbean

QiaoqianLin:


tylerwowen:
You might want to rebase or pull from teletraan master.

PR #1094, opened 2022-12-02
Rodimus token remove cache

aagxxi:
Knox keys should be able to be rotated without restarting. Knox already has means to rotate the primary key on a live service, and Knox libraries recommend to access the local storage each time the key is used. So all the caching logic is being removed.


aagxxi:
@robbintt @yongwen This PR needs review. I do not have permissions to change Reviewers, thus this comment tagging you.

PR #1095, opened 2022-12-02
Rodimus token strip newlines

aagxxi:
Teletraan API and Teletraan Worker read Rodimus API keys from Knox. A newline can be inserted into a Knox key and result in a malformed authentication header. Defensive programming should avoid this mistake.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
@@ -74,7 +74,7 @@ public class RodimusManagerImpl implements RodimusManager {
 
     private void setAuthorization() throws Exception {
         String currKnoxKey = new String(this.fsKnox.getPrimaryKey());
-        String rodimusKey = currKnoxKey.replaceAll("[\n\r\t]*$", "");
+        String rodimusKey = currKnoxKey.replaceAll("[\n\r\t]", "");
         this.headers.put("Authorization", String.format("token %s", rodimusKey));
     }
 
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
@@ -74,7 +74,8 @@ public class RodimusManagerImpl implements RodimusManager {
 
     private void setAuthorization() throws Exception {
         String currKnoxKey = new String(this.fsKnox.getPrimaryKey());
-        this.headers.put("Authorization", String.format("token %s", currKnoxKey));
+        String rodimusKey = currKnoxKey.replaceAll("[\n\r]*$", "");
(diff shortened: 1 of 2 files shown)


aagxxi:
@robbintt @yongwen @tylerwowen This PR needs review. I do not have permissions to change Reviewers, thus this comment tagging you.

deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
```diff
@@ -73,40 +72,24 @@ public boolean shouldSkipClass(Class<?> c) {
...
-        }
-    }
-
     private void setAuthorization() throws Exception {
-        if (this.cachedKey == null) {
-            this.refreshCachedKey();
-        }
-        this.headers.put("Authorization", String.format("token %s", this.cachedKey));
+        String currKnoxKey = new String(this.fsKnox.getPrimaryKey());
+        String rodimusKey = currKnoxKey.replaceAll("[\n\r]*$", "");
```
robbintt: Can you please use a non-regex approach? Also, this method is deprecated and moved, so please do not use it.  You can use any non-regex method to replace all of a certain character with ''.

https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/StringUtils.html#replaceAll-java.lang.String-java.lang.String-java.lang.String-

robbintt (changes requested):
Please check the comment, the method used is deprecated and we should not add any regex.

deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
robbintt: From discussion: it is ok to use the regex if you simplify it to only matching the 3 desired characters.

aagxxi: done

deploy-service/common/src/main/java/com/pinterest/deployservice/rodimus/RodimusManagerImpl.java
```diff
@@ -73,40 +72,24 @@ public boolean shouldSkipClass(Class<?> c) {
...
-        {
-            return this.cachedKey != null;
-        }else{
-            return ! prevKnoxKey.equals( this.cachedKey );
-        }
-    }
-
     private void setAuthorization() throws Exception {
-        if (this.cachedKey == null) {
-            this.refreshCachedKey();
```
yongwen, line 93: seems like we remove the Cache handling here, is that mean that we no longer use cache in the new code

aagxxi: yes, we remove the cache on Teletraan side. We relay on the Knox interface to handling key rotation and caching, as recommended by Knox documentation.

tylerwowen:
Outdated

PR #1099, opened 2022-12-06
for basic new cluster page, we get the host types according to arch

liyaqin1:


tylerwowen (approved):
approved for deployment

PR #1100, opened 2022-12-08
Teletraan stage has lingering group in DB after removing capacity

liyaqin1:
tested on this env: https://deploy-integ.pinadmin.com/env/for-test/test4/config/cluster/config/

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -82,6 +82,7 @@ class EnvCapacityBasicCreateView(View):
             'capacity_creation_info': json.dumps(capacity_creation_info)})
 
     def post(self, request, name, stage):
+        ret = 200
         log.info("Post to capacity with data {0}".format(request.body))
         try:
             cluster_name = '{}-{}'.format(name, stage)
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -109,9 +109,13 @@ def post(self, request, name, stage):
 
             return HttpResponse("{}", content_type="application/json")
         except NotAuthorizedException as e:
+            environs_helper.remove_env_capacity(
+                request, name, stage, capacity_type="GROUP", data=cluster_name)
             log.error("Have an NotAuthorizedException error {}".format(e))
             return HttpResponse(e, status=403, content_type="application/json")
         except Exception as e:
+            environs_helper.remove_env_capacity(
+                request, name, stage, capacity_type="GROUP", data=cluster_name)
```
tylerwowen, line 118: I think it would be better if we can do the "roll back" in the `finally` block and control the behavior using a flag such as `remove_env_capacity = True`.
That way if we add new exceptions in the future, it's less likely to make a mistake.

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -106,14 +107,19 @@ def post(self, request, name, stage):
...
             log.error("Have an error {}".format(e))
-            return HttpResponse(e, status=500, content_type="application/json")
+            ret = 500
+        finally:
+            if ret == 200:
+                return HttpResponse("{}", content_type="application/json")
+            else:
+                environs_helper.remove_env_capacity(
+                    request, name, stage, capacity_type="GROUP", data=cluster_name)
+                return HttpResponse(e, status=ret, content_type="application/json")
```
tylerwowen, line 122: Is the exception `e` available in this scope?

liyaqin1: yes

tylerwowen (approved):

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -188,14 +195,19 @@ def post(self, request, name, stage):
...
+            ret = 403
         except Exception as e:
             log.error("Have an error {}", e)
-            return HttpResponse(e, status=500, content_type="application/json")
+            ret = 500
+        finally:
+            if ret == 200:
+                return HttpResponse("{}", content_type="application/json")
+            else:
+                environs_helper.remove_env_capacity(
```
yongwen, line 208: will this remove the cluster?  can you make sure this command will rollback the previous action:  environs_helper.update_env_basic_config and environs_helper.add_env_capacity? 

also, what if the error happen in environs_helper.update_env_basic_config or environs_helper.add_env_capacity, will this call failed unexpectedly.

yongwen: looks like these block of code is exactly from previous method, can we consolidate them into a common function?

liyaqin1: Yes, environs_helper.remove_env_capacity also removes the clusterName in environs.
If error happens during db operations, then we definitely will get an error on UI. After that, everything will be consistent.
For basic and advanced page, the post method has a little bit difference.

PR #1101, opened 2022-12-08
Update changedate on deploy-board.js

rfleur01:
updating changedate so users do not have to clear cache to get latest js - cache busting 

The fix is the changing of the date, which forces chrome to treat deploy-board.js as a new file rather than use the cached version.

Test Plan:
1. Locally, render the cluster configuration page
2. Validate that the page renders correctly
3. Comment out the getRemainingCapacity method in deploy-board.js
4. Change the changedate on the static resource for deploy-board.js in the base.html
5. go to the cluster configuration age
6. Confirm that the page is not render properly and you see a getRemainingCapacity is not defined error in the console

1. Deploy build to integ with the getRemainingCapacity method commented out in deploy-board.js (the goal is to cache a broken version of the page)
2. Go to cluster configuration to cache the page
3. deploy a build with the fixed getRemainingCapacity method with the changedate on the static resource for deploy-board.js in the base.html updated
4. Clear cache and confirm that the cluster configuration page render correctly

### DIFF
--- a/deploy-board/deploy_board/static/js/deploy-board.js
+++ b/deploy-board/deploy_board/static/js/deploy-board.js
@@ -56,14 +56,14 @@ function getUrlParameter(sParam) {
     }
 }
 
-function getRemainingCapacity(capacityInfo, placementList) { 
-    if (!capacityInfo || !placementList) {
-        return Infinity;
-    }
-    let filteredPlacements = capacityInfo.filter(placement => placementList.includes(placement.id));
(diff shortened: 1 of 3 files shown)


liyaqin1:
How to test? Thanks.

rfleur01:
See updated description @liyaqin1. Also discussed with @robbintt that updating the changedate will resolve the cache issue

liyaqin1 (approved):

liyaqin1 (approved):

PR #1109, opened 2023-01-09
Handle multi-line in script config

tylerwowen:
When users try to insert a multi-line script config, by either typing or pasting, prevent that by replacing \r or \n with a space and show a warning banner. 
In order to do so, introduced the "warning" type banner. 

Additionally, added warning if line is too long ( > 256)

https://user-images.githubusercontent.com/8442875/211421982-ea48c4b1-d711-46d6-a715-56e379b34f35.mov

### DIFF
--- a/deploy-board/deploy_board/templates/configs/config_map.tmpl
+++ b/deploy-board/deploy_board/templates/configs/config_map.tmpl
@@ -104,6 +104,15 @@
             $('#resetConfigMapBtnId').removeAttr('disabled');
         });
 
+        $('#mapConfigFormId,#newEntryFormId').on('input', 'textarea', function() {
+            const currentVal = $(this).val();
+            if (currentVal.match(/[\r\n\v]+/g)) {
+                $(this).val(currentVal.replace(/[\r\n\v]+/g, ''));
+                globalNotificationBanner.warning =
+                    "New line (\\r\\n) is not allowed in script configs and removed. Also note that line wraping is enabled.";
+            }
+        });
+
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/templates/configs/config_map.tmpl
```diff
@@ -104,6 +105,22 @@
             $('#resetConfigMapBtnId').removeAttr('disabled');
         });
 
+        $('#mapConfigFormId,#newEntryFormId').on('input', 'textarea', function() {
+            const currentVal = $(this).val();
+            if (currentVal.match(/[\r\n\v]+/g)) {
+                $(this).val(currentVal.replace(/[\r\n\v]+/g, ' '));
+                globalNotificationBanner.warning =
+                    "Multi-line (\\r\\n) is not allowed in script config and thus replaced by SPACE. Also note that line wraping is enabled.";
```
robbintt, line 113: I do not recommend we replace \r or \n, I think it should only warn the customer and not replace the newlines by space if we have existing use cases. If a user tries to update that, then discovers they can no longer use multiline, it could cause a huge headache. Do you know which users are using multiline in their script config (database contents)?

If there are no cases, then it seems safe.

robbintt (approved):
r+ but check comments, ensure no users have multiline in their script configs before merging.

deploy-board/deploy_board/templates/configs/config_map.tmpl
tylerwowen: 2 having this problem and they are both addressed.

PR #1110, opened 2023-01-14
Rsinha/test deploy sentinel

rashmi59:
Teletraan changes to enable launch template migration. The code changes have been made for the following purposes.
1. cluster creation - when we create a cluster, we want to set the useLaunchTemplate parameter for the backend to be able to read the value. Although the backend already handles default if no parameter is passed, but this is the right flow to go about it.
2. cluster update - we want to be able to pass on the useLaunchTemplate obtained on get cluster so that the flag remains consistent on updating cluster for any other changes


deploy-board/deploy_board/templates/configs/new_capacity.html
```diff
@@ -188,6 +188,7 @@ <h4 class="panel-title pull-left pointer-cursor">Basic Settings</h4>
             clusterInfo['capacity'] = Number(this.instanceCount);
             clusterInfo['archName'] = capacityCreationInfo['defaultArch'];
             clusterInfo['baseImageId'] = this.baseImageValue;
+            clusterInfo['useLaunchTemplate'] = false;
```
rashmi59, line 191: let me know if directly putting false here is appropriate, or I should do this in cluster_view.py. Thanks

robbintt: It's better to put it in the clusters_view.py and pass the variable through. I think it will have a permanent value of "true" once the migration is complete, so it is best to put it in the right place for now.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -193,7 +193,7 @@ <h4 class="panel-title pull-left">Capacity</h4>
                 }
             }),
         // TODO: The default host type comes from settings and should probably not be an abstract_name since they are not guaranteed to be unique.
-        selectedHostTypeValue: capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType).id,
+        selectedHostTypeValue: capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType) != undefined ? capacityCreationInfo.hostTypes.find(hostType => hostType.abstract_name === capacityCreationInfo.defaultHostType).id : "",
```
robbintt, line 196: Is it ok for this to be empty?

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -472,6 +472,7 @@ <h4 class="panel-title pull-left pointer-cursor">
                         isSelected: o.id === currentCluster.baseImageId
                     }
                 }),
+                useLaunchTemplate: currentCluster.useLaunchTemplate != null ? currentCluster.useLaunchTemplate : false,
```
robbintt, line 475: What is the case when currentCluster.useLaunchTemplate == null?

robbintt:
Can you please update the PR description to include more comprehensive "Why?" that is a little more specific to the code changes?

deploy-board/deploy_board/templates/configs/new_capacity.html
robbintt: Also, since it is separate in the advanced review, that would reduce it to just 1 location where there is the hardcoded assignment, which will reduce the chance of errors in the future.

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
rashmi59: When we roll back the rodimus side to current master branch which doesnt have the useLaunchTemplate code, we want to default to false.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
rashmi59: the id is string. ideally this situation should never happen as it wasnt even being handled earlier. Let me know what other suggestion you have.

robbintt: alright, might be good to add a log line to detect this, so we can determine any exceptional cases. not sure the importance of this

robbintt (approved):

PR #1116, opened 2023-01-27
abort all update events for the specific image if the cancel button is clicked

liyaqin1:
tested ok locally:
Screenshot 2023-01-27 at 1 37 03 PM
After clicking Cancel, the status/result changed from INIT to ABORTED

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/base_images_events.html
+++ b/deploy-board/deploy_board/templates/clusters/base_images_events.html
@@ -23,21 +23,21 @@
     </div>
     <div class="row">
         {% if tags %}
-        <button class="deployToolTip btn btn-default btn-block" id="demoteBtnId">   
-            <span class="glyphicon glyphicon-circle-arrow-down"></span> Demote
-        </button>
+            <button class="deployToolTip btn btn-default btn-block" id="demoteBtnId">   
+                <span class="glyphicon glyphicon-circle-arrow-down"></span> Demote
+            </button>
         {% else %}
-        <button class="deployToolTip btn btn-default btn-block" id="promoteBtnId">
-            <span class="glyphicon glyphicon-circle-arrow-up"></span> Promote
-        </button>
+            <button class="deployToolTip btn btn-default btn-block" id="promoteBtnId">
+                <span class="glyphicon glyphicon-circle-arrow-up"></span> Promote
+            </button>
         {% endif %}
     </div>
-    {% if cancel == "yes" %}
-    <div class="row">
-        <button class="deployToolTip btn btn-default btn-block" id="cancelBtnId">   
-            <span class="glyphicon glyphicon-remove-circle"></span> Cancel
-        </button>
-    </div>
+    {% if cancellable %}
+        <div class="row">
+            <button class="deployToolTip btn btn-default btn-block" id="cancelBtnId">   
+                <span class="glyphicon glyphicon-remove-circle"></span> Cancel
+            </button>
+        </div>
     {% endif %}
 </div>
 <script>
--- a/deploy-board/deploy_board/templates/clusters/base_images_events.html
+++ b/deploy-board/deploy_board/templates/clusters/base_images_events.html
@@ -78,10 +78,9 @@
         });
     });
     $("#cancelBtnId").click(function () {
-        var image_id = '{{ image_id }}/';
         $.ajax({
             type: 'PUT',
-            url: `${location.protocol}//${location.host}/clouds/cancel_image_update/${image_id}`,
+            url: `${location.protocol}//${location.host}/clouds/cancel_image_update/{{ image_id }}/`,
             dataType: "json",
             beforeSend: function (xhr, settings) {
                 var csrftoken = getCookie('csrftoken')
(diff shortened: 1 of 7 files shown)


deploy-board/deploy_board/templates/clusters/base_images_events.html
```diff
@@ -32,11 +32,15 @@ <h4 class="panel-title pull-left">Base Images Events</h4>
     </button>
     {% endif %}
     </div>
+    <div class="row">
+    <button class="deployToolTip btn btn-default btn-block" id="cancelBtnId">   
+    <span class="glyphicon glyphicon-remove-circle"></span> Cancel
+    </button>
+    </div>
```
tylerwowen, line 39: Needs some formatting.

tylerwowen: Also, I think we should
1. only show this when demote is present. 
2. or better, when demote is present and there is any event is in the INIT state. So if every event is COMPLETED, can't cancel.

deploy-board/deploy_board/templates/clusters/base_images_events.html
```diff
@@ -74,6 +77,25 @@ <h4 class="panel-title pull-left">Base Images Events</h4>
             },
         });
     });
+    $("#cancelBtnId").click(function () {
+        var image_id = '{{ image_id }}/';
+        $.ajax({
+            type: 'PUT',
+            url: `${location.protocol}//${location.host}/clouds/cancel_image_update/${image_id}`,
```
tylerwowen, line 84: Does the following work? If it does, then we don'e need line 81. 
```suggestion
            url: `${location.protocol}//${location.host}/clouds/cancel_image_update/{{ image_id }}/`,
```

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -337,6 +337,14 @@ def demote_image(request, image_id):
     return HttpResponse("{}", content_type="application/json")
 
 
+def cancel_image_update(request, image_id):
+    try:
+        baseimages_helper.cancel_image_update(request, image_id)
+    except Exception as e:
```
tylerwowen, line 343: In Rodimus I had code to return 400 error for bad requests. We should try to handle those client side errors instead of making everything 500. Same applies to promote and demote.

deploy-board/deploy_board/templates/clusters/base_images_events.html
liyaqin1: yes, formatting makes sense.
and in my opinion, we only need to check the events. if there is any event with INIT state, we have cancel button.

deploy-board/deploy_board/templates/clusters/base_images_events.html
liyaqin1: yes, make sense. ^^

liyaqin1:
if there is event with INIT state, we have cancel button; otherwise, we don't.
Before click cancel:
Screenshot 2023-01-27 at 5 10 44 PM
After click cancel:
Screenshot 2023-01-27 at 5 11 01 PM

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -332,6 +334,8 @@ def promote_image(request, image_id):
 def demote_image(request, image_id):
     try:
         baseimages_helper.demote_image(request, image_id)
+    except IllegalArgumentException as e:
```
ntascii, line 337: how do we raise this exception IllegalArgumentException ?

deploy-board/deploy_board/templates/clusters/base_images_events.html
```diff
@@ -22,24 +22,47 @@ <h4 class="panel-title pull-left">Base Images Events</h4>
...
+        </button>
+        {% endif %}
+    </div>
+    {% if cancel == "yes" %}
+    <div class="row">
+        <button class="deployToolTip btn btn-default btn-block" id="cancelBtnId">   
+            <span class="glyphicon glyphicon-remove-circle"></span> Cancel
+        </button>
     </div>
+    {% endif %}
```
tylerwowen, line 41: Indentation

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -377,10 +391,17 @@ def get_base_image_events(request, image_id):
 
     tags = baseimages_helper.get_image_tag_by_id(request, image_id)
 
+    cancel = "no"
```
tylerwowen, line 394: Can we use `True` or `False` here?

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -377,10 +391,17 @@ def get_base_image_events(request, image_id):
...
+    for event in base_images_events:
+        if event['state'] == "INIT":
+            cancel = "yes"
+            break
+
     return render(request, 'clusters/base_images_events.html', {
         'base_images_events': base_images_events,
         'image_id': image_id,
         'tags': tags,
+        'cancel': cancel,
```
tylerwowen, line 404: I think `cancellable` is better

deploy-board/deploy_board/webapp/cluster_view.py
tylerwowen: Yeah we should raise in deploy-board/deploy_board/webapp/helpers/base_client.py

tylerwowen (approved):

ntascii (approved):
lgtm

PR #1119, opened 2023-01-31
Revert host type UI updates to support dropdown search and multiple pages 

rfleur01:
This is not a full revert. 3 out of 4 of the files will be reverted.
File that will not be reverted is from Rodimus: common/src/main/resources/sql/tools/mysql/schema-update-20.sql


robbintt:
Can you add in the description which file was not reverted so it is clear to someone that this isn't a complete revert?

PR #1118, opened 2023-01-30
Deploy board Cluster ami config changes

tylerwowen:
- Refactored `baseimage-select` to support image pinning
- Removed duplicate logics from 3 places: cluster creation, cluster creation advanced, cluster configuration to `base-image-utils.js`. 
- Created `label-select2` component. 
- Created **Base Image Update History** page (ClusterBaseImageHistoryView) to show update events for a cluster.

## Pinned Image
pinned image 2

## Auto update
auto update 2

## Warning
ami_warning2

## Update History
update history

PR #1125, opened 2023-02-02
show all AMIs on cluster configuration page

liyaqin1:
In cluster page when the cluster is already created, there is only the image which is being in use is showing (https://jira.pinadmin.com/browse/CDP-5662)

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -464,7 +464,7 @@
                         }).filter(function(item){return item.name!='assign_public_ip'}),
                 assignPublicIP: initialAssignPublicIP,
                 awsRole: currentCluster.configs.aws_role,
-                baseimages: baseImageSorted.map(function(o) {
+                baseimages: info.baseImages.map(function(o) {
                     var acceptance = o.acceptance ? ' [' + o.acceptance + ']' : '';
                     return {
                         value: o.id,


liyaqin1:
looks good to me.
the AMI list on config page is the same as on creation page:
Screenshot 2023-02-01 at 5 46 18 PM

deploy-board/deploy_board/templates/clusters/cluster_configuration.html
```diff
@@ -464,7 +464,7 @@ <h4 class="panel-title pull-left pointer-cursor">
                         }).filter(function(item){return item.name!='assign_public_ip'}),
                 assignPublicIP: initialAssignPublicIP,
                 awsRole: currentCluster.configs.aws_role,
-                baseimages: baseImageSorted.map(function(o) {
+                baseimages: info.baseImages.map(function(o) {
```
osoriano, line 467: https://github.com/pinterest/teletraan/blob/a164d2e7750b98dd59eec139f0b2036aaae2a74b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html#L445

Looks like this change will no longer filter for accepted images?

liyaqin1: we don’t show accepted images in cluster creation page. So, to keep creation and configuration consistent, I plan to make them the same

liyaqin1: I am not sure why there are some code to filter accepted and sort the AMIs. In creation page, it just show all the AMIs.

liyaqin1:
The pr https://github.com/pinterest/teletraan/pull/1118  fixed this issue already , so abandon this pr

PR #1129, opened 2023-02-09
Host Type Value Update

rfleur01:
Background: selectedHostType value was not update properly which was preventing a cluster update

Solution: add additional code to make sure the value is update when the selector is updated

Test:
1. Navigate to a Cluster Configuration and check the selectedHostType value id in the console
2. Screenshot 2023-02-09 at 2 15 21 PM
3. Update the value in the host type selector and confirm that  the selectedHostType value id in the console has updated
4. Screenshot 2023-02-09 at 2 15 44 PM
5. Save the cluster configuration 
6. 
Screenshot 2023-02-09 at 2 16 03 PM
8. Refresh the page to validate that the cluster configuration was updated 
9. 
Screenshot 2023-02-09 at 2 16 13 PM


tylerwowen (approved):
This works but please create a ticket so we can refactor this later. We should stick with vue and not mingle vue and jQeury.

PR #1130, opened 2023-02-13
set default to launch template instead of launch config

rashmi59:
Setting use launch template false. Tested in dev 1.

To be checkedin only once it is deemed that launch template is stable and working fine in production.


rfleur01:
RE-tested in Dev1

1. Create a cluster
2. Confirm that the Launch Template was created by default in the DB
image
4. Confirm that the Launch Template was created in AWS
image

robbintt (approved):

PR #1145, opened 2023-03-01
Add proxy option for teletraan

kynanlalone:
add HTTP proxy support to the Teletraan service.  Configuration of the proxy is optional and disabled by default.  Tested with a locally listening proxy.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
@@ -22,6 +22,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
+import java.net.Proxy;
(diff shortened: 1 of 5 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/common/HTTPClient.java
```diff
@@ -31,6 +33,34 @@ public class HTTPClient {
...
+        if (Boolean.TRUE.equals(useProxy)) {
+            if (httpProxyAddr == null) {
+                throw new IllegalArgumentException("useProxy was configured but missing required httpProxyAddr");
+            }
+            this.useProxy = useProxy;
+            this.httpProxyAddr = httpProxyAddr;
+            this.httpProxyPort = httpProxyPort;
+        }
+    }
+    public boolean getUseProxy() {
```
tylerwowen, line 54: new line between methods. Would recommend run formatter.

tylerwowen (approved):

PR #1147, opened 2023-03-01
Base images page improvemetns

ntascii:
## Summary 
Some improvements in the base images: 
- Include search AMIs based on image type (such as cmp_base_blah) 
- Fix pagination in base images pages 
- Add new page to filter AMIs based on image type. Currently, it's hard to navigate or find AMIs coming from different types. 
- Add new page to see AMI update events of an AMI based on AMI name instead of its Id. It makes it easier for monitoring during the update process. 
- apply some auto format for the touched files. 

## Test Plan
Tested locally
Search box
Screenshot 2023-03-01 at 10 34 51 AM
Update Events per ami-id 
Screenshot 2023-03-01 at 10 35 17 AM
Filtered by abstract name
Screenshot 2023-03-01 at 10 35 36 AM
Pagination
Screenshot 2023-03-01 at 10 36 08 AM

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/base_images.html
+++ b/deploy-board/deploy_board/templates/clusters/base_images.html
@@ -9,7 +9,32 @@
     <li><a href="/clouds/baseimages/">Base Images</a></li>
 </ul>
 {% endblock %}
+{% block navbar-search %}
+<form class="navbar-form navbar-left" role="search">
+    <div class="form-group has-feedback">
+        <input id="baseImageSearchInputId" type="text" class="form-control" placeholder="Search by Abstract Name...">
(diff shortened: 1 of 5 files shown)


tylerwowen (commented):
I think we should avoid introducing the concept "AMI" and stick with conventions of using "abstract_name" or "provider_name". They are confusing enough, so please don't add another one.

ntascii:
That makes sense, but eventually we should rename the old `provider_name` to avoid confusion. `provider_name` is not the name of cloud provider but name of the ami in the provider. They might be intentionally created that way to ensure the abstraction across multiple cloud providers long time ago. However, our internal terms should align with external concepts. Let me change ami_name to provider_name.

tylerwowen (approved):

PR #1150, opened 2023-03-08
Fix base images list per abstract name when there is no golden image

ntascii:
## Summary 
Check if we have golden image for the region first before tagging 
## Test Plan 
Test locally 
Screenshot 2023-03-08 at 1 27 51 PM

### DIFF
--- a/deploy-board/deploy_board/webapp/cluster_view.py
+++ b/deploy-board/deploy_board/webapp/cluster_view.py
@@ -403,9 +403,8 @@ def get_base_images_by_abstract_name(request, abstract_name):
 
     # add golden tag to images 
     for image in base_images:
-        if golden_images.get(image['cell_name']): 
-            if image['id'] == golden_images[image['cell_name']]['id']:
-                image['tag'] = 'current_golden'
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -403,8 +403,9 @@ def get_base_images_by_abstract_name(request, abstract_name):
 
     # add golden tag to images 
     for image in base_images:
-        if image['id'] == golden_images[image['cell_name']]['id']:
-            image['tag'] = 'current_golden'
+        if golden_images.get(image['cell_name']): 
+            if image['id'] == golden_images[image['cell_name']]['id']:
+                image['tag'] = 'current_golden'
```
tylerwowen, line 408:
```suggestion
            image['tag'] = 'current_golden' if golden_images[image['cell_name']] and image['id'] == golden_images[image['cell_name']]['id']
```

tylerwowen (approved):

tylerwowen (approved):

PR #1151, opened 2023-03-09
Allow pin image as long as it's cmp prefixed

tylerwowen:
This removed the inconsistency between initial cluster setting and fetched data.


ntascii (approved):
lgtm

PR #1158, opened 2023-03-10
Consolidate logic for Pin image

ntascii:
## Summary 
Always show Pin image. That’s the reflection of DB status.
- If there is golden - allow unpin (auto update) and pin
- If there is no golden  - disabled pin image
  - if cluster is pinned, no problem
  - if cluster is unpinned, automatically pin image
## Test plan 
Test locally and dev1
No golden, disable Pin Image checkbox and auto checked.
Screenshot 2023-03-10 at 12 03 37 PM
There is a golden, reflect db status 
 Auto update
Screenshot 2023-03-10 at 12 10 55 AM
Pin Image 
Screenshot 2023-03-10 at 12 09 38 PM
Save and it works.

### DIFF
--- a/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
@@ -124,7 +124,9 @@ var capacitySetting = new Vue({
         baseImageHelpData: [],
         baseImages: mapBaseImagesToOptions(baseImagesSorted),
         baseImageValue: defaultImageId,
-        pinImage: getPinImageValue(goldenImage, true ), // Default auto update if there is a golden image 
+        // Default to Pin image if there is a golden image before GA
+        // After GA, default to Auto Update.
+        pinImage: getPinImageValue(goldenImage, false),
         pinImageEnabled: isPinImageEnabled(goldenImage),
         configOptions: Object.keys(info.configList).map(
             function (key) {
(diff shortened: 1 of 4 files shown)


tylerwowen (changes requested):
Can you please remove formatting changes? There are too many unrelated change and can make code blame difficult.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -88,300 +106,297 @@ <h4 class="panel-title pull-left">Capacity</h4>
...
+                    return {
+                        name: key,
+                        value: info.defaultCMPConfigs[key]
+                    }
+                }).filter(function (item) { return item.name != 'cmp_group' }),
+            assignPublicIP: false,
+            baseImageHelpData: [],
+            baseImages: mapBaseImagesToOptions(baseImagesSorted),
+            baseImageValue: defaultImageId,
+            pinImage: getPinImageValue(goldenImage, true), // Default auto update if there is a golden image
```
tylerwowen, line 146: We should be careful not to publish Golden image util it's GA.

ntascii:
I added the auto formating in the later commit so you can see the first commit for the change.

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
ntascii: Good point. Default to Pin Image until GA will be better.

ntascii:
Updated the default for auto update and reverted the auto format commit for reviewing easier.

tylerwowen (dismissed):

tylerwowen (approved):

PR #1161, opened 2023-03-16
Allow special chars for repo in the api request

liyaqin1:
we have to fetch commit info from github and the repo is `pinternal/...`, so that we have to allow more chars in the request.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Commits.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Commits.java
@@ -49,7 +49,7 @@ public class Commits {
     @GET
     @Timed
     @ExceptionMetered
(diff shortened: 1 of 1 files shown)


liyaqin1:
tested ok on integ instance:
Screenshot 2023-03-16 at 1 50 27 PM

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Commits.java
```diff
@@ -58,6 +58,7 @@ public CommitBean getCommit(
             @ApiParam(value = "Commit's scm type, either github or phabricator", required = false)@PathParam("scm") String scm,
             @ApiParam(value = "Commit's repo", required = true)@PathParam("repo") String repo,
             @ApiParam(value = "Commit SHA", required = true)@PathParam("sha") String sha) throws Exception {
+        repo = repo.replace("%2F", "/");
```
osoriano, line 61: Would be nice to use a library for this instead of decoding ourselves

osoriano (approved):

PR #1170, opened 2023-03-31
Add origin/ before branch if it is a github branch

liyaqin1:
For now, the builds api call only return the commits on phabricator if the branch is specified.
Screenshot 2023-03-31 at 4 01 32 PM
Screenshot 2023-03-31 at 4 01 47 PM
We have to also return the commits on github which has "origin/" before the branch name.


liyaqin1:
tested ok on integ. now the github commits are returned.
Screenshot 2023-03-31 at 4 11 33 PM

liyaqin1:
according to teletraan dri, user are forced to remove origin/ on their side.

PR #1172, opened 2023-04-05
Add warning when Teletraan user tries to save new config

liyaqin1:
Teletraan to display config change and let user confirm when config change is saved

### DIFF
--- a/deploy-board/deploy_board/templates/configs/config_map.tmpl
+++ b/deploy-board/deploy_board/templates/configs/config_map.tmpl
@@ -176,15 +176,30 @@
         });
 
         $('#saveEntryBtnId').click(function () {
-            $('#newEntryModalId').modal('hide');
+            $('#saveEntryModalId').modal('hide');
+            var btn = $('#saveConfigMapBtnId');
             $.ajax({
                 type: 'POST',
                 url: '/env/{{ env.envName }}/{{ env.stageName }}/config/map/',
                 data: $('#mapConfigFormId').serialize(),
                 dataType: 'json',
-                success: function (data) {
-                    window.location.href=window.location.href
+                beforeSend: function () {
+                    btn.button('loading');
                 },
+                success: function (data) {
+                    if(data != null && data.success == false) {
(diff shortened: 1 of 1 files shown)


liyaqin1:
Screenshot 2023-04-04 at 7 20 58 PM

deploy-board/deploy_board/templates/configs/config_map.tmpl
```diff
@@ -42,6 +42,30 @@
...
+            <form id="saveEntryFormId" class="form-horizontal" method="post" role="form">
+                <div class="modal-header">
+                    <button type="button" class="close" data-dismiss="modal"><span
+                            aria-hidden="true">&times;</span><span class="sr-only">Close</span>
+                    </button>
+                    <h4 class="modal-title" id="saveEnvModalLabel">Save New {{ panelTitle }}</h4>
+                </div>
+                <div class="modal-body">
+                    Are you sure you want to <strong>save new config</strong> for this cluster?<br>
+                    Please double check the new added config with red color before saving. Thanks.
```
ntascii, line 58: red color means new entries or error entries? If it's new probably we should have them in different color. Red often means error I guess.

liyaqin1: new entries. what's the best color for new entries on your mind please?

ntascii: not an UX expert but probably blue green or yellow.

liyaqin1:
Current behavior:
After adding, it is not easy to figure out what changed. When saving, there is no confirmation box.

PR behavior:
After adding, the new item is in yellow color. When saving, there is a confirmation box to ask user to double check.

deploy-board/deploy_board/templates/configs/config_map.tmpl
liyaqin1: ok, used yellow ^^

ntascii:
thanks.

ntascii (approved):

PR #1174, opened 2023-04-07
pass scm type in hotfix patch

liyaqin1:
customer got error when loading commits in hotfix patch ui


liyaqin1:
The error is:
image
After the fix, the error is gone.

ntascii (approved):

PR #1176, opened 2023-04-10
Remove optional-1 and optional-2 from changefeed

QiaoqianLin:


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/ConfigHistoryHandler.java
```diff
@@ -44,8 +44,9 @@
 
 public class ConfigHistoryHandler {
     private static final Logger LOG = LoggerFactory.getLogger(ConfigHistoryHandler.class);
-    private static final String CHANGEFEED_TEMPLATE = "{\"type\":\"%s\",\"environment\":\"%s\",\"description\":\"%s\",\"author\":\"%s\","
-            + "\"automation\":\"%s\",\"source\":\"Teletraan\",\"optional-1\":\"%s\",\"optional-2\":\"%s\"}";
+    private static final String CHANGEFEED_TEMPLATE =
+        "{\"type\":\"%s\",\"environment\":\"%s\",\"url\":\"%s\",\"description\":\"%s %s\",\"author\":\"%s\","
+            + "\"automation\":\"%s\",\"source\":\"Teletraan\",\"nimbus_uuid\":\"%s\"}";
```
tylerwowen, line 49: It'd be nice if this can be formatted in the same way it is in the `DeployHandler`.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -100,7 +105,7 @@ public class DeployHandler implements DeployHandlerInterface{
     private Allowlist buildAllowlist;
 
 
-    private final class NotifyJob implements Callable<Void> {
+    public final class NotifyJob implements Callable<Void> {
```
tylerwowen, line 108: Why changing to `public`

tylerwowen (approved):

PR #1178, opened 2023-04-13
DeployBoard syntax error update

rfleur01:
Background:
Previously, for error messages using the external_id from nimbus, there could be situations where it is not intialized and assigned, so updated need to be made to code to ensure that any nimbus issues are correctly logged and surfaced to the user instead of syntax errors. 

Test plan:
Create a stage 
Screenshot 2023-04-12 at 10 12 28 AM
Confirm that you receive two error messages in the Ui and in the logs stating that the stage could not be created and the nimbus id could not be deleted. 
Screenshot 2023-04-12 at 10 12 34 AM
Screenshot 2023-04-12 at 10 12 49 AM

Create a stage by cloning an existing one

Screenshot 2023-04-12 at 10 15 31 AM

Confirm that you receive two error messages in the Ui and in the logs stating that the stage could not be created and the nimbus id could not be deleted. 
Screenshot 2023-04-12 at 10 17 02 AM
Screenshot 2023-04-12 at 10 18 50 AM

### DIFF
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -802,7 +802,6 @@ def post_create_env(request):
 
     if clone_env_name and clone_stage_name:
         try:
-            raise TeletraanException("test") #test
             external_id = environs_helper.create_identifier_for_new_stage(request, env_name, stage_name)
             common.clone_from_stage_name(request, env_name, stage_name, clone_env_name,
                                         clone_stage_name, description, external_id)
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -865,38 +865,34 @@ def post_add_stage(request, name):
     external_id = None
     if from_stage:
         try:
-            raise TeletraanException("test") #test
             external_id = environs_helper.create_identifier_for_new_stage(request, name, stage)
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -859,35 +860,37 @@ def post_add_stage(request, name):
...
             common.clone_from_stage_name(request, name, stage, name, from_stage, description, external_id)
         except TeletraanException as detail:
+            message = 'Failed to create stage {}/{}: {}'.format(name, stage, detail)
+            log.error(message)
+            messages.add_message(request, messages.ERROR, message)
             if external_id:
                 try:
                     environs_helper.delete_nimbus_identifier(request, external_id)
-                except:
+                except TeletraanException as detail:
```
osoriano, line 875: Seems this could slightly change the exception handling behavior. Is it needed?

osoriano (approved):

deploy-board/deploy_board/webapp/env_views.py
rfleur01: I added it to ensure continuity between this line and line 890 and the exception that is going to come from the nimbusclient.py that handles the nimbus Response is going to be a `TeletraanException`, so I think it might be necessary to handle it and output the detail.

PR #1179, opened 2023-04-14
Init database for local dev

knguyen100000010:
Update docker compose file to init database for local dev purpose. To test on devapp,

1. Build
2. Clean up local data: `sudo rm -rf /tmp/.data/deploydb`
3. `docker compose down && docker compose up`
4. Verify deploy service can start successfully.

Without this change, the above steps will result in deploy service could not start because database is not created.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     build: ./deploy-service/teletraanservice
     ports:
       - "8080:8080"
-      - "8081:8081"
+      - "8082:8082"
     command: "/opt/deploy-service/teletraanservice/bin/run.sh -c /opt/deploy-service/teletraanservice/bin/server.yaml"
     environment:
       - TELETRAAN_DB_HOST=mysql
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
     build: ./deploy-service/teletraanservice
     ports:
       - "8080:8080"
-      - "8081:8081"
+      - "8082:8082"
```
knguyen100000010, line 21: Avoid port conflict on devapp

docker-compose.yml
```diff
@@ -37,6 +37,6 @@ services:
       - "/tmp/.data/deploydb:/var/lib/mysql"
       - "./deploy-service/common/src/main/resources/sql:/var/teletraan/sql"
       - "./tools/mysql:/var/teletraan/tools"
+      - "./tools/mysql/docker-init:/docker-entrypoint-initdb.d"
     environment:
-      MYSQL_ROOT_PASSWORD: ""
```
knguyen100000010, line 41: Unnecessary since empty password is allowed.

tools/mysql/docker-init/initdb.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Exit upon any errors
+set -e
+
+cd /var/teletraan/tools
+
+./deploydb.sh -f ../sql/deploy.sql
+./upgrade.sh
```
knguyen100000010, line 9: This script will be invoked by mysql container to create the database and update tables.

docker-compose.yml
robbintt: what is the conflict?

robbintt (approved):
Might not need to worry about the port change

docker-compose.yml
knguyen100000010: When I ran on devapp, port 8081 is already used by other containers. Not a big deal, but changing the port will make it run the 1st time without us manually changing the port.

PR #1183, opened 2023-04-25
Do all the operations after the async call returns

liyaqin1:
Do all the operations after the async call returns

### DIFF
--- a/deploy-board/deploy_board/static/js/components/capacitycomponents.js
+++ b/deploy-board/deploy_board/static/js/components/capacitycomponents.js
@@ -332,7 +332,6 @@ Vue.component("static-capacity-config", {
             this.showImbalanceWarning = this.imbalanceWarning != '';
 
             if (this.terminationLimit !== null) {
-                let terminatingHostCount = 0;
                 async function doAjax(args) {
                     let result;
                     try {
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/static/js/components/capacitycomponents.js
```diff
@@ -345,23 +344,23 @@ Vue.component("static-capacity-config", {
...
-                doAjax().then( (data) => terminatingHostCount = data );
-                //console.log(terminatingHostCount);
-
-                if (-sizeIncrease > this.terminationLimit) {
-                    this.showTerminationError = true;
-                    this.terminationError = getTerminationLimitAlertMessage(false);
-                } else if (-sizeIncrease > this.terminationLimit - terminatingHostCount) {
-                    this.showTerminationError = true;
-                    this.terminationError = getTerminationLimitAlertMessage(true);
+                doAjax().then( (data) => {
```
tylerwowen, line 353: Since you are improving this. I'd suggest a more complete solution.
```suggestion
fetch(url, {
      headers: {
        "Content-Type": "application/json",
      },
    })
.then((terminatingHosts) =>
 {
   ...
 }
).catch(function(error){
  console.log(error);
});;
```

tylerwowen (approved):

PR #1186, opened 2023-04-26
update the api to get the host type lists for asg which is created outside of Teletraan

liyaqin1:
Try to fix a bug for the ASGs created outside of Teletraan.


liyaqin1:
tested ok:
Screenshot 2023-04-26 at 2 41 28 PM

tylerwowen (approved):

PR #1193, opened 2023-05-05
Provide more info on Teletraan UI when building hotfix

liyaqin1:
- Show the name of the hotfix job that to be trigger by Teletraan
- Tell the user what the exception Teletraan gets if something is wrong when triggering the hotfix build

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HotfixStateTransitioner.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HotfixStateTransitioner.java
@@ -122,7 +122,8 @@ public class HotfixStateTransitioner implements Runnable {
                         "&SUFFIX=" + hotBean.getOperator() + "_" + buildBean.getScm_commit_7() +
                         "&HOTFIX_ID=" + hotBean.getId() + "&REPO=" + hotBean.getRepo();
                     // Start job and set start time
-                    jenkins.startBuild(hotBean.getJob_name(), buildParams);
+                    String ret = jenkins.startBuild(hotBean.getJob_name(), buildParams);
+                    hotBean.setError_message("Get Response from Jenkins: " + ret);
                     LOG.info("Starting new Jenkins Job (hotfix-job) for hotfix id {}", hotfixId);
 
                     transition(hotBean);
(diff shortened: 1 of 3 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HotfixStateTransitioner.java
```diff
@@ -76,6 +76,8 @@ void processBatch() throws Exception {
             } catch (Throwable t) {
                 // Catch all throwable so that subsequent job not suppressed, also long error in DB
                 LOG.error("HotfixStateTransitioner failed to process {} " + hotfixId, t);
+                hotBean.setError_message("Get Exception: " + t);
```
ntascii, line 79: Do we show the error to the UI?

ntascii:
With the change in this PR, any changes when a hotfix pipeline is not founded? I recall currently it keeps spinning and finally timeout, that caused the confusion to customers. Can we handle better than that like checking if the hotfix job exists?

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HotfixStateTransitioner.java
liyaqin1: Yea, I think this is the new requirement after that incident.

liyaqin1:
> With the change in this PR, any changes when a hotfix pipeline is not founded? I recall currently it keeps spinning and finally timeout, that caused the confusion to customers. Can we handle better than that like checking if the hotfix job exists?

Now we are showing the jenkins url on the UI no matter the job exists or not. So, user can click the url to figure out by themselves when it is spinning on Teletraan UI.

ntascii:
sgtm

ntascii (approved):

liyaqin1:
tested on dev1. When there is no building, we show the link to Jenkins job so that customers will not get confused.
Screenshot 2023-05-11 at 11 32 15 AM

liyaqin1:
Got exception message on UI after I deployed the backend code to worker node 
Screenshot 2023-05-11 at 5 34 35 PM

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HotfixStateTransitioner.java
```diff
@@ -120,7 +122,9 @@ public void transitionHotfixState(HotfixBean hotBean) throws Exception {
                         "&SUFFIX=" + hotBean.getOperator() + "_" + buildBean.getScm_commit_7() +
                         "&HOTFIX_ID=" + hotBean.getId() + "&REPO=" + hotBean.getRepo();
                     // Start job and set start time
-                    jenkins.startBuild(hotBean.getJob_name(), buildParams);
+                    String ret = jenkins.startBuild(hotBean.getJob_name(), buildParams);
+                    hotBean.setError_message("Get Response from Jenkins: " + ret);
```
mitchellngoodman, line 126: Since "error message" is being used as just a "message" it could be helpful to have a comment explaining that this is not an error but being reused for generic message

mitchellngoodman (approved):

tylerwowen (dismissed):

tylerwowen (approved):

PR #1198, opened 2023-05-19
provide tag options for ami promotion

liyaqin1:
Currently we promote to golden (like golden prod) but the implemetnation should support different golden tags like latest and canary.
This PR is to introduce the tags and rollout to the clusters based on the tag.
Currently:
Click "Promote", tag will be "GOLDEN"
After:
We need 3 tags: GOLDEN_LATEST, GOLDEN_CANARY, GOLDEN
Clicking Promote will bring the options to which stage we want to promote.

### DIFF
--- a/deploy-board/deploy_board/static/datatables/css/dataTables.bootstrap.css
+++ b/deploy-board/deploy_board/static/datatables/css/dataTables.bootstrap.css
@@ -27,6 +27,7 @@ div.dataTables_wrapper div.dataTables_length label {
 div.dataTables_wrapper div.dataTables_length select {
   width: 75px;
   display: inline-block;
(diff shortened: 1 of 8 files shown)


liyaqin1:
tested on dev1, looks good. When I promote to latest, I see the clusters with the corresponding keyword in the name were trying to upgrade; and then I promote to canary, it still works. And then I promote to prod, still looks good.
Screenshot 2023-05-19 at 4 43 03 PM

ntascii (approved):
lgtm

liyaqin1:
After demoting, no tags exists anymore.
Screenshot 2023-05-19 at 5 29 27 PM

PR #1203, opened 2023-06-02
add path when getting github commits

liyaqin1:
For now, when getting github commits, there is no way to provide path.
In this PR, I added path as one param when querying github api.
Also, increase the default size from 30 to 100 for sending metrics to pinalytics for Pinboard services.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
@@ -191,6 +191,7 @@ public class GithubManager extends BaseManager {
         // TODO: Do not RETRY since it will timeout the thrift caller, need to revisit
         Map<String, String> params = new HashMap<String, String>();
         params.put("sha", startSha);
+        params.put("path", path);
 
         setHeaders();
         String jsonPayload = httpClient.get(url, null, params, headers, 1);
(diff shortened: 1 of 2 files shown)


liyaqin1:
Tested ok.
When not providing path, I got more than 40 commits; when providing path api, I only got 10 commits.
I compared the result with the real commits under pinternal/pinboard, it looks correct to me.
Screenshot 2023-06-02 at 11 48 11 AM

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
```diff
@@ -191,6 +191,7 @@ public Queue<CommitBean> getCommits(String repo, String startSha, boolean keepHe
         // TODO: Do not RETRY since it will timeout the thrift caller, need to revisit
         Map<String, String> params = new HashMap<String, String>();
         params.put("sha", startSha);
+        params.put("path", path);
```
ntascii, line 194: Is this a file path? will it change the existing behavior?

ntascii (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
liyaqin1: it is a folder/file path.
it will not change the existing behavior.
path is in the api, but it doesn’t take effect when querying github.
now, it will take effect, if no path provided, it will consider the whole repo.

PR #1206, opened 2023-06-12
Automatically replace leader at the end in a cluster replacement

liyaqin1:
According to this ER ticket https://jira.pinadmin.com/browse/ER-2390, we have to implement a new feature for replacing a special host at the end automatically during a cluster replacement.
Tested on dev1: https://deploy-dev1.pinadmin.com/env/ami_sd_validation/yaqin-test/
First marked a random node to be a leader and then did cluster replacement several times.
Each time the leader was replaced at the end.
It succeeded once.
Other times it time out.
But each time I can see the non leader was terminated first.
I only have two hosts on the dev1 test env.
I will test 3 hosts on integ.
Hopefully it will not timeout on integ.

PR on Rodimus side: https://github.com/pinternal/rodimus/pull/169

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
@@ -59,7 +59,7 @@ public class DBHostDAOImpl implements HostDAO {
     private static final String GET_HOST_BY_ENVID_AND_HOSTID = "SELECT DISTINCT e.* FROM hosts e INNER JOIN groups_and_envs ge ON ge.group_name = e.group_name WHERE ge.env_id=? AND e.host_id=?";
     private static final String GET_HOST_BY_ENVID_AND_HOSTNAME1 = "SELECT hs.* FROM hosts hs INNER JOIN groups_and_envs ge ON ge.group_name = hs.group_name WHERE ge.env_id=? AND hs.host_name=?";
     private static final String GET_HOST_BY_ENVID_AND_HOSTNAME2 = "SELECT hs.* FROM hosts hs INNER JOIN hosts_and_envs he ON he.host_name = hs.host_name WHERE he.env_id=? AND he.host_name=?";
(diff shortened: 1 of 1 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
```diff
@@ -59,7 +59,7 @@ public class DBHostDAOImpl implements HostDAO {
     private static final String GET_HOST_BY_ENVID_AND_HOSTID = "SELECT DISTINCT e.* FROM hosts e INNER JOIN groups_and_envs ge ON ge.group_name = e.group_name WHERE ge.env_id=? AND e.host_id=?";
     private static final String GET_HOST_BY_ENVID_AND_HOSTNAME1 = "SELECT hs.* FROM hosts hs INNER JOIN groups_and_envs ge ON ge.group_name = hs.group_name WHERE ge.env_id=? AND hs.host_name=?";
     private static final String GET_HOST_BY_ENVID_AND_HOSTNAME2 = "SELECT hs.* FROM hosts hs INNER JOIN hosts_and_envs he ON he.host_name = hs.host_name WHERE he.env_id=? AND he.host_name=?";
-    private static final String GET_RETIRED_HOSTIDS_BY_GROUP = "SELECT DISTINCT host_id FROM hosts WHERE can_retire=1 AND group_name=? AND state not in (?,?,?)";
+    private static final String GET_RETIRED_HOSTIDS_BY_GROUP = "SELECT DISTINCT host_id FROM hosts WHERE (can_retire=1 or can_retire=3) AND group_name=? AND state not in (?,?,?) ORDER BY can_retire";
```
knguyen100000010, line 62: Minor format: `can_retire=1 or can_retire=3` should be `can_retire=1 OR can_retire=3`

knguyen100000010 (approved):

ntascii:
curious how to mark a node as a leader? do we need to unmark so that the host can be replaced?

liyaqin1:
> curious how to mark a node as a leader? do we need to unmark so that the host can be replaced?

you can either update the database (hosts table) directly or via this API:
Screenshot 2023-06-14 at 11 06 12 AM
I tested with both methods, both worked as expected. 

Once you marked it as value 3, it will be put at the end of the return result from teletraan to rodimus. It will be replaced automatically at the end.

ntascii (approved):

PR #1208, opened 2023-06-16
Deploy agent wait for first successful puppet run

ntascii:
## Summary 
Deploy agent will wait for successful puppet run before running. It will: 
- Check if the state file exist
- Check with the puppet exit code and last summary run of puppet
This PR also fix the logging issues so that we can see the log when the perquisites don't meet. Before, the log was eaten by the cron setup and usage of print.

Emit metric to statsboard, only when the precondition doesn't meet for monitoring and investigation. 
I will also add a prerun check for telefig binary availability in the next PR

## Test plan 
UTs
`tox` for running all
or 
`bazel test tests:test_utils` to run the new tests. 

Test locally `bazel run deployd:deploy-agent -- --use-facter` with similar tests on a remote host below. 

Deployed on a test host. 
1. Delete state.yaml file 
```
E23:51:46 [deployd.common.utils] Waiting for first puppet run.
I23:51:46 [deployd] Deploy agent cannot start because the prerequisites on puppet and telefig did not meet.
```
2. Changed exit code and last summary of puppet. Deployment output. 
```
I00:01:53 [deployd.common.stats] Successfully send the metric to statsboard
I00:01:53 [deployd.common.utils] Puppet failures: 3
E00:01:53 [deployd.common.utils] Last puppet run was not successful.
I00:01:53 [deployd] Deploy agent cannot start because the prerequisites on puppet and telefig did not meet.
```


deploy-agent/deployd/common/stats.py
tylerwowen, line 1: I think we should use the [PinStatsd](https://github.com/pinternal/pinstatsd) lib like the rest of the file.

deploy-agent/deployd/agent.py
```diff
@@ -112,12 +111,6 @@ def _send_deploy_status_stats(self, deploy_report):
             tags['stage_name'] = self._response.deployGoal.stageName
         if deploy_report.status_code:
             tags['status_code'] = deploy_report.status_code
-        if deploy_report.output_msg: 
-            if check_telefig_unavailable_error(deploy_report.output_msg):
-                tags['error_source'] = DeployErrorSource.TELEFIG
-                tags['error'] = DeployError.TELEFIG_UNAVAILABLE
-            elif deploy_report.output_msg.find("teletraan_config_manager") != -1:
-                tags['error_source'] = DeployErrorSource.TELEFIG
```
tylerwowen, line 120: Why removed these?

deploy-agent/deployd/common/utils.py
```diff
@@ -101,36 +105,109 @@ def uptime():
...
+        return True
+
+    # If failed, double check with puppet last summary
+    puppet_summary = load_puppet_summary(config)
+    puppet_failures = puppet_summary.get('events', {}).get(
+        'failure', None) if puppet_summary else None
+    log.info(f"Puppet failures: {puppet_failures}")
+
+    if puppet_failures != 0:
+        send_statsboard_metric(name='deployd.first_puppet_failed', value=1,
```
tylerwowen, line 197: I think we can send both way not just failure, with a 'result' tag to differentiate.

deploy-agent/deployd/agent.py
ntascii: That check might have a perf issue due to string find and I plan to have a check for telefig in the next PR. However, to deploy this version we will need to remove this check to avoid the issue while running in production.

deploy-agent/deployd/common/utils.py
ntascii: we don't want to send too many requests to statsboard using http calls, we also care for puppet failures only so sending both will not add much value and waste resource.

deploy-agent/deployd/common/stats.py
ntascii: that needs to send to metrics cache and metrics agent sidecar up running to send metrics. In case we don't have deployments, metrics agent cannot run, we cannot send metrics. so we need to send to statsboard directly in this case.

tylerwowen (approved):

PR #1233, opened 2023-08-15
Update client.py to use a default key name instead of using constant string

liyaqin1:


deploy-agent/deployd/client/client.py
```diff
@@ -178,7 +178,7 @@ def _read_host_info(self):
                 self._stage_type_fetched = True
 
             if not self._account_id:
-                ec2_metadata = facter_data.get("ec2_metadata.identity-credentials.ec2.info", None)
+                ec2_metadata = facter_data.get(account_id_key, None)
```
ntascii, line 181: do you need to bump the new version to get the change?

liyaqin1: let me bump to 1.2.45

PR #1237, opened 2023-08-18
Improve handling of host and group mapping

knguyen100000010:
"If a host belonging to an autoscaling group, it must alway be mapped to the corresponding group."

If this is true, this is an improvement to the mapping procedure when host pings Teletraan.

This improvement is to fix the cases where, for some reasons, deploy-agent doesn't include the autoscaling group name in `groups`. When this is the case, a new host will stuck in `PROVISIONED` state forever.


ntascii (approved):

ntascii:
might need to rebase the branch

PR #1241, opened 2023-08-25
hide aws account id if there is no content

liyaqin1:
currently, the ui shows none when there is no account id


liyaqin1:
tested with some test data and it looks good to me:
if we don't have account id, there is no column for it:
Screenshot 2023-08-25 at 11 12 57 AM
if there is account id, we show it:
Screenshot 2023-08-25 at 11 13 14 AM

ntascii (approved):

PR #1242, opened 2023-08-25
Consider the situation that accountId is string null

liyaqin1:
The account id is default to be NULL in the database. When a new host is up, if the account is not a sub account, it will be updated as string "null".
This PR fixed two issues:
if a new host is updated the account_id as “null” , it still shows the column on UI;
if a host in the agents table but not in hosts table, we got error on UI.

### DIFF
--- a/deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
+++ b/deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
@@ -23,7 +23,7 @@
                         <i class="fa fa-fw {{ agentStat | agentIcon }}"></i>
                         </a>
                         {% elif report.showMode == "primaryAcct" %}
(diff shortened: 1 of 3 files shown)


robbintt (dismissed):

deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
```diff
@@ -32,7 +32,7 @@
                         </a>
                         {% endif %}
                         {% elif report.showMode == "subAcct" %}
-                        {% if agentStat.agent.accountId and agentStat.agent.accountId != "998131032990" %}
+                        {% if agentStat.agent.accountId and agentStat.agent.accountId != "998131032990" and agentStat.agent.accountId != "null" %}
```
ntascii, line 35: So None in python gets converted to "null" in httml template?

ntascii (approved):

ntascii (approved):

PR #1256, opened 2023-09-13
Publish build events to event bridge

knguyen100000010:
Publish an event to Event Bridge when a build is created or deleted.

Using async calls and log errors (if any) to not disturbing the build publishing API.

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/AwsFactory.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/config/AwsFactory.java
@@ -1,7 +1,6 @@
 package com.pinterest.teletraan.config;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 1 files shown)


ntascii (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/events/EventBridgePublisher.java
```diff
@@ -0,0 +1,54 @@
...
+    PutEventsRequestEntry entry = PutEventsRequestEntry.builder()
+        .eventBusName(eventBusName)
+        .source(TELETRAAN_SOURCE)
+        .detail(buildEventDetailJson(buildBean, action))
+        .detailType(DETAIL_TYPE)
+        .build();
+
+    try {
+      eventBridgeAsyncClient.putEvents(e -> e.entries(entry).build()).whenCompleteAsync((response, err) -> {
+        if (err != null) {
```
ntascii, line 39: any retry when there is an error?

ntascii (approved):

PR #1258, opened 2023-09-15
Update deploy agent logging

osoriano:
Contains changes to improve debuggability such as:

- Update error logging

  In case of tarball extraction failures, log the exception and stack
  trace. These exceptions could cause the host to become stuck. The stack
  trace can help with debugging.

- Update checksum verification warning

  When the sha1 reuqest for a build returns non-200, the verification is
s. kipped.

  Since this happens often and is usually expected, switch to
  `log.warning`.

  Also, update the warning message to clarify the sha1 verification is
  skipped

- Add `S` flag to curl

  From the curl documentation:

    -S, --show-error
           When used with -s, --silent, it makes curl show an error message if it fails.

  Add the flag to show the error in case of failures, which an be helpful
  for debugging

Testing done:
  Ran the downloader command locally. Verified it succeeded.
  Also tested the error case by modifying the extracted dir and re-running the downloader. Ensured stack trace was logged

### DIFF
--- a/deploy-agent/deployd/common/status_code.py
+++ b/deploy-agent/deployd/common/status_code.py
@@ -3,9 +3,9 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#  
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


ntascii (approved):
lgtm

PR #1270, opened 2023-09-28
Introduce the Teletraan universal library

tylerwowen:
Created a new library that includes 2 components: metrics and events 
1. Port changes from Rodimus.
2. Improved `events` package to address some issue and be more generic.

## Tests
Validations are done in the stacked PRs. #1276 & #1277

### DIFF
--- a/deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/events/GenericEventPublisher.java
+++ b/deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/events/GenericEventPublisher.java
@@ -1,0 +1,46 @@
+package com.pinterest.teletraan.universal.events;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import lombok.Getter;
+import reactor.core.publisher.Flux;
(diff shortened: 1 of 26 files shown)


deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/events/GenericEventPublisher.java
```diff
@@ -0,0 +1,46 @@
...
+            EmitResult result = eventsSink.tryEmitNext((AppEvent) event);
+            if (result.isFailure()) {
+                LOG.error("Failed to publish event, cause: " + result);
+            }
+        } catch (Exception ex) {
+            LOG.error("Failed to publish event", ex);
+        }
+    }
+
+    // for testing
```
ntascii, line 39: do we still need this?

tylerwowen: Yes `terminate()` is used in test cases

ntascii (approved):

PR #1271, opened 2023-09-29
Add comment to reported success or failure

killertiger:


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/SimpleAgentJanitor.java
tylerwowen, line 1: AgentJanitor inherits these methods so some double counting.
Actually for us, only AgentJanitor matters.

killertiger: Thanks Tyler, that makes sense.
I removed the success report from AgentJanitor, since it is already reported by SimpleAgentJanitor

PR #1282, opened 2023-10-05
Update agent when health status changes

ntascii:
## Summary 
Update agent when there is a container health change 

## Test plan 
Test on dev1. tbu


gzpcho (approved):
lgtm

PR #1284, opened 2023-10-05
Clean up py2 remnants for loading test data

gzpcho:


gzpcho:
Using absolute imports is recommended and supported in both py2 and py3

ntascii (approved):

PR #1296, opened 2023-10-12
Upgrade deploy board to py3

ntascii:
### Summary 
Deploy board is still in py2, this is the last py2 codebase in our systems. Upgrade to py3 so that we can have a better codebase and we can have security updates moving forward
### Test plan
Tested locally and on dev1 
Together with configs change.


deploy-board/deploy_board/webapp/auth.py
```diff
@@ -19,11 +19,14 @@
...
-try:
-    import urllib2 as http
-except ImportError:
-    from urllib import request as http
+import base64  
+
+# try:
+#     import urllib2 as http
+# except ImportError:
+#     from urllib import request as http
```
gzpcho, line 27: nit: remove commented code

gzpcho:
Can we use absolute imports instead of relative?

deploy-board/deploy_board/webapp/auth.py
```diff
@@ -224,20 +227,21 @@ def handle_oauth2_response(self, code, state_with_data, **kwargs):
...
-        scope = unicode(self.scope)
+        scope = str(self.scope)
 
         # generate and set state
         state = self.oauth_handler.state_generator(**kwargs)
         self.oauth_handler.state_setter(state, **kwargs)
 
         # hack to add data to state
-        state_with_data = state + json.dumps(data).encode('base64')
+        encoded_data = base64.b64encode(json.dumps(data).encode('utf-8')).decode('utf-8')
```
gzpcho, line 243: This is decoded as per `.decode('utf-8')`, no? Variable is called `encoded_data`. Not sure which one is the intended

deploy-board/deploy_board/webapp/capacity_views.py
```diff
@@ -48,7 +48,9 @@ def get(self, request, name, stage):
             if basic_cluster_info:
                 base_image_id = basic_cluster_info.get('baseImageId')
                 base_image = baseimages_helper.get_by_id(request, base_image_id)
-                asg_cluster = autoscaling_groups_helper.get_group_info(request, cluster_name)
+                
+                # asg_cluster = autoscaling_groups_helper.get_group_info(request, cluster_name) NO COMMIT
```
gzpcho, line 52: nit: remove commented code

gzpcho (changes requested):
Thank you for this migration! Left a few nitpicks and questions

deploy-board/deploy_board/webapp/auth.py
ntascii: encode with base64 and convert to a `str`.

deploy-board/deploy_board/webapp/capacity_views.py
ntascii: yeah, forgot to remove some that I put few months ago.

gzpcho (approved):

ntascii:
> nit: thoughts on absolute imports vs relative? 
just converted some existing py2 imports using relative import in py3. We can convert some relative to absolute for better readability. There are cases where relative one is more convenient though. Most of the time, I would prefer absolute if it's not too lengthy.

PR #1297, opened 2023-10-12
Allow custom metrics name

tylerwowen:
Update the naming convention so when a metric name starts with `custom.`, no naming convention will be applied.


killertiger (approved):
LGTM

PR #1299, opened 2023-10-13
Remove commit tag to reduce metrics cardinality

tylerwowen:
Log the commit instead
```
INFO  [2023-12-07 20:06:37,612] com.pinterest.deployservice.handler.CommonHandler: Successfully sent deploy event: com.pinterest.deployservice.events.DeployEvent@2862ea99[commit=7079487,env=sample-service-1,operator=<null>,stage=canary,operator=<null>,resource=deployed_build,tags={env,sample-service-1,stage,canary},timestamp=1701979597612]
```

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/events/DeployEvent.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/events/DeployEvent.java
@@ -13,7 +13,7 @@ public class DeployEvent extends ResourceChangedEvent {
     private final String operator;
 
     public DeployEvent(Object source, String env, String stage, String commit, String operator) {
-        super("deployed_build", operator, source, ImmutableMap.of("env", env, "stage", stage, "commit", commit));
+        super("deployed_build", operator, source, ImmutableMap.of("env", env, "stage", stage));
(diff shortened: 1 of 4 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/events/DeployEvent.java
```diff
@@ -13,7 +13,7 @@ public class DeployEvent extends ResourceChangedEvent {
     private final String operator;
 
     public DeployEvent(Object source, String env, String stage, String commit, String operator) {
```
osoriano, line 15: Could potentially remove the `commit` param/field since it seems unused now

osoriano (dismissed):

deploy-service/common/src/main/java/com/pinterest/deployservice/events/DeployEvent.java
tylerwowen: Good point. Added logging so that information isn't lost

osoriano (approved):

osoriano (approved):

PR #1300, opened 2023-10-14
go through all rows to figure out the account id

liyaqin1:
[CDP-7084](https://jira.pinadmin.com/browse/CDP-7084)

### DIFF
--- a/deploy-board/deploy_board/webapp/host_views.py
+++ b/deploy-board/deploy_board/webapp/host_views.py
@@ -74,8 +74,9 @@ def get_host_id(hosts):
     return None
 
 def get_account_id(hosts):
-    if hosts:
-        return hosts[0].get('accountId')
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -74,8 +74,9 @@ def get_host_id(hosts):
     return None
 
 def get_account_id(hosts):
-    if hosts:
-        return hosts[0].get('accountId')
+    for host in hosts:
+        if host and host.get('accountId') and host.get('accountId') not in {'NULL', 'null'}:
```
gzpcho, line 78:
```python
if host and host.get('accountId') and {'NULL', 'null'} not in host.get('accountId'):
```

gzpcho (approved):

PR #1311, opened 2023-10-20
Emit metrics about deploys and hosts

tylerwowen:
Created a worker to emit deploys and hosts metrics.


deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBDeployDAOImpl.java
```diff
@@ -79,7 +80,9 @@ public class DBDeployDAOImpl implements DeployDAO {
             "AND NOT EXISTS (SELECT 1 FROM environs WHERE environs.deploy_id = deploys.deploy_id) "
             + "ORDER BY last_update ASC LIMIT ?";
     private static final String COUNT_DAILY_DEPLOYS =
-        "SELECT COUNT(*) FROM deploys WHERE DATE(FROM_UNIXTIME(start_date*0.001)) = CURDATE()";
+        "SELECT COUNT(*) FROM deploys WHERE start_date*0.001 >= UNIX_TIMESTAMP(CURDATE())";
```
tylerwowen, line 83: Since `start_date` is indexed, this optimization could save 50% of execution time.

liyaqin1 (approved):

PR #1322, opened 2023-10-30
Add Golden tags for base image list filtered by abstract name

liyaqin1:
The tags for base image list are correct. The tags for base image list filtered by abstract name are not correct. Using this PR, the tags are correct now for list with abstract name.
Tested on dev1.


liyaqin1:
Tested on dev1 and it looks good to me:
Screenshot 2023-10-30 at 5 09 04 PM
I checked all the images on the page on image list page without abstract name, looks consistent.

ntascii (approved):

PR #1324, opened 2023-11-02
keep env_status file and add fields in the report section to do auto redeploy

liyaqin1:
We may have trouble by deleting a service in file env_status. In this PR, on deployd side, we set RESET_BY_SYSTEM to trigger redeploy just like Teletraan UI did by clicking "Retry".

### DIFF
--- a/deploy-agent/deployd/agent.py
+++ b/deploy-agent/deployd/agent.py
@@ -133,12 +133,11 @@ class DeployAgent(object):
                         status.report.redeploy = 0
                     healthStatus = get_container_health_info(status.build_info.build_commit, status.report.envName, status.report.redeploy)
                     if healthStatus:
-                        
                         if "redeploy" in healthStatus:
                             status.report.redeploy = int(healthStatus.split("-")[1])
                             status.report.wait = 0
                             status.report.state = "RESET_BY_SYSTEM"
-                            status.report.containerHealthStatus = "unhealthy and redeploying"
+                            status.report.containerHealthStatus = None
                         else:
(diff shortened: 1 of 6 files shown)


ntascii:
we might need a different state to classify between Manual reset from user (RESET) and auto RESET by bad health.

deploy-agent/deployd/agent.py
```diff
@@ -125,30 +125,37 @@ def serve_build(self):
             self._executor = Executor(callback=PingServer(self), config=self._config)
         # include healthStatus info for each container
         if len(self._envs) > 0:
-            need_to_delete = []
+            #need_to_delete = []
```
ntascii, line 128: remove commented code?

liyaqin1: done

liyaqin1:
> we might need a different state to classify between Manual reset from user (RESET) and auto RESET by bad health.

sounds good, thanks

liyaqin1:
Tested on my devapp and everything on deployd side looks correct.
When manually triggering a bad container, the redeploy times increased by 1 and the state is RESET_BY_SYSTEM:
Screenshot 2023-11-03 at 9 49 44 PM
On TAS side, since my devapp is not really having environments on dev1, it is not possible to trigger the redeployment.
But if I manually edit the state to be RESET_BY_SYSTEM for an existing stage on dev1, I can see it goes to the path as expected:
Screenshot 2023-11-03 at 9 55 45 PM

ntascii (approved):

PR #1325, opened 2023-11-02
Allow agent state to be set as RESET by deployd like Teletraan UI does

liyaqin1:
Corresponding PR on deployd side: https://github.com/pinterest/teletraan/pull/1324 and https://github.com/pinterest/teletraan/pull/1329

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/AgentState.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/AgentState.java
@@ -24,6 +24,8 @@ package com.pinterest.deployservice.bean;
  *      Agent being paused by user manually
  * RESET:
  *      Agent should retry last failure
+ * REDEPLOY:
+ *      Agent should redeploy according to deployd
  * DELETE:
  *      Agent should delete its status file
  * UNREACHABLE:
@@ -37,6 +39,7 @@ public enum AgentState {
     PAUSED_BY_SYSTEM,
     PAUSED_BY_USER,
     RESET,
+    REDEPLOY,
     DELETE,
     UNREACHABLE,
     STOP
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/AgentState.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/AgentState.java
@@ -24,8 +24,8 @@ package com.pinterest.deployservice.bean;
  *      Agent being paused by user manually
  * RESET:
  *      Agent should retry last failure
- * REDEPLOY:
- *      Agent should redeploy according to deployd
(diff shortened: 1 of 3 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/bean/AgentState.java
```diff
@@ -24,6 +24,8 @@
  *      Agent being paused by user manually
  * RESET:
  *      Agent should retry last failure
+ * REDEPLOY:
```
ntascii, line 27: RESET_BY_SYSTEM? same with PAUSED_BY_SYSTEM

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -612,7 +622,15 @@ void process(String envId, EnvironBean env, PingReportBean report, AgentBean age
                 // Special case when agent state is RESET, start from beginning
                 if (updateBean.getState() == AgentState.RESET) {
```
ntascii, line 623: || updateBean.getState() == AgentState.RESET

` LOG.debug("GoalAnalyst case 1.0.2 - host {} work on the same deploy {}, but agent state is {}`

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/AgentState.java
liyaqin1: sounds good, thanks

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
liyaqin1: sounds good, thanks

ntascii (approved):

liyaqin1:
After deploying this PR to TAS, I tested these cases:
1. with deploy-agent-pex current version 1.2.46, tested these UI operations: reset, restart, deploy and launch a new host;
2. with deploy-agent-pex latest version 1.2.52, tested: reset, restart and deploy;
3. with deploy-agent-pex latest version 1.2.52, tested: auto redeploy

For auto redeploy, when the max retry time is hit, there is no redeployment:
Screenshot 2023-11-06 at 6 01 40 PM
Screenshot 2023-11-06 at 6 02 35 PM

when the max retry time is not hit, there is auto redeployment:
before redeployment:
Screenshot 2023-11-06 at 6 03 40 PM
after redeployment:
Screenshot 2023-11-06 at 6 06 06 PM
Screenshot 2023-11-06 at 6 07 52 PM

ntascii (approved):

PR #1333, opened 2023-11-13
Remove duplicated call

knguyen100000010:
Instance refresh resume call has just been added on server side [here](https://github.com/pinternal/rodimus/blob/c16c142bd383fcf8e74dd9bdf50c5f0570ae8fcd/rodimusservice/src/main/java/com/pinterest/rodimus/resource/ClusterReplacements.java#L84-L86).

No need to call it on client side.

When users start an instance refresh, Instance Refresh process may have been paused, so we need to resume it first.


gzpcho (approved):
lgtm

PR #1336, opened 2023-11-14
Remove unneccessary site metrics

ntascii:
## Summary 
The old statsboard links are broken due to envoy migration. The section also doesn't bring much value. We might bring the embeded dashboard directly later on.  

## Test plan 
http://dev-anhn:8080/


gzpcho (approved):
lgtm!

PR #1341, opened 2023-11-16
Update host_tags database table when the host has deployment constraint

liyaqin1:
If there are tags update after Teletraan worker is done for the deployment constraint, the tags in host_tags db table will become stale. Using tags sent by deployd agent on the host every minute, we are sure the host_tags is update to date.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
@@ -547,18 +547,14 @@ public class GoalAnalyst {
         }
 
         // update host_tags table
-        LOG.debug("ec2Tags: {}", ec2Tags);
         if (ec2Tags != null && env != null && env.getDeploy_constraint_id() != null) {
             String constraintId = env.getDeploy_constraint_id();
-            LOG.debug("constraint id: {}", constraintId);
             DeployConstraintBean deployConstraintBean = deployConstraintDAO.getById(constraintId);
             String tagName = deployConstraintBean.getConstraint_key();
-            LOG.debug("tag name: {}", tagName);
             ObjectMapper mapper = new ObjectMapper();
             Map<String, String> tags = mapper.readValue(ec2Tags, Map.class);
             if (tags.containsKey(tagName)) {
                 String tagValue = tags.get(tagName);
-                LOG.debug("tag value: {}", tagValue);
                 HostTagBean hostTagBean = hostTagDAO.get(host_id, tagName);
                 if (hostTagBean == null) {
                     hostTagBean = new HostTagBean();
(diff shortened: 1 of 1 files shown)


liyaqin1:
Tested on dev1.
Before adding a deployment constraint:
```
mysql> select deploy_constraint_id from environs where env_name="deploy-open-sentinel-dev1" and stage_name="test-timeout";
+----------------------+
| deploy_constraint_id |
+----------------------+
| NULL                 |
+----------------------+
1 row in set (0.00 sec)

mysql> select * from host_tags;
Empty set (0.01 sec)
```
After adding a constraint:
Screenshot 2023-11-17 at 2 26 12 PM

```
mysql> select deploy_constraint_id from environs where env_name="deploy-open-sentinel-dev1" and stage_name="test-timeout";
+------------------------+
| deploy_constraint_id   |
+------------------------+
| RCMSzYtDR2GO4FAD2CbX_A |
+------------------------+
1 row in set (0.00 sec)

mysql> select * from deploy_constraints where constraint_id="RCMSzYtDR2GO4FAD2CbX_A";
+------------------------+----------------+-----------------+--------------+------------+---------------+---------------+
| constraint_id          | constraint_key | constraint_type | max_parallel | state      | start_date    | last_update   |
+------------------------+----------------+-----------------+--------------+------------+---------------+---------------+
| RCMSzYtDR2GO4FAD2CbX_A | Name           | GROUP_BY_GROUP  |            1 | PROCESSING | 1700259626653 | 1700259626728 |
+------------------------+----------------+-----------------+--------------+------------+---------------+---------------+
1 row in set (0.00 sec)

mysql> select * from host_tags;
+---------------------+------------------------+----------+-------------+---------------+
| host_id             | env_id                 | tag_name | tag_value   | create_date   |
+---------------------+------------------------+----------+-------------+---------------+
| i-0dd4b6ce5062c36f0 | 8ss-6DLmQeWe7EMejOVWEw | Name     | dev-yaqinli | 1700259626786 |
+---------------------+------------------------+----------+-------------+---------------+
1 row in set (0.00 sec)

mysql> 
```

liyaqin1:
We can definitely update the table via deployd pingRequest without Teletraan worker. As we can see in the database, 10 minutes later after the deployment constraint is added, the worker is still not done with the host_tags table update. Via this new method, host_tags is updated immediately after the constraint is added.
```
mysql> select * from deploy_constraints where constraint_id="RCMSzYtDR2GO4FAD2CbX_A";
+------------------------+----------------+-----------------+--------------+------------+---------------+---------------+
| constraint_id          | constraint_key | constraint_type | max_parallel | state      | start_date    | last_update   |
+------------------------+----------------+-----------------+--------------+------------+---------------+---------------+
| RCMSzYtDR2GO4FAD2CbX_A | Name           | GROUP_BY_GROUP  |            1 | PROCESSING | 1700259626653 | 1700259909105 |
+------------------------+----------------+-----------------+--------------+------------+---------------+---------------+
1 row in set (0.00 sec)

mysql> 

```

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -532,6 +546,26 @@ void process(String envId, EnvironBean env, PingReportBean report, AgentBean age
             }
         }
 
+        // update host_tags table
+        if (ec2Tags != null && env != null && env.getDeploy_constraint_id() != null) {
```
ntascii, line 550: we might want to check if we need to update hostTag, if there is no difference we can just skip updating to avoid db writes.

liyaqin1: A very good point

liyaqin1: updated

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -640,7 +640,7 @@ public PingResult ping(PingRequestBean pingRequest, boolean rate_limited) throws
 
         // Now we have all the relevant envs, reports and agents, let us do some
         // analysis & pick the potential install candidate and uninstall candidate
-        GoalAnalyst analyst = new GoalAnalyst(deployDAO, environDAO, hostName, hostId, envs, reports, agents);
+        GoalAnalyst analyst = new GoalAnalyst(deployConstraintDAO, hostTagDAO, deployDAO, environDAO, hostName, hostId, envs, reports, agents, ec2Tags);
```
ntascii, line 643: instead of having the update in GoalAnalyst, can we have a separate method for updateHostTags() similar to updateHostStatus and updateHosts

liyaqin1: nope, we need to update per envId. And the envId logic is in GoalAnalyst instead of PingHandler

ntascii (approved):

PR #1348, opened 2023-11-21
Add account id validator

ntascii:
## Summary 
Added account allow list to avoid deployments in a strange account. List of allowed account will be specified in teletraanconfigs.  

## Test plan 
Deployed to dev1 and checked logs for a test deployment. 
```
{"level":"INFO","logger":"com.pinterest.deployservice.common.MetricsDataFactory","thread":"dw-21 - POST /v1/system/ping","message":"account id 998131032990 allow lists [998131032990, 562567494283]","timestamp":"2023-11-21T09:00:30.414+0000"}
```

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
@@ -16,6 +16,7 @@
 package com.pinterest.deployservice;
 
 
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -476,7 +477,7 @@ public class ServiceContext {
         return accountAllowList;
     }
 
-    public void setAccountAllowList(List<String> accountAllowList) {
+    public void setAccountAllowList(Collection<String> accountAllowList) {
         this.accountAllowList = new HashSet<String>(accountAllowList);
     }
 }
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
@@ -16,6 +16,11 @@
 package com.pinterest.deployservice;
 
 
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+
+import org.apache.commons.dbcp.BasicDataSource;
+
 import com.pinterest.deployservice.allowlists.Allowlist;
 import com.pinterest.deployservice.buildtags.BuildTagsManager;
 import com.pinterest.deployservice.chat.ChatManager;
@@ -48,11 +53,6 @@ import com.pinterest.deployservice.pingrequests.PingRequestValidator;
 import com.pinterest.deployservice.rodimus.RodimusManager;
 import com.pinterest.deployservice.scm.SourceControlManagerProxy;
 import com.pinterest.teletraan.universal.events.AppEventPublisher;
-
-import org.apache.commons.dbcp.BasicDataSource;
-
-import java.util.List;
-import java.util.concurrent.ExecutorService;
 
 public class ServiceContext {
     private BasicDataSource dataSource;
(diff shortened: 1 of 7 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/pingrequests/Ec2InstanceValidator.java
```diff
@@ -1,25 +1,44 @@
...
+                    bean.getHostId() == null ? "null" : bean.getHostId());
+        }
+
+        // Validate account id of EC2 instance
+        if (accountAllowList == null) {
+            return;
+        }
+
+        String accountId = bean.getAccountId();
+        if (accountId == null || accountId.isEmpty()) {
```
tylerwowen, line 33:
```suggestion
        if (StringUtils.isNotBlank(accountId)) {
```

deploy-service/common/src/main/java/com/pinterest/deployservice/pingrequests/Ec2InstanceValidator.java
```diff
@@ -1,25 +1,44 @@
...
+            return;
+        }
+
+        if (!accountAllowList.contains(accountId)) {
+            LOG.warn("Ignore ping request from host {} with unknown account id {}", bean.getHostName(), accountId);
+            throw new DeployInternalException(
+                    String.format("Host id %s is not from the account allow list"),
+                    bean.getHostId() == null ? "null" : bean.getHostId());
```
tylerwowen, line 41:
```suggestion
                    StringUtils.defaultString(bean.getHostId()));
```

deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
```diff
@@ -102,6 +102,7 @@ public class ServiceContext {
     private Long agentCountCacheTtl;
     private Long maxParallelThreshold;
     private BuildEventPublisher buildEventPublisher;
+    private List<String> accountAllowList;
```
tylerwowen, line 105: This should either be `Collection`, so we accept a wide range of containers,
or `Set` so we can have some performance guarantee.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/TeletraanServiceConfiguration.java
tylerwowen, line 1: should not have this format only change

deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
ntascii: not sure if they are working with yaml configs. https://github.com/spring-projects/spring-boot/issues/1415 Based on the current usages, list is widely used so I'd keep it for now, we can consider move to others if needed.

tylerwowen: We can convert in the factory. Currently the performance doesn't matter. However, since we are doing this for every ping, and if the allow list grows, then the performance will matter at some point.

deploy-service/common/src/main/java/com/pinterest/deployservice/pingrequests/Ec2InstanceValidator.java
tylerwowen: My bad, should be `isBlank`.
Besides, you need `import org.apache.commons.lang3.StringUtils;`

tylerwowen (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/ServiceContext.java
```diff
@@ -469,4 +471,12 @@ public void setBuildEventPublisher(
         BuildEventPublisher buildEventPublisher) {
         this.buildEventPublisher = buildEventPublisher;
     }
+
+    public Set<String> getAccountAllowList() { 
+        return accountAllowList;
+    }
+
+    public void setAccountAllowList(List<String> accountAllowList) {
```
tylerwowen, line 479: non-blocking, it's better to have `Collection<String> accountAllowList`

tylerwowen (approved):

PR #1349, opened 2023-11-22
Rename stage type env var

osoriano:
The Teletraan stage type is now exposed as an environment variable to
the host.

Rename the env var to match the expected name

### DIFF
--- a/deploy-agent/deployd/common/config.py
+++ b/deploy-agent/deployd/common/config.py
@@ -105,7 +105,7 @@ class Config(object):
         if deploy_status.report.stageName:
             self._environ['STAGE_NAME'] = deploy_status.report.stageName
         if deploy_status.report.stageType:
-            self._environ['COMPUTE_STAGE_TYPE'] = deploy_status.report.stageType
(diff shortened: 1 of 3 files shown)


deploy-agent/deployd/common/config.py
```diff
@@ -105,7 +105,7 @@ def update_variables(self, deploy_status):
         if deploy_status.report.stageName:
             self._environ['STAGE_NAME'] = deploy_status.report.stageName
         if deploy_status.report.stageType:
-            self._environ['COMPUTE_STAGE_TYPE'] = deploy_status.report.stageType
+            self._environ['COMPUTE_ENV_TYPE'] = deploy_status.report.stageType
```
gzpcho, line 108: Should we also rename the `stageType` field in the report?

gzpcho (approved):

deploy-agent/deployd/common/config.py
gzpcho: What is the difference between env `{envName}-{stageType}` and this compute env?

ntascii (approved):

PR #1350, opened 2023-11-22
Allow deploy agent run serverless on existing build 

ntascii:
## Summary 
We can reproduce end to end deployment flow with an existing build on the host. 
Add a flag to specify the starting deploy stage so we can skip downloading. 
I will use it to POC running multiple deploy agents on a host. This can be used for testing telefig configs and deployment scripts as well. 
## Test plan 
Added UT
Ran on devapp and a sample host. 
```bazel-bin/deployd/deploy-agent -e prod -f /etc/deployagent.conf --use-facter --mode serverless --build "{\"build_branch\": \"master\",\"build_name\": \"beacon\",\"build_commit\": \"bfb061eb8f0031fb26d373b34787ff0b389a6315\"}" --env-name beacon --deploy-stage PRE_RESTART```

### DIFF
--- a/deploy-agent/deployd/__init__.py
+++ b/deploy-agent/deployd/__init__.py
@@ -27,4 +27,4 @@ REDEPLOY_MAX_RETRY = 3
 # 2: puppet applied successfully with changes
 PUPPET_SUCCESS_EXIT_CODES = [0, 2]
 	
-__version__ = '1.2.53'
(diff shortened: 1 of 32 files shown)


gzpcho (approved):

gzpcho:
> Ran on devapp and a sample host.

Can you share the command(s) you used?

PR #1353, opened 2023-11-28
Short in-memory cache for Knox

tylerwowen:
Add short 1-minute in-memory cache for Knox key. This will remove most disk access for key read yet to ensure key replacement doesn't require a restart of the application.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/common/KnoxKeyReader.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/common/KnoxKeyReader.java
@@ -31,6 +31,7 @@ public class KnoxKeyReader implements KeyReader {
 
     private String knoxKeyId;
     private KnoxKeyManager knoxManager;
+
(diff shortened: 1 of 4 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/common/KnoxKeyReader.java
```diff
@@ -15,36 +15,64 @@
...
+    private String knoxKeyId;
     private KnoxKeyManager knoxManager;
 
-    public void init(String key) {
+    private final LoadingCache<String, Optional<String>> knoxCache;
+
+    public KnoxKeyReader() {
+        knoxCache = CacheBuilder.newBuilder()
+                .maximumSize(1)
+                .expireAfterWrite(1, TimeUnit.MINUTES)
```
ntascii, line 40: should we make it configurable?

ntascii (approved):

PR #1355, opened 2023-11-30
update stage_type if it is DEFAULT and delivery_type is different with it 

liyaqin1:
We will keep one source of truth for stage_type on Teletraan side.
When the delivery_type is different with stage_type and stage_type is not DEFAULT, we reject the deployment API call;
When the delivery_type is different with stage_type and stage_type is DEFAULT, we update the stage_type;

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
@@ -425,7 +425,7 @@ public class DeployHandler implements DeployHandlerInterface{
         deployBean.setOperator(operator);
         
         //compare the current stage_type 
-        if(StringUtils.isEmpty(deliveryType) == false && envBean.getStage_type().toString().equals(deliveryType) == false) {
+        if(StringUtils.isNotBlank(deliveryType) && !envBean.getStage_type().toString().equals(deliveryType)) {
             if (envBean.getStage_type() != EnvType.DEFAULT) {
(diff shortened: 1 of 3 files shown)


liyaqin1:
Feel free to test here:
Screenshot 2023-12-05 at 5 39 40 PM

osoriano (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -421,6 +423,39 @@ public String deploy(EnvironBean envBean, String buildId, String desc, String op
         deployBean.setDescription(desc);
         deployBean.setDeploy_type(DeployType.REGULAR);
         deployBean.setOperator(operator);
+        
+        //compare the current stage_type 
+        if(StringUtils.isEmpty(deliveryType) == false && envBean.getStage_type().toString().equals(deliveryType) == false) {
```
ntascii, line 428: StringUtils.isNotBlank(deliveryType)

ntascii: nit:!envBean.getStage_type().toString().equals(deliveryType)

liyaqin1: updated

osoriano (approved):
LGTM thanks. From the discussion, this is a short term solution and we can plan a longer term solution separately

PR #1357, opened 2023-12-01
Use enum for bidirectional mappings

gzpcho:
Use python3 preferred way to define and access bidirectional mappings. 

Test Plan:

PR #1358, opened 2023-12-06
Increase unit test coverage

gzpcho:
Add unit test to increase coverage of changes made in #1357


deploy-agent/tests/unit/deploy/client/test_base_client.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import unittest
+from abc import ABCMeta, abstractmethod
+from future.utils import with_metaclass
```
ntascii, line 17: why we need this? the codebase is already py3.

ntascii (approved):

PR #1367, opened 2023-12-07
Show cluster config only if cluster exists

gzpcho:
[CDP-7278](https://jira.pinadmin.com/browse/CDP-7278)


rfleur01 (changes requested):
Some services only have a group and no corresponding cluster record, so you should also  conditionally the component based on the existence of the group as well

PR #1340, opened 2023-11-15
Send ec2 tags from deployd to TAS

liyaqin1:
For ec2 tags used by deployment constraint, no need to use Teletraan worker to query from CMDB. We can get them from deployd agent running on the ec2 host.

PR #1370, opened 2023-12-08
Add availability_zone in the ec2_tags

liyaqin1:
Related PR: https://github.com/pinterest/teletraan/pull/1340

### DIFF
--- a/deploy-agent/deployd/__init__.py
+++ b/deploy-agent/deployd/__init__.py
@@ -27,4 +27,4 @@ REDEPLOY_MAX_RETRY = 3
 # 2: puppet applied successfully with changes
 PUPPET_SUCCESS_EXIT_CODES = [0, 2]
 	
-__version__ = '1.2.54'
+__version__ = '1.2.55'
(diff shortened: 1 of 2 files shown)


ntascii:
can we update the info using availablityZone in PinRequest instead of using ec2 tags `PingRequest(hostId=i-08211786ab8852f3c, hostName=helloworlddummyservice-server-u18-0a03fd62, hostIp=10.3.253.98, agentVersion=1.2.46-pinterest-1.0.2, autoscalingGroup=helloworlddummyservice-server-u18, availabilityZone=us-east-1e, stageType=latest, groups=['CMP',`

liyaqin1:
> can we update the info using availablityZone in PinRequest instead of using ec2 tags `PingRequest(hostId=i-08211786ab8852f3c, hostName=helloworlddummyservice-server-u18-0a03fd62, hostIp=10.3.253.98, agentVersion=1.2.46-pinterest-1.0.2, autoscalingGroup=helloworlddummyservice-server-u18, availabilityZone=us-east-1e, stageType=latest, groups=['CMP',`

On TAS side, we only handle the tags like this https://sourcegraph.pinadmin.com/github.com/pinterest/teletraan/-/blob/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java?L647:9
So, it is better to keep all ec2 tags in one place. Also, the tag name is availability_zone instead of availablityZone as you can see here:
Screenshot 2023-12-18 at 6 06 58 PM

ntascii (dismissed):

ntascii (approved):

PR #1371, opened 2023-12-08
Add return type hints

gzpcho:


ntascii (commented):
how did you get the right types? any way we can automate?

gzpcho:
> how did you get the right types? any way we can automate?

A lot of manual inspection 😅

deploy-agent/deployd/agent.py
```diff
@@ -455,7 +458,7 @@ def _update_deploy_alias(self, deploy_goal):
                                                                   deploy_goal.envName))
 
     @staticmethod
-    def plan_changed(old_response, new_response):
+    def plan_changed(old_response, new_response) -> Union[PingResponse, bool, DeployGoal]:
```
ntascii, line 461: this might be a signal that we need to refactor the code to have a single return type. Btw, can we use `PingResponse | bool | DeployGoal`?

deploy-agent/deployd/common/helper.py
```diff
@@ -59,7 +60,7 @@ def get_build_id(filename, env_name):
         return False, None
 
     @staticmethod
-    def get_stale_builds(build_timestamps, num_builds_to_retain=2):
+    def get_stale_builds(build_timestamps, num_builds_to_retain=2) -> Generator:
```
ntascii, line 63: why the type is generator?

deploy-agent/deployd/agent.py
gzpcho: Agreed to your first point. That syntax is only supported in python 3.10 and above. https://peps.python.org/pep-0604/

deploy-agent/deployd/common/helper.py
gzpcho: This is a generator function since it calls the `yield` keyword. https://stackoverflow.com/questions/43658999/what-is-the-return-type-hint-of-a-generator-function

gzpcho:
Will split this PR into 3 where each change 1/ agent & clients 2/common dir 3/download dir for easier and safer rollout

gzpcho:
Split up into the following PRs:
#1403 
#1404 
#1405  
#1406

PR #1372, opened 2023-12-11
handle exception from GitHub get commit API

knguyen100000010:
### DO NOT MERGE

Teletraan get-commit API throws 500 errors to clients when a given commit hash it not found in GitHub (Teletraan is calling GitHub).

Reproduce the problem with:

```
v1/commits/github/[OWNER]/[REPO]/[INVALID_SHA]
```

## Fix

Catch `IOException` and re-throw an internal exception with helpful message.

## Test

Will test in dev

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
@@ -23,7 +23,7 @@ import com.pinterest.deployservice.exception.TeletaanInternalException;
 
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/scm/GithubManager.java
```diff
@@ -19,40 +19,40 @@
 import com.pinterest.deployservice.common.EncryptionUtils;
 import com.pinterest.deployservice.common.HTTPClient;
 import com.pinterest.deployservice.common.KnoxKeyReader;
+import com.pinterest.deployservice.exception.TeletaanInternalException;
+
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
-
+import org.apache.commons.lang.StringUtils;
```
tylerwowen, line 26: Let's use lang3

tylerwowen (approved):

PR #1379, opened 2023-12-13
do not allow deleting build associated with deployment

knguyen100000010:
## Why

A deployment is associated with a build. If there exists a deployment D for build B, this build **_should not_** be deleted. Otherwise:
1. Deployment D will then refer to a **_nonexistent_** build.
2. If D is the current deployment for an environment, or being actively deployed, this will cause problem since the build info is missing.


ntascii (approved):

liyaqin1 (approved):
@knguyen100000010 The code looks good to me. I am wondering how to delete a build. I could not find the button on Teletraan UI. Why we have such Teletraan API? Can we disable the API? Thanks.

PR #1381, opened 2023-12-14
extending the existing prompt to mention x86 architecture in basic setting page

liyaqin1:
Its confusing that instance type won't let you select graviton instances on that view but there is no mention of architecture (x86 v aarch64).


liyaqin1:
Current UI:
Screenshot 2023-12-13 at 4 54 12 PM

Improved UI:
Screenshot 2023-12-13 at 4 58 45 PM

robbintt (approved):

PR #1390, opened 2024-01-09
Correct variable name to containerHealthStatus

gzpcho:
serviceHealthStatus was a typo on the actual name returned by the Teletraan API. Also, improved the condition that renders the health icon. Here are some of the possible values it might see for containerHealthStatus:
- "service-a:healthy"
- "service-a:unhealthy"
- ""
- "service-a:healthy;service-b:healthy"
- "service-a:healthy;service-b:unhealthy"

Example:

Display Mode: Container Status
Screenshot 2024-01-10 at 3 45 15 PM


robbintt:
Could you add before/after screenshots and description of the manual testing?

gzpcho:
Previously only the "Not Found" icon was shown. Attached a screenshot of a healthy host example

PR #1391, opened 2024-01-09
fix typo

liyaqin1:
Remove duplicate property in Teletraan API


liyaqin1:
We have duplicate property in AgentBean like:
Screenshot 2024-01-05 at 3 24 45 PM
Using this PR to fix, we got:
Screenshot 2024-01-09 at 3 13 59 PM

ntascii (approved):

PR #1405, opened 2024-01-17
[deploy-agent] Return type hints in staging & types folders

gzpcho:
Improve readability for low risk:

> The Python runtime does not enforce function and variable type annotations.

### DIFF
--- a/deploy-agent/deployd/staging/transformer.py
+++ b/deploy-agent/deployd/staging/transformer.py
@@ -17,6 +17,7 @@ import traceback
 from string import Template
 import re
 import logging
+from typing import List
 from deployd import IS_PINTEREST
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


deploy-agent/deployd/staging/transformer.py
```diff
@@ -73,7 +73,7 @@ def _translate(self, from_path, to_path):
             log.error('Fail to translate script {}, stacktrace: {}'.format(from_path,
                                                                            traceback.format_exc()))
 
-    def transform_scripts(self, script_dir, template_dirname, script_dirname):
+    def transform_scripts(self, script_dir, template_dirname, script_dirname) -> list:
```
ntascii, line 76: probably we should be consistent to use either list or List from typing. wdyt?

deploy-agent/deployd/types/deploy_goal.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, jsonValue=None):
             self.firstDeploy = jsonValue.get('firstDeploy')
             self.isDocker = jsonValue.get('isDocker')
 
-    def __key(self):
+    def __key(self) -> tuple:
```
ntascii, line 54: same with tuple and Tuple from typing

deploy-agent/deployd/staging/transformer.py
gzpcho: good catch, thanks!

deploy-agent/deployd/types/deploy_goal.py
gzpcho: good catch, thanks!

ntascii (approved):

PR #1408, opened 2024-01-19
Added possibility to restart all failed hosts without checking any

vitalii-honchar:
# Context

Made button to restart failed deployment on all failed hosts available without checking any hosts.

# Proof of Work

## Failed hosts for current deployment
Screenshot 2024-01-19 at 10 45 26 AM

## Confirmation of retrying deployment

Screenshot 2024-01-19 at 10 56 07 AM

## Active button without selected hosts
Screenshot 2024-01-19 at 10 46 34 AM

## Button works without selected hosts

Screenshot 2024-01-18 at 7 04 46 PM

### DIFF
--- a/deploy-board/deploy_board/templates/hosts/hosts.tmpl
+++ b/deploy-board/deploy_board/templates/hosts/hosts.tmpl
@@ -6,27 +6,27 @@
       <strong>Deploy:</strong><a href="/deploy/{{ deployId }}"> {{ deployId }}</a>
        <div class="btn-toolbar pull-right">
           <button id="pauseButton"
-          class="btn btn-default btn-sm" data-toggle="tooltip"
+          class="btn btn-default btn-sm checked-hosts-btn" data-toggle="tooltip"
           title="Pause selected hosts">
           <span class="glyphicon glyphicon-pause"></span> Pause
           </button>
           <button id="resumeButton"
-          class="btn btn-default btn-sm" data-toggle="tooltip"
+          class="btn btn-default btn-sm checked-hosts-btn" data-toggle="tooltip"
           title="Resume selected hosts">
           <span class="glyphicon glyphicon-play"></span> Resume
           </button>
           <button id="resetButton"
-          class="btn btn-default btn-sm" data-toggle="tooltip"
+          class="btn btn-default btn-sm checked-hosts-btn" data-toggle="tooltip"
           title="reset selected hosts">
           <span class="glyphicon glyphicon-repeat"></span> Reset
           </button>
           <button id="terminateButton"
-          class="btn btn-default btn-sm" data-toggle="tooltip"
+          class="btn btn-default btn-sm checked-hosts-btn" data-toggle="tooltip"
           title="Soft terminate selected hosts">
           <span class="glyphicon glyphicon-off"></span> Terminate
           </button>
           <button id="forceTerminateButton"
-          class="btn btn-default btn-sm" data-toggle="tooltip"
+          class="btn btn-default btn-sm checked-hosts-btn" data-toggle="tooltip"
           title="Soft terminate selected hosts">
           <span class="glyphicon glyphicon-remove"></span> Force Terminate
           </button>
@@ -128,6 +128,12 @@
     });
 
     $('#hostTable input[type="checkbox"]').click(function(){
+        if ($("#hostTable input:checkbox:checked").length <= 0) {
+            $(".checked-hosts-btn").prop("disabled", true);
+        } else {
+            $(".checked-hosts-btn").prop("disabled", false);
+        }
+
         if ($(this).is(":checked")) {
           checkedHosts.push($(this).val());
         } else {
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/templates/hosts/hosts.tmpl
```diff
@@ -88,15 +88,15 @@
     <div class="pull-left">
         <button id="reset_{{ deployId }}"
                 class="deployToolTip btn btn-primary" data-toggle="tooltip"
-                title="Retry the current deploy step on this host.">
-            Retry deploy on all hosts
+                title="Retry deploys on all failed hosts.">
```
tylerwowen, line 91: Based on the query, it only retry current deploy
```java
    private static final String RESET_FAILED_AGENTS =
        "UPDATE agents SET state='RESET' " +
        "WHERE env_id=? AND deploy_id=? AND " +
        "status!='SUCCEEDED' AND status!='UNKNOWN' AND state!='RESET'";
```

vitalii-honchar: @tylerwowen yes, our intention is to restart deployment on all failed hosts not all deployments on the all failed hosts.

And this query will restart all failed hosts for current deployment. We don't want to restart all deployments on failed hosts, because there is a possibility that these failed hosts for current deployment will contains another not failed deployments, so it's not need to restart.

tylerwowen: My comment is regarding the change from "the current deploy step" to "deploys". It ls less clear after your change.

deploy-board/deploy_board/templates/hosts/hosts.tmpl
```diff
@@ -85,18 +85,18 @@
...
+            Retry deploys on all failed hosts
         </button>
     </div>
     {% endif %}
 
     {% if pinterest and host_ids %}
-    <div class="pull-left">
-        <button class="deployToolTip btn btn-default btn-block" data-target="#terminateHost"
+    <div class="pull-left ml-1">
+        <button class="deployToolTip btn btn-default btn-block checked-hosts-btn" data-target="#terminateHost"
```
tylerwowen, line 99: This button shouldn't be blocked either, right?

deploy-board/deploy_board/templates/hosts/hosts.tmpl
```diff
@@ -147,9 +152,9 @@
         });
 
         if ($("#hostTable input:checkbox:checked").length <= 0) {
-            $(".btn").prop("disabled", true);
+            $(".checked-hosts-btn").prop("disabled", true);
         } else {
-            $(".btn").prop("disabled", false);
+            $(".checked-hosts-btn").prop("disabled", false);
         }
```
tylerwowen, line 158: Can one of these be removed? 
```
if ($("#hostTable input:checkbox:checked").length <= 0) {
            $(".checked-hosts-btn").prop("disabled", true);
        } else {
            $(".checked-hosts-btn").prop("disabled", false);
        }
```

vitalii-honchar: As far as we don't have any buttons which can be clicked only by checked hosts, I deleted this JQuery code to enable/disable buttons.

tylerwowen: The Pause, Resume etc. buttons on the upper right corner still need to disabled when there is no host checked, right?

vitalii-honchar: Oh, yes. I fixed this.  Also we can't delete code from your comment https://github.com/pinterest/teletraan/pull/1408#discussion_r1459469501 because this code is responsible for enabling and disabling buttons.

Screenshot 2024-01-19 at 4 51 55 PM

tylerwowen: LG. My original question is regarding whether one of the repeated listener can be removed.

vitalii-honchar: No, them are responsible for different things:
this listener https://github.com/pinterest/teletraan/pull/1408/files/0cc10746727c3be9870601e0d788bedec37c14d0#diff-9311ec701ea3da55954c03dcb3ad085f0319ea31328bc44bab0319cd33f26915R130-R135 
is responsible for enable and disable button after user click.

this listener
https://github.com/pinterest/teletraan/pull/1408/files/0cc10746727c3be9870601e0d788bedec37c14d0#diff-9311ec701ea3da55954c03dcb3ad085f0319ea31328bc44bab0319cd33f26915L149-L153

is responsible for initially render button on the first page load. So we can't remove any of them.

vitalii-honchar: This is not repeated listeners, them are different. Of course we can refactor code and create a separated function for it, but again as for me this is out of scope of current ticket.

tylerwowen: What about
```html
<button id="resumeButton"
          class="btn btn-default btn-sm checked-hosts-btn" data-toggle="tooltip"
          title="Resume selected hosts" disabled>
```
mark as resolved

tylerwowen (approved):
LGTM

PR #1412, opened 2024-01-23
build: Pre-commit hooks

gzpcho:
Pre-commit hooks are non-blocking and only run on the files with changes being commit. You can skip a pre-commit hook by specifying `SKIP=<hook_id>` before `git commit`. The reason for these changes is for having a single consistent style and avoid small mistakes through the use linters. 

Pre-commit hooks included
- Recommended general purpose hooks
- black for formatting .py files
- flake8 for linting .py files 
- djlint for formatting and linting django template files

Setup
```
pip install pre-commit tox
pre-commit install
```

Test Execution
```
pre-commit run --all-files
tox -e black
tox -e flake8
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,25 @@
+repos:
+  - repo: 'https://github.com/pre-commit/pre-commit-hooks'
+    rev: v2.3.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: 'https://github.com/psf/black'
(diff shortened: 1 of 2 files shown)


gzpcho:
> If this is something we care, it's better to also have these in CI job.

One caveat of that approach, whether it is Jenkins or GHA, is that if we ran it on branch it would create a new commit every time and you would need to pull those changes every time. We also don't want these to block the PR as they are mainly attempts to improve new and existing code rather than get in the way at the moment. I think that if people use this often, we can evaluate making a blocking step of our CI.

PR #1425, opened 2024-01-23
Add LongTaskTimer to track launching hosts

tylerwowen:
Same as #1394 

Main logic of `DefaultHostClassifier` and `MetricsEmitter`

```mermaid
flowchart TD
    S([Run at an interval]) --> A0[(Fetch new data from DB)]
    A0 --> A1{Process differences}
    A1 --> R1[New hosts]
    A1 ---> |Launched or exceeded max threshold| R2[Removed hosts]
    R1 -.-> R3[Timeout hosts]
    R4 -.-> R3[Timeout hosts]
    A1 --> R4[Carryover hosts]
    R2 --> A3{Exceeded SLO?}
    A3 --> |No| T0((sucess++))
    A3 --> |Yes| T1((failed++))
    R4 --> Init[Initializing hosts]
    R3 -.-> Init
    R1 --> Init
    Init -.-> |Current data| A1
    Init --> A2[Report histogram]
```

PR #1426, opened 2024-01-23
Add metrics for first deploy success, provision latency

tylerwowen:
Add metrics for first deploy success, provision latency and deploy latency

**Stack**:
- #1425
- #1424

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
@@ -51,7 +51,7 @@ public class DBHostDAOImpl implements HostDAO {
     private static final String GET_GROUP_SIZE = "SELECT COUNT(host_id) FROM hosts WHERE group_name=?";
     private static final String GET_ALL_HOSTS_BY_GROUP = "SELECT * FROM hosts WHERE group_name=? AND state!='TERMINATING'";
     private static final String GET_HOST_BY_NAME = "SELECT * FROM hosts WHERE host_name=?";
-    private static final String GET_HOST_BY_HOSTID = "SELECT * FROM hosts WHERE host_id=?";
+    private static final String GET_HOST_BY_HOSTID = "SELECT * FROM hosts WHERE host_id=? ORDER BY create_date";
     private static final String GET_HOSTS_BY_STATES = "SELECT * FROM hosts WHERE state in (?, ?, ?) GROUP BY host_id ORDER BY last_update";
     private static final String GET_GROUP_NAMES_BY_HOST = "SELECT group_name FROM hosts WHERE host_name=?";
     private static final String GET_STALE_AGENTLESS_HOST_IDS = "SELECT DISTINCT hosts.host_id FROM hosts LEFT JOIN hosts_and_agents ON hosts.host_id = hosts_and_agents.host_id WHERE hosts.last_update < ? AND hosts_and_agents.host_id IS NULL ORDER BY hosts.last_update DESC LIMIT ?";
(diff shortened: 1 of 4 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java
```diff
@@ -88,7 +94,7 @@ public class PingHandler {
     private static final PingResponseBean NOOP;
     private static final Set<String> EMPTY_GROUPS;
     private static final String PINTEREST_MAIN_AWS_ACCOUNT = "998131032990";
-    //private static final long AGENT_COUNT_CACHE_TTL = 5 * 1000;
+    private static final String INFRA_LATENCY_TIMER_NAME = CUSTOM_NAME_PREFIX + "teletraan.%s.infra_latency";
```
ntascii, line 97: how about provisioning latency?

ntascii (approved):

ntascii (dismissed):

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
```diff
@@ -51,7 +51,7 @@ public class DBHostDAOImpl implements HostDAO {
     private static final String GET_GROUP_SIZE = "SELECT COUNT(host_id) FROM hosts WHERE group_name=?";
     private static final String GET_ALL_HOSTS_BY_GROUP = "SELECT * FROM hosts WHERE group_name=? AND state!='TERMINATING'";
     private static final String GET_HOST_BY_NAME = "SELECT * FROM hosts WHERE host_name=?";
-    private static final String GET_HOST_BY_HOSTID = "SELECT * FROM hosts WHERE host_id=?";
+    private static final String GET_HOST_BY_HOSTID = "SELECT * FROM hosts WHERE host_id=? ORDER BY create_date";
```
osoriano, line 54: Any reason to update this?

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -368,22 +377,38 @@ AgentBean genUpdateBeanByReport(PingReportBean report, AgentBean agent) {
...
 
+    private void emitMetrics(AgentBean updateBean) {
+        try {
+            EnvironBean env = envs.get(updateBean.getEnv_id());
+            Metrics.timer(String.format(DEPLOY_LATENCY_TIMER_NAME, env.getEnv_name(), env.getStage_name()))
+                    .record(Duration.ofMillis(updateBean.getFirst_deploy_time() - updateBean.getStart_date()));
+            Metrics.counter(FIRST_DEPLOY_COUNTER_NAME, "success",
+                    String.valueOf(updateBean.getStatus().equals(AgentStatus.SUCCEEDED)),
+                    "env", env.getEnv_name(),
+                    "stage", env.getStage_name())
```
osoriano, line 405: Any reason to set the env/stage in the metric name instead of the tags? Looks like both approaches are used here

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
tylerwowen: Yes. The current order is indeterministic. The new usage [bellow](https://github.com/pinterest/teletraan/pull/1426/files#diff-067abe2f4de3c1b2b383cf5d982f41aa5d8c8232b22d68b5b617549816b70698R232) will benefit from the ordering. I also just code references and no existing code depend on the order of the query.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
tylerwowen: That's a good question. The reason is that the metrics is moved from Rodimus, and it has env and stage as tags there. Since the counter name will be changed anyways, I think it's better to have the env and stage in the name.

osoriano (dismissed):

osoriano (approved):

PR #1430, opened 2024-01-24
Added validation for a form to create base images in UI

vitalii-honchar:
# Context

To always create valid base images validation for a form which creates new base images were added:
- `Cell` field should starts from `aws-` prefix if provider is AWS.
- `Cloud specific image id ` should starts from `ami-` prefix if provider is AWS.

Added `basic` field to the request for create base_images which became required after recently added validation for endpoints which creates base_images.

# Proof Of Work

## Form with 2 invalid fields

Screenshot 2024-01-24 at 5 58 41 PM

## Form with 1 invalid field
 
Screenshot 2024-01-24 at 5 58 57 PM


deploy-board/deploy_board/templates/clusters/base_images.html
```diff
@@ -168,4 +177,60 @@ <h4 class="modal-title">Create Base Image Confirm</h4>
...
+        var errorElement = $("#create-base-image-form-errors");  
+        var errors = "";
+        var validationErrors = [validateProviderName(provider), validateCellName(provider)];
+        validationErrors
+            .filter(err => err != null)
+            .forEach((error, i) => {
+                errors += createValidationErrorHtml(i === 0, "Validation error: " + error);
+            });
+
+        if (errors.length === 0) {
```
ntascii, line 193: falsify errror?

ntascii (approved):

deploy-board/deploy_board/templates/clusters/base_images.html
vitalii-honchar: @ntascii I didn't get, can you please explain your comment?

vitalii-honchar: @ntascii `errors.length === 0` uses 3 `=`, because this is JS strict check for equation - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

yongwen:
while you are here, could you also change the order of the fields, putting the "provider" as the first one?

vitalii-honchar:
I changed an order of fields

Screenshot 2024-01-25 at 4 31 22 PM

yongwen (approved):

PR #1434, opened 2024-01-29
fix issue with findall

knguyen100000010:
Fix this problem:

```
"exc_info": "(<class 'TypeError'>, TypeError('cannot use a string pattern on a bytes-like object')
```


gzpcho (approved):
thanks you! can you attach how you validated the changes?

PR #1445, opened 2024-02-03
Fix keyError on Teletaan UI

liyaqin1:
Usually, we don't have this error:
Screenshot 2024-02-02 at 5 20 08 PM

But if someone call the API to add an AMI and the cell is not in the list, we could encounter this error.
Screenshot 2024-02-02 at 5 21 26 PM


liyaqin1:
After the fix, the error is gone.
Screenshot 2024-02-02 at 5 26 12 PM

ntascii (approved):

PR #1447, opened 2024-02-06
fetch ec2 tags from aws instead of from facter command

liyaqin1:
If a user adds tags through the AWS console, the previous method wouldn't be able to fetch these tags using the 'facter' command.

### DIFF
--- a/deploy-agent/deployd/client/client.py
+++ b/deploy-agent/deployd/client/client.py
@@ -18,6 +18,7 @@ import os
 import socket
 import traceback
 import json
+import boto3
 
 from deployd.client.base_client import BaseClient
 from deployd.client.restfulclient import RestfulClient
@@ -171,10 +172,21 @@ class Client(BaseClient):
             # so need to read ec2_tags from facter and parse Autoscaling tag to cover this case
             if not self._autoscaling_group:
                 ec2_tags = facter_data.get(ec2_tags_key)
-                if ec2_tags:
-                    ec2_tags['availability_zone'] = self._availability_zone
-                self._ec2_tags = json.dumps(ec2_tags) if ec2_tags else None
                 self._autoscaling_group = ec2_tags.get(asg_tag_key) if ec2_tags else None
+
+            # Obtain all EC2 tags, inclusive of those that are outside the scope of the 'facter' command.
+            self._ec2_tags = None
+            try:
+                ec2 = boto3.resource('ec2', region_name='us-east-1')
+                ec2instance = ec2.Instance(self._id)
+                all_tags = {}
+                for tag in ec2instance.tags:
+                    all_tags[tag["Key"]] = tag["Value"]
+                all_tags["availability_zone"] = self._availability_zone 
+                self._ec2_tags = json.dumps(all_tags) 
+            except Exception:
+                log.warn("Failed to get host {} tags.".format(self._id))
+                pass
 
             if not self._stage_type and not self._stage_type_fetched:
                 self._stage_type = facter_data.get(stage_type_key, None)
--- a/deploy-agent/setup.py
+++ b/deploy-agent/setup.py
@@ -24,7 +24,7 @@ console_scripts = ['deploy-agent = deployd.agent:main',
                    'deploy-stager = deployd.staging.stager:main']
 
 install_requires = [
-    "requests==2.20.0",
+    "requests>=2.20.0",
     "gevent>=1.0.2,<=1.2.2; python_version < '3'",
     "gevent>=1.0.2,<=1.5.0; python_version < '3.8'",
     "gevent==20.12.0; python_version >= '3.8'",
--- a/deploy-agent/setup.py
+++ b/deploy-agent/setup.py
@@ -24,13 +24,14 @@ console_scripts = ['deploy-agent = deployd.agent:main',
                    'deploy-stager = deployd.staging.stager:main']
 
 install_requires = [
-    "requests>=2.20.0",
+    "requests==2.20.0",
     "gevent>=1.0.2,<=1.2.2; python_version < '3'",
     "gevent>=1.0.2,<=1.5.0; python_version < '3.8'",
     "gevent==20.12.0; python_version >= '3.8'",
     "lockfile==0.10.2",
     "boto>=2.39.0",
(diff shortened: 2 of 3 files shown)


deploy-agent/setup.py
```diff
@@ -30,6 +30,8 @@
     "gevent==20.12.0; python_version >= '3.8'",
     "lockfile==0.10.2",
     "boto>=2.39.0",
+    "boto3>=1.12.6",
```
gzpcho, line 33: Why not the latest version? 
1.34.36

gzpcho: https://deps.dev/pypi/boto3/1.34.36/versions

deploy-agent/setup.py
```diff
@@ -30,6 +30,8 @@
...
+    "urllib3<1.27,>=1.25.4",
```
gzpcho: Same as above, is there a reason for this specific version range?
https://deps.dev/pypi/urllib3/2.2.0/versions

deploy-agent/setup.py
liyaqin1: Given that we currently use requests==2.20.0, we need urllib3<1.25, >=1.21.1. The challenge here is that we must identify the precise boto3 and botocore versions which can adequately support urllib3 within the range of <1.25 and >=1.21.1.

deploy-agent/setup.py
liyaqin1: Given that we currently use requests==2.20.0, we need urllib3<1.25, >=1.21.1. The challenge here is that we must identify the precise boto3 and botocore versions which can adequately support urllib3 within the range of <1.25 and >=1.21.1.

deploy-agent/deployd/client/client.py
```diff
@@ -171,11 +172,22 @@ def _read_host_info(self):
...
                 ec2_tags = facter_data.get(ec2_tags_key)
-                if ec2_tags:
-                    ec2_tags['availability_zone'] = self._availability_zone
-                self._ec2_tags = json.dumps(ec2_tags) if ec2_tags else None
                 self._autoscaling_group = ec2_tags.get(asg_tag_key) if ec2_tags else None
 
+            # Obtain all EC2 tags, inclusive of those that are outside the scope of the 'facter' command.
+            self._ec2_tags = None
+            try:
+                ec2 = boto3.resource('ec2', region_name='us-east-1')
```
ntascii, line 180: I will cause additional aws requests right? do we currently fetch these tags from cmdb?

liyaqin1:
Tested on a test host with this latest 1.2.58 version. Got the correct tags on AWS console side:
`I0207 00:26:32.410.477876663208 MainThread /root/.pex/installed_wheels/502183de0cb11c911983d26ffd0f8af0f0f55ad06ae3d2f4e9c2ce7f94dfa18a/deploy_agent-1.2.58-py3-none-any.whl/deployd/client/client.py:204] Host information is loaded. Host name: helloworlddummyservice-server-test-osoriano-integ-0a01d5a4, IP: 10.1.213.164, host id: i-07a2bddf23a2706af, agent_version=1.2.46-pinterest-1.0.2, autoscaling_group: helloworlddummyservice-server-test-osoriano-integ, availability_zone: us-east-1e, ec2_tags: {"aws:ec2launchtemplate:id": "lt-0e2fc39c66fc4c8a9", "Name": "helloworlddummyservice-server-test-osoriano-integ-0a01d5a4", "aws:autoscaling:groupName": "helloworlddummyservice-server-test-osoriano-integ", "usage_tag": "SRE", "aws:ec2launchtemplate:version": "2", "Autoscaling": "helloworlddummyservice-server-test-osoriano-integ", "service_mapping": "helloworlddummyservice-server-test-osoriano", "**testTags": "yaqinli",** "availability_zone": "us-east-1e"}, stage_type: None, group: ['CMP', 'helloworlddummyservice-server-test-osoriano-integ', 'envoy-sidecar', 'singer-prod-sidecar', 'tcollector-prod-sidecar', 'mcrouter-prod-sidecar', 'metrics-agent-prod-sidecar', 'zk_update_monitor-prod-sidecar'], account id: 998131032990`

Screenshot 2024-02-06 at 4 30 30 PM

We can see the constraint is setup correctly here:
Screenshot 2024-02-06 at 4 31 36 PM

We can confirm that the tag originated from the deployd on the host and not from a CMDB query. This is because the CMDB has not been updated with the new tag yet, given that I had only added it a few minutes ago:
Screenshot 2024-02-06 at 4 27 30 PM

Further evidence for this can be found in the following log record:
Screenshot 2024-02-06 at 4 34 57 PM

deploy-agent/setup.py
gzpcho: Why not upgrade requests if that's the case?

deploy-agent/setup.py
gzpcho: Why not upgrade requests if that's the case?

deploy-agent/deployd/client/client.py
liyaqin1: Yes, it does invoke AWS. We need to retrieve all tags from AWS and then save them in our database. Without this step, a tag added by a user cannot be fetched through the 'facter' command. As of now, all tags are stored in our database by querying them from the CMDB.

deploy-agent/setup.py
gzpcho: The latest requests version is 2.31.0. And because it's a minor version change it shouldn't have breaking changes. Can we use that one?

deploy-agent/setup.py
gzpcho: The latest requests version is 2.31.0. And because it's a minor version change it shouldn't have breaking changes. Can we use that one?

liyaqin1: The principal aim of this pull request is to incorporate and utilize boto3. I'd prefer not to meddle with the current version of request. If you're keen to streamline the versioning aspect, you're more than welcome to invest some time enhancing and upgrading the versions. Thank you.

liyaqin1:
I have decided to discard this Pull Request as the 'facter' command does actually retrieve all tags. It's just that the delay seems a bit lengthy because it attempts to gather information every hour.

PR #1448, opened 2024-02-12
Prevent alarm created without scaling policy

knguyen100000010:
Prevent alarm creation if selected scaling policy is not existent.

image

### DIFF
--- a/deploy-board/deploy_board/templates/groups/asg_config.html
+++ b/deploy-board/deploy_board/templates/groups/asg_config.html
@@ -16,7 +16,6 @@
 {% include "groups/group_side_panel.html" %}
 </div>
 
-
 {% endblock %}
 
 {% block new-builds-panel %}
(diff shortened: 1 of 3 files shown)


deploy-board/deploy_board/templates/groups/asg_config.html
```diff
@@ -26,7 +25,14 @@
...
 
-
+{% if storage %}
+    {% for message in storage %}
+        <div class="alert alert-warning" role="alert">
+            <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+            <strong>Warning!</strong> {{ message | safe}}
+        </div>
+    {% endfor %}
+{% endif %}
```
rfleur01, line 35: Isn't this block similar to the `message_banner.tmpl`? Would it still work you include it in this file and rename "storage" to "messages" on line 1132 in group_view.py?

rfleur01 (approved):

PR #1459, opened 2024-02-21
Step Scaling Validation

knguyen100000010:
- Added validation for step scaling configurations. Tested.
- Fix a small issue with button hiding (add button should only be displayed when target tracking scaling tab is selected).

### DIFF
--- a/deploy-board/deploy_board/templates/groups/asg_policy.tmpl
+++ b/deploy-board/deploy_board/templates/groups/asg_policy.tmpl
@@ -153,7 +153,7 @@
 
         <div id="step-scaling" class="tabcontent">
             <h4>Step Scaling <sup><a href="https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-scaling-simple-step.html" target="_blank">What is step scaling?</a></sup></h4>
-                        <form id="stepAutoscalingPolicyConfigFormId" class="form-horizontal" role="form" method="post" action="/groups/{{ group_name }}/autoscaling/update_policy/">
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/templates/groups/asg_policy.tmpl
```diff
@@ -455,6 +453,159 @@ aria-hidden="true">
...
+
+            function validateAndParseSteps(stepString, isUp, isStep) {
+                if (stepString.length == 0) {
+                    alert("Steps or adjustments cannot be empty");
+                    return null;
+                }
+
+                var stepArray = stepString.split(',').map(Number);
+                for (i = 0; i < stepArray.length; i++) {
+                    if (isNaN(stepArray[i])) {
```
rwxzhu, line 549: nit: can this be Number.isNaN instead? I think the latter is preferred for some type conversion pitfalls w/ the global method, [for ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description).

rwxzhu (approved):
LGTM, just a couple of nits.

rwxzhu (approved):

ntascii:
> Added validation for step scaling configurations. Tested.
Can you add a test case and steps you've taken to check your work?

deploy-board/deploy_board/templates/groups/asg_policy.tmpl
```diff
@@ -455,6 +453,159 @@ aria-hidden="true">
         </div>
 
         <script>
+            function validateStepScalingConfig(stepScalingConfig) {
```
ntascii, line 456: should these js code be in a separate .js file? the template file is getting lengthy

PR #1466, opened 2024-02-23
Move private build on dev logic outside of bean

rwxzhu:
Follow up to #1453.

After merging, it was noticed that the setter was also used by the Apache DbUtils BeanProcessor when parsing database ResultSet -> EnvironBean, which caused cases where old `DEV` stages without allow_private_build enabled would still have private builds enabled. This change removes that scenario, so that the allow_private_build flag is always the source of truth.

### Before
Screenshot 2024-02-23 at 10 38 41 AM
Screenshot 2024-02-23 at 10 39 23 AM

### After
Screenshot 2024-02-23 at 10 46 55 AM


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -291,7 +291,7 @@ String internalDeploy(EnvironBean envBean, DeployBean deployBean) throws Excepti
         EnvironBean updateEnvBean = new EnvironBean();
         updateEnvBean.setDeploy_id(deployId);
         updateEnvBean.setDeploy_type(deployBean.getDeploy_type());
-        updateEnvBean.setStage_type(envBean.getStage_type());
+        updateEnvBean.updateStage_type(envBean.getStage_type());
```
vitalii-honchar, line 294: Looks good, but this changes fixes already present bug in the project. Please add unit tests that proofs that fix really works and we will have automated testing which will check that we will not have the same problem in the future.

vitalii-honchar (changes requested):

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
rwxzhu: Thanks, just added some. I ended up moving the business logic outside of the bean. I did add some tests from the API side, but didn't do the ones from db -> object as they wouldn't be relevant anymore.

liyaqin1:
The code appears to be well-written and efficient. However, consider revising the title of the PR to more accurately reflect its purpose or content.

PR #1483, opened 2024-03-01
highlight sidecars with matching host environment

rwxzhu:
If a sidecar agent has the same environment as the host environment, then the sidecar should be highlighted for ease of discovery, and the sidecars tab should not be folded.

#### Example of host with highlighted sidecars
Screenshot 2024-03-04 at 9 37 32 PM
Screenshot 2024-03-04 at 9 32 49 PM

### DIFF
--- a/deploy-board/deploy_board/templates/hosts/host_details.html
+++ b/deploy-board/deploy_board/templates/hosts/host_details.html
@@ -45,7 +45,7 @@
     {% if has_host_env_sidecar_agents %}
     {% include "panel_heading.tmpl" with panel_title="Sidecar Agent Details" panel_body_id="sidecarAgentDetailsId" direction="down" copy_host_name_button=True %}
     <div id="sidecarAgentDetailsId" class="collapse in panel-body">
-        <span>This host has relevant sidecars in it's environment. These have been marked with <span class="bg-success">green backgrounds</span> for visibility.</span>
+        <span>This is a sidecar deployment on the host. The relevant sidecar has been marked with a <span class="bg-info">blue background</span> for visibility.</span>
     {% else %}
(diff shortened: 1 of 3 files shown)


deploy-board/deploy_board/webapp/host_views.py
```diff
@@ -188,6 +188,14 @@ def get(self, request, name, stage, hostname):
             is_protected = autoscaling_groups_helper.is_hosts_protected(request, asg, [host_id])
 
         agent_wrappers, is_unreachable = get_agent_wrapper(request, hostname)
+        has_host_env_sidecar_agents = any(
+            x for x in agent_wrappers['sidecars']
```
ntascii, line 192: nit:x > agent

deploy-board/deploy_board/templates/hosts/host_details.html
```diff
@@ -42,11 +42,21 @@
 
 {% if agent_wrappers %}
 <div class="panel panel-default">
+    {% if has_host_env_sidecar_agents %}
+    {% include "panel_heading.tmpl" with panel_title="Sidecar Agent Details" panel_body_id="sidecarAgentDetailsId" direction="down" copy_host_name_button=True %}
+    <div id="sidecarAgentDetailsId" class="collapse in panel-body">
+        <span>This host has relevant sidecars in it's environment. These have been marked with <span class="bg-success">green backgrounds</span> for visibility.</span>
```
ntascii, line 48: Nit: This is a sidecar deployment on the host. 
I think instead of highlight green (which can infer that the deployment was succeeded), we can do in blue and make the text, eg.`Environment: beacon/prod-aws-us-east-1a Deploy: 86fEpuN8S3yMY9kri9eStw` bold.

rwxzhu: Thanks, makes sense. Updated the text and the styling and attached the new screenshots above.

ntascii (approved):

PR #1497, opened 2024-03-06
remove UNKNOWN tag from ami dropdown

rwxzhu:
Remove UNKNOWN status in AMI dropdown

Screenshot 2024-03-06 at 5 34 23 PM
Screenshot 2024-03-06 at 11 15 18 AM

### DIFF
--- a/deploy-board/deploy_board/static/js/base-image-utils.js
+++ b/deploy-board/deploy_board/static/js/base-image-utils.js
@@ -15,12 +15,12 @@ function getDefaultBaseImageId(baseImagesSorted) {
 
 function mapBaseImagesToOptions(baseImagesSorted) {
     return baseImagesSorted.map(o => {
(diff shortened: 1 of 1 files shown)


vitalii-honchar (approved):

ntascii (approved):
tks

ntascii:
as you are here, can you change [GOLDEN] -> [CURRENT_GOLDEN], also check if we can have other old golden amis have tag [GOLDEN]

rwxzhu:
> can you change [GOLDEN] -> [CURRENT_GOLDEN]  

done 👍 
> check if we can have other old golden amis have tag [GOLDEN]

@ntascii this should be possible though we might want to make a new endpoint in rodimus for getting all golden ami id's given a cell + name to make it more performant. It looks like with the currently existing endpoints we would have to get the tags for each base image in the dropdown, which would be ~100 api calls on page load. I'm thinking of leaving this for another PR.

ntascii (approved):

PR #1500, opened 2024-03-07
Added deduplication and sorting for displayed accounts in the dropdown menu

vitalii-honchar:
## Context

Added deduplication and sorting for displayed accounts in the dropdown menu. Fixes an issue when Teletraan receives duplicates of accounts with the same id for different cells. (But that accounts contains different configs as well, that's why we need them to store in Rodimus)

### Before these changes

Screenshot 2024-03-07 at 5 22 24 PM

### After these changes

Screenshot 2024-03-07 at 5 37 48 PM


ntascii:
it should show the account only without binding with the region, we have a separate dropdown for region.

vitalii-honchar:
@ntascii it will not show region as we discussed yesterday on the call.

ntascii (approved):

PR #1512, opened 2024-03-12
add accounts page

rwxzhu:
Create an accounts page that lists all accounts. Add a link to it on every page's navbar.

#### Manual Testing
tested on integ rodimus
Screenshot 2024-03-13 at 9 42 37 AM


deploy-board/deploy_board/templates/accounts/accounts.html
```diff
@@ -0,0 +1,40 @@
...
+                    <th class="col-lg-2">Name</th>
+                    <th class="col-lg-2">Cell</th>
+                    <th class="col-lg-2">Provider</th>
+                    <th class="col-lg-2"></th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for account in accounts %}
+                <tr>
+                    <td>{{ account.id }}</td>
```
vitalii-honchar, line 29: Display `ownerId` from the field `data.ownerId` on the top level for an account record with the `id` field.
`ownerId` contains AWS Account number and it's a primary interest for our users.

`id` is an unique identifier inside our system, but our users have minor interest for this field.

deploy-board/deploy_board/templates/accounts/accounts.html
```diff
@@ -0,0 +1,40 @@
...
+                    <td>{{ account.cell }}</td>
+                    <td>{{ account.cloudProvider }}</td>
+                    <td><a href="/accounts/{{ account.cloudProvider }}/{{ account.cell }}/{{ account.id }}">Details</a></td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+{% endblock %}
```
vitalii-honchar, line 40: Please add empty line at the end of every new file. It can be achieved throughout IDE config, for example Intelij IDEA:
Screenshot 2024-03-12 at 5 27 44 PM

deploy-board/deploy_board/templates/accounts/accounts.html
```diff
@@ -0,0 +1,40 @@
...
+{% block main %}
+<div class="panel panel-default">
+    <div class="panel-heading clearfix">
+        <h4 class="panel-title pull-left">Account Details</h4>
+    </div>
+    <div class="panel-body table-responsive">
+        <table class="table table-striped table-bordered table-condensed table-hover">
+            <thead>
+                <tr>
+                    <th>ID</th>
```
vitalii-honchar, line 19: Add a column `Owner ID` after the column `ID`

deploy-board/deploy_board/templates/accounts/accounts.html
rwxzhu: do we want to showthe account.id at all, or should we hide it and have owner id be the id. I'm not sure if it would be confusing to a user unfamiliar with how aws account ids look like.

rwxzhu: added a tooltip so that this shouldn't be an issue.
Screenshot 2024-03-13 at 9 43 16 AM

vitalii-honchar: Let's show account.id as well for internal communication

vitalii-honchar (approved):

PR #1519, opened 2024-03-13
Update utils.py to disable  IMDSv1 request for deploy-agent facter call

liyaqin1:
By default, AWS SDK will retry IMDSv1 request if v2 request is not successful. This change is to disable the behavior and prevent the IMDSv1 request.
Why is IMDSv1 bad?
IMDSv1 is dangerous to use with AWS EC2 because it lacks any built-in security features. The metadata endpoint is publicly accessible, which means that anyone who can reach the EC2 instance can potentially access the metadata. This can be a significant risk if an attacker gains access to the EC2 instance.

### DIFF
--- a/deploy-agent/deployd/common/utils.py
+++ b/deploy-agent/deployd/common/utils.py
@@ -207,7 +207,7 @@ def get_info_from_facter(keys) -> Optional[dict]:
         # increment stats - facter calls
         create_sc_increment('deployd.stats.internal.facter_calls_sum', 1)
         log.info(f"Fetching {keys} keys from facter")
-        cmd = ['facter', '-jp']
+        cmd = ['AWS_EC2_METADATA_V1_DISABLED=true', 'facter', '-jp']
         cmd.extend(keys)
         output = subprocess.run(cmd, check=True, stdout=subprocess.PIPE).stdout
         # timing stats - facter run time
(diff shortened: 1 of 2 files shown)


liyaqin1:
related jira ticket https://jira.pinadmin.com/browse/CDP-7759

deploy-agent/deployd/common/utils.py
```diff
@@ -207,7 +207,7 @@ def get_info_from_facter(keys) -> Optional[dict]:
         # increment stats - facter calls
         create_sc_increment('deployd.stats.internal.facter_calls_sum', 1)
         log.info(f"Fetching {keys} keys from facter")
-        cmd = ['facter', '-jp']
+        cmd = ['AWS_EC2_METADATA_V1_DISABLED=true', 'facter', '-jp']
```
ntascii, line 210: should it be set in cronjob in puppet? https://sourcegraph.pinadmin.com/github.com/pinternal/puppet-control-repo/-/blob/modules/teletraan/manifests/cron.pp?L10:125

liyaqin1: If we configure this within a Puppet cron job, we'll need to wait for the Puppet deployment to complete; similarly, if we set the environment variable in this Pull Request as proposed, we'll have to wait for the new deployd version to be fully deployed. Therefore, it is analogous in both scenarios.

liyaqin1: discussed off line

liyaqin1:
Current all env vars for deploy agent are set in puppet so having a single place to see all env vars is better than spreading the env vars configs in different places, which makes maintaining and debugging more difficult in the future.

PR #1522, opened 2024-03-14
Made dynamically displaying accounts depending on available hosts reports

vitalii-honchar:
# Context
Made dynamically displaying accounts depending on available hosts reports.

- Hide Accounts panel on an environment landing page if no hosts were provisioned from an account.

# Tests

## Display account from which cluster was provisioned

Screenshot 2024-03-14 at 12 29 34 PM

## Display hard coded primary accounts if hosts were provisioned from that cluster
Screenshot 2024-03-14 at 12 58 00 PM

## Don't display accounts panel if no information about account is available

Screenshot 2024-03-14 at 12 32 13 PM

### DIFF
--- a/deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
+++ b/deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
@@ -279,22 +279,6 @@
 </div>
 {% endwith %}
 
-<div class="panel panel-default">
-    <div class="panel-heading clearfix">
-        <h4 class="panel-title pull-left"><i class="fa fa-fw fa-user"></i> Accounts</h4>
-    </div>
-    <div class="panel-body">
-        <div class="row">
-            {% for account in accounts %}
-                <a href="#" class="deployToolTip btn btn-xs btn-default host-btn">
-                    <small>{{ account.name }}</small>
-                    <i class="fa fa-fw fa-user"></i>
-                </a>
-            {% endfor %}
-        </div>
-    </div>
-</div>
-
 <script>
     $(function () {
         $('.deployToolTip').tooltip({container: "#toolTipContent", delay: { show: 400, hide: 10 }});
--- a/deploy-board/deploy_board/webapp/env_views.py
+++ b/deploy-board/deploy_board/webapp/env_views.py
@@ -238,7 +238,7 @@ def update_deploy_progress(request, name, stage):
     cluster = clusters_helper.get_cluster(request, env.get('clusterName'))
     if cluster is not None:
         add_account_from_cluster(request, cluster, accounts)
-    add_legacy_accounts(accounts)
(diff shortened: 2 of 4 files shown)


deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -262,25 +262,43 @@ def update_deploy_progress(request, name, stage):
...
             "name": f"{AWS_PRIMARY_ACCOUNT} / Primary AWS account / Primary account",
             "ownerId": AWS_PRIMARY_ACCOUNT,
-        })
-    if not skip_sub_account:
-        accounts.append({
-            "name": f"{AWS_SUB_ACCOUNT} / Sub AWS account / Legacy sub account",
+        }
+    if account_id == AWS_SUB_ACCOUNT:
+        return {
+            "name": f"{AWS_SUB_ACCOUNT} / Sub AWS account / AWS account with manually provisioned cluster",
```
ntascii, line 291: / Moka account / AWS account for Spark on EKS

deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
```diff
@@ -279,21 +279,23 @@
...
-    </div>
-    <div class="panel-body">
-        <div class="row">
-            {% for account in accounts %}
-                <a href="#" class="deployToolTip btn btn-xs btn-default host-btn">
-                    <small>{{ account.name }}</small>
-                    <i class="fa fa-fw fa-user"></i>
-                </a>
-            {% endfor %}
+{% if accounts %}
```
ntascii, line 282: add a quick description for the account section such as "AWS account is being used in this stage". Main purpose of the section is to show which account the hosts belong to right?

vitalii-honchar: Added fixes:
1. Moved the `Account` panel before the `Hosts` panel.
2. Added tooltip.

### Before mouse hover on the tooltip
Screenshot 2024-03-14 at 2 06 51 PM

### After mouse hover on the tooltip
Screenshot 2024-03-14 at 2 07 15 PM

ntascii: we might need to make the style of Build section and Account section to be similar, looks like Account is bigger than build while build is the one people care more about.

ntascii: besides, I think it's better to make the account section to not consume lots of area here

ntascii (approved):

deploy-board/deploy_board/templates/deploys/deploy_progress.tmpl
vitalii-honchar: I changed UI: showed in the meeting

ntascii (approved):

PR #1523, opened 2024-03-15
Change audit log behavior

tylerwowen:
After digging into the dropwizard bundled reqest log more, I've figured out a better solution for audit logs. 

I'm now able to inject the user name, which was the only missing piece of info we need, to the request context and update the request log config to log something like

```
{"timestamp":"2024-03-15T01:06:12.931+0000","method":"GET","protocol":"HTTP/1.1","remoteAddress":"127.0.0.1","uri":"/v1/envs/names","status":500,"principal":{"name":"Anonymous"}}
```


deploy-service/teletraanservice/bin/server.yaml
```diff
@@ -13,10 +13,17 @@ server:
...
+        archivedFileCount: 5
         currentLogFilename: /tmp/teletraan/access.log
-        timeZone: UTC
+        layout:
+          type: access-json
+          timestampFormat: "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
+          requestAttributes: [name]
+          customFieldNames:
+            requestAttributes: principal
+          includes: [timestamp, remoteAddress, protocol, method, requestUri, statusCode]
```
vitalii-honchar, line 26: Can we add environment type in the log message?

Example of environments and environments types:
- Production:  `env: production`
- Stage: `env: stage`
- Dev: `env: dev`

And specify environment typ as an environment variable during deployment of the Teletraan service, like `export ENV=production` which will make easier to separate logs from different environments in Kibana.

tylerwowen: We have other mechanisms for separating them in Kibana, such as `o11y_hostname`.

vitalii-honchar (approved):

PR #1524, opened 2024-03-15
add account id column to deploy views

rwxzhu:
Add account id field to deploy details.

### Manual Testing
visit deployment details page at `/deploy/<deploy_id>`
#### "legacy" account, from agent report
Screenshot 2024-03-19 at 9 51 30 AM

#### account info from cluster/rodimus
Screenshot 2024-03-19 at 9 52 31 AM

#### unable to find in report or cluster
Screenshot 2024-03-19 at 12 44 12 PM


deploy-board/deploy_board/templates/deploys/deploy_details.tmpl
```diff
@@ -83,6 +83,12 @@
             {% include "builds/build_details.tmpl" %}
         </td>
     </tr>
+    <tr>
+        <td>Account</td>
+        <td>
+            <a href="/clouds/accounts/{{ account_info.cloudProvider }}/{{ account_info.cell }}/{{ account_info.id }}">{{ account_info.id }}</a>
```
vitalii-honchar, line 89: Let's hide this link if account is `None` for backward compatibility in cases when Deploy Board will be deployed before cluster management service.

Or if deployment was done in the cluster which was provisioned manually.

deploy-board/deploy_board/webapp/helpers/accounts_helper.py
```diff
@@ -15,11 +15,14 @@ def get_all_accounts(request):
...
-        return rodimus_client.get(f"/accounts/AWS/{cell}/{account_id}", request.teletraan_user_id.token)
+        return rodimus_client.get(f"/accounts/{provider}/{cell}/{account_id}", request.teletraan_user_id.token)
     except Exception as e:
         log.error(f"Can't get account by cell and account_id: "
                   f"cell = {cell}, account_id = {account_id}, error = {e}")
         # return None for backward-compatibility
         return None
+    
+def get_default_account(request, cell, provider='AWS'):
+    return get_by_cell_and_id(request, cell, "default", provider)
```
vitalii-honchar, line 28: Let's add `try except` for backward compatibility: deploy board may be deployed before cluster management service, that's why `get_default_account` will return 404 error.

So please change this function to:
```Python
def get_default_account(request, cell, provider="AWS"):
    try:
        return get_by_cell_and_id(request, cell, "default", provider)
    except Exception as e:
        # log exception
        return None
```

deploy-board/deploy_board/templates/deploys/deploy_details.tmpl
rwxzhu: updated, thanks

deploy-board/deploy_board/webapp/helpers/accounts_helper.py
rwxzhu: marking as resolved since exceptions should be caught by `get_by_cell_and_id`

deploy-board/deploy_board/templates/deploys/deploy_details.tmpl
```diff
@@ -83,6 +83,22 @@
             {% include "builds/build_details.tmpl" %}
         </td>
     </tr>
+    <tr>
```
vitalii-honchar, line 86: Let's hide `account` column if `deploy_accounts` list is empty:
```
{% if deploy_accounts %}
        <td>Account</td>
        <td>
            {% for account in deploy_accounts %}
                {% if account.legacy_name %}
                <div class="deployToolTip btn btn-xs btn-default host-btn">
                {{ account.legacy_name }}
                </div>
                {% else %}
                <a class="deployToolTip btn btn-xs btn-default host-btn" href="/clouds/accounts/{{ account.cloudProvider }}/{{ account.cell }}/{{ account.id }}">
                    {{ account.data.ownerId }} / {{ account.name }}
                </a>
                {% endif %}
            {% endfor %}
        </td>
    </tr>
{% endif %}

```

deploy-board/deploy_board/webapp/account_utils.py
```diff
@@ -0,0 +1,54 @@
+from deploy_board.settings import AWS_PRIMARY_ACCOUNT, AWS_SUB_ACCOUNT
+from .helpers import clusters_helper, accounts_helper, deploys_helper
+from . import agent_report
+
+def create_legacy_ui_account(account_id):
```
vitalii-honchar, line 5: `util` naming is an anti pattern. Let's use more meaningful names because `util` may means anything.
Maybe rename file to just `accounts.py`?

deploy-board/deploy_board/templates/deploys/deploy_details.tmpl
rwxzhu: updated, thanks

vitalii-honchar (approved):
LGTM

PR #1527, opened 2024-03-18
Added warning message for old base images of sub-account has chosen

vitalii-honchar:
# Context

Next: https://github.com/pinterest/teletraan/pull/1528

Added warning message for old base images of sub-account has chosen.

- Display warning message if chosen image is oldest than 2024-03-15 and sub-account has chosen.

# Tests

## Displayed warning message

Note: date has converted in my timezone, that's why it's 2024-03-14

Screenshot 2024-03-18 at 12 50 32 PM

## No warning message

Screenshot 2024-03-18 at 12 53 37 PM

## Two warning messages

Screenshot 2024-03-18 at 12 55 30 PM

### DIFF
--- a/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
+++ b/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
@@ -83,24 +83,21 @@ Vue.component('baseimage-select', {
             warningText: '',
         }
     },
-    props: ['imageNames', 'baseImages', 'selectedImageName', 'selectedBaseImage', 'pinImage', 'pinImageEnabled'],
+    props: ['imageNames', 'baseImages', 'selectedImageName', 'selectedBaseImage', 'pinImage', 'pinImageEnabled', 'accountOwnerId'],
     methods: {
         helpClick: function () {
             this.$emit('help-clicked')
         },
-        tryShowWarning: function (baseImageId, pinImage) {
+        tryShowWarning: function (baseImageId, pinImage, accountOwnerId) {
             if (!pinImage) {
                 this.showWarning = false;
             } else {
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
```diff
@@ -83,24 +83,21 @@ Vue.component('baseimage-select', {
...
-                let baseImage = this.baseImages.find(i => i.value == baseImageId);
-                let age = Math.round((Date.now() - new Date(baseImage.publishDate)) / ONE_DAY);
-                if (age > 180) {
-                    this.warningText = `The base image configured is over 180 days old (${age} days). Please consider update or opt-in Auto Update (Unpin image).`;
-                    this.showWarning = true;
-                } else {
-                    this.showWarning = false;
-                }
+                const baseImage = this.baseImages.find(i => i.value === baseImageId);
+                const warnings = this.createWarningMessages(baseImage, accountOwnerId);
```
ntascii, line 96: we might need to have a common var_name for this `account_number` or `acccount_owner_id` across teletraan and rodimus repos.

ntascii (approved):

deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
vitalii-honchar: Rodimus already uses `ownerId` naming, so I used it here, but I think that `account_number` looks more naturally for AWS users, so maybe we should rename all `ownerId` to `account_number`, but this requires some amount of work, so let's move it out of the scope of current project.

I will create a task for it.

PR #1539, opened 2024-03-22
update default account name

rwxzhu:
Change the default legacy account name from AWS Primary Account to default. Also update the styling for legacy account buttons to have disabled styling to signal they are not clickable.

Screenshot 2024-03-22 at 11 32 00 AM

### DIFF
--- a/deploy-board/deploy_board/webapp/accounts.py
+++ b/deploy-board/deploy_board/webapp/accounts.py
@@ -5,7 +5,7 @@ from . import agent_report
 def create_legacy_ui_account(account_id):
     if account_id == AWS_PRIMARY_ACCOUNT:
         return {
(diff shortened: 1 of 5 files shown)


deploy-board/deploy_board/webapp/accounts.py
```diff
@@ -5,7 +5,7 @@
 def create_legacy_ui_account(account_id):
     if account_id == AWS_PRIMARY_ACCOUNT:
         return {
-            "legacy_name": f"{AWS_PRIMARY_ACCOUNT} / Primary AWS account",
```
ntascii, line 8: I think we should do the other way around, ie. keep Primary AWS account and make all the `/ default` one to be ` Primary AWS account`

rwxzhu: sounds good, I'll address this in rodimus then since that's where the default value comes from.

PR #1544, opened 2024-03-26
Updates to support other cluster/group name formats

rfleur01:
## Background: 
For legacy groups that are converted to clusters, the cluster name needs to match the ASG and the group in Teletraan. This can cause the UI to return an error because the format of the ASG is not "{env_name}-{stage_name}". It could be "{env_name} {stage_name}"

## Tests:
Test all the changes made to ensure the new format is supported and there is no regression for the existing format.
New Format Cluster: helloworlddummyservice-server-rfleur
Old Format Cluster: rfleur-dev1

<details><summary>Show Replacement Activities</summary>
<p>
image
image
</p>
</details> 

<details><summary>Auto Scaling Settings/Auto Scaling Configuration</summary>
<p>
image
image
</p>
</details> 

<details><summary>Update Capacity</summary>
<p>
image
image
image
image
</p>
</details> 

<details><summary>Click to see AMI(s) used by Hosts</summary>
<p>
image
image
</p>
</details> 

<details><summary>Host Terminate</summary>
<p>
image
image
</p>
</details> 

<details><summary>Cluster Config GET</summary>
<p>
image
image
</p>
</details> 

<details><summary>Cluster Config POST</summary>
<p>
image
image
</p>
</details> 

<details><summary>Cluster Config History</summary>
<p>
image
image
</p>
</details> 

<details><summary>Clone Cluster</summary>
<p>
image
image

image
image

</p>
</details> 

<details><summary>Enable Cluster Replacement</summary>
<p>
image
image
</p>
</details> 

<details><summary>Auto Cluster Refresh View</summary>
<p>
image
image
</p>
</details> 

<details><summary>Cluster Replacement Details</summary>
<p>
image
image
</p>
</details> 

<details><summary>Base Image History</summary>
<p>
image
image
</p>
</details> 

<details><summary>Cluster Refresh Status</summary>
<p>
image
image

</p>
</details>


deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -610,8 +611,9 @@ def _gen_message_for_refreshing_cluster(request, last_cluster_refresh_status, la
 
 def _getLastClusterRefreshStatus(request, env):
     try:
+        cluster_name = get_cluster_name(request, env.name, env.stage)
         replace_summaries = clusters_helper.get_cluster_replacement_status(
-            request, data={"clusterName": '{}-{}'.format(env["envName"], env["stageName"])})
+            request, data={"clusterName": '{}'.format(cluster_name)})
```
rwxzhu, line 616: nit: we can replace this with
```suggestion
            request, data={"clusterName": 'cluster_name})
```

deploy-board/deploy_board/webapp/cluster_view.py
```diff
@@ -1448,3 +1446,21 @@ def get(self, request, name, stage):
...
 
         return render(request, 'clusters/base_image_history.html', data)
+
+def get_cluster_name(request, name, stage):
+    cluster_name = '{}-{}'.format(name, stage)
+    current_cluster = clusters_helper.get_cluster(request, cluster_name)
+    if current_cluster is None:
+        env = environs_helper.get_env_by_stage(request, name, stage)
+        cluster_name = env.get("clusterName")
+    return cluster_name
```
rwxzhu, line 1456: do we need the check for `{}-{}` format cluster existing, or could we change this to 
```suggestion
def get_cluster_name(request, name, stage):
    env = environs_helper.get_env_by_stage(request, name, stage)
    cluster_name = env.get("clusterName")
    return cluster_name
```

deploy-board/deploy_board/webapp/env_views.py
liyaqin1: There is a typo, no single quote needed ^^

liyaqin1:
@rfleur01 Could you clarify the purpose of the cluster name? We already have the ASG name in rodimus, as well as the environment name and stage name in deploy. What specific role does the cluster name play? Thank you.

deploy-board/deploy_board/webapp/cluster_view.py
rfleur01: Not necessarily. I left it just to be extra careful. If we entirely depend on the env to lookup a cluster there could be scenario where the cluster name on the env does not exist or is a mismatch with the actual cluster or an API issue that prevents the lookup. Not likely, but still possible. If we don't have that link for any reason we still want to be able to get/update those clusters.

rfleur01:
@liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.

deploy-board/deploy_board/webapp/env_views.py
```diff
@@ -1952,3 +1954,19 @@ def override_session(request, name, stage):
     session_num = request.GET.get('session_num')
     schedules_helper.override_session(request, name, stage, session_num)
     return HttpResponse(json.dumps(''))
+
+def get_cluster_name(request, name, stage):
+    cluster_name = '{}-{}'.format(name, stage)
+    current_cluster = clusters_helper.get_cluster(request, cluster_name)
+    if current_cluster is None:
+        env = environs_helper.get_env_by_stage(request, name, stage)
```
rwxzhu, line 1962: nit: it looks like for a couple of places where the new methods are called, the env is already available before the method is called. Could we pass in the env instead, and save an api call? Something like
```suggestion
def get_cluster_name(request, name, stage, env=None):
    cluster_name = '{}-{}'.format(name, stage)
    current_cluster = clusters_helper.get_cluster(request, cluster_name)
    if current_cluster is None:
        if env is None:
            env = environs_helper.get_env_by_stage(request, name, stage)
```

rfleur01: yeah good call

rfleur01: updated

liyaqin1:
> @liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.

How do we generate the cluster name?

rfleur01:
> > @liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.
> 
> How do we generate the cluster name?

This does change how cluster names are generated. New clusters will still be generated in the {env name}-{stage name} format. It only adds support to get and update clusters that do not have the {env name}-{stage name}.

liyaqin1:
> > > @liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.
> > 
> > 
> > How do we generate the cluster name?
> 
> This does change how cluster names are generated. New clusters will still be generated in the {env name}-{stage name} format. It only adds support to get and update clusters that do not have the {env name}-{stage name}.

If the env/stage is lacking a cluster name generated as {env name}-{stage name}, what alternative method should be used for generating it?

rfleur01:
> > > > @liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.
> > > 
> > > 
> > > How do we generate the cluster name?
> > 
> > 
> > This does change how cluster names are generated. New clusters will still be generated in the {env name}-{stage name} format. It only adds support to get and update clusters that do not have the {env name}-{stage name}.
> 
> If the env/stage is lacking a cluster name generated as {env name}-{stage name}, what alternative method should be used for generating it?

This is related to the work I am doing to convert legacy groups to clusters. Many of these groups have ASGs that are not in the {env name}-{stage name} format, which prevents the UI from rendering their cluster configurations and other UI components. In short, they just  wouldn't be generated via the UI. They could be generated via API OR in this case via DB insertion (via script), which is the main reason for this PR.

liyaqin1:
> > > > > @liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.
> > > > 
> > > > 
> > > > How do we generate the cluster name?
> > > 
> > > 
> > > This does change how cluster names are generated. New clusters will still be generated in the {env name}-{stage name} format. It only adds support to get and update clusters that do not have the {env name}-{stage name}.
> > 
> > 
> > If the env/stage is lacking a cluster name generated as {env name}-{stage name}, what alternative method should be used for generating it?
> 
> This is related to the work I am doing to convert legacy groups to clusters. Many of these groups have ASGs that are not in the {env name}-{stage name} format, which prevents the UI from rendering their cluster configurations and other UI components. In short, they just wouldn't be generated via the UI. They could be generated via API OR in this case via DB insertion (via script), which is the main reason for this PR.

My question is, how can you determine the actual ASG name if it is not generated through Teletraan and thus does not follow the {env name}-{stage name} convention?

rfleur01:
> > > > > > @liyaqin1 using the format {env name}-{stage name} causes clusters that do not follow that format to not be accessible via the UI. This update leans on the cluster_name to account for names that might not be in the {env name}-{stage name}.
> > > > > 
> > > > > 
> > > > > How do we generate the cluster name?
> > > > 
> > > > 
> > > > This does change how cluster names are generated. New clusters will still be generated in the {env name}-{stage name} format. It only adds support to get and update clusters that do not have the {env name}-{stage name}.
> > > 
> > > 
> > > If the env/stage is lacking a cluster name generated as {env name}-{stage name}, what alternative method should be used for generating it?
> > 
> > 
> > This is related to the work I am doing to convert legacy groups to clusters. Many of these groups have ASGs that are not in the {env name}-{stage name} format, which prevents the UI from rendering their cluster configurations and other UI components. In short, they just wouldn't be generated via the UI. They could be generated via API OR in this case via DB insertion (via script), which is the main reason for this PR.
> 
> My question is, how can you determine the actual ASG name if it is not generated through Teletraan and thus does not follow the {env name}-{stage name} convention?

It's determined via group_name in the groups table. The script uses that value as the cluster name when creating the cluster record. The script also populates the cluster_name value at the env level so this PR leverages that value.

rwxzhu (approved):
lgtm

PR #1557, opened 2024-04-01
send metrics via metrics-agent instead of calling api

liyaqin1:


liyaqin1:
Tested on test stage ok: https://deploy.pinadmin.com/env/deploy-init/test/
Got metrics here:
https://statsboard.pinadmin.com/share/vtqwv

ntascii (approved):

PR #1563, opened 2024-04-04
added default pinfo_environment prod

Cjpilbdev:
Prevent user from using non-prod pinfo_environment for cmp_base clusters required in https://jira.pinadmin.com/browse/CDP-6704

### DIFF
--- a/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
+++ b/deploy-board/deploy_board/templates/configs/new_capacity_adv.html
@@ -121,6 +121,14 @@ const defaultImageId = goldenImage ? goldenImage.id : getDefaultBaseImageId(base
 
 const hieraPaths = `{{ puppet_hiera_paths }}`;
 const puppetRepository = `{{ puppet_repository }}`;
+var setPinfoEnvironment = function(value, allUserData){
+    if(value.startsWith('cmp_base')){
+        var pinfo_environment = allUserData.find(({name}) => name === "pinfo_environment");
+        if(pinfo_environment !== undefined){
+            pinfo_environment.value = 'prod';
+        }        
+    }
+};
 
 var capacitySetting = new Vue({
     el: "#mainPanel",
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/templates/configs/new_capacity_adv.html
```diff
@@ -121,6 +121,14 @@ <h4 class="panel-title pull-left">Capacity</h4>
 
 const hieraPaths = `{{ puppet_hiera_paths }}`;
 const puppetRepository = `{{ puppet_repository }}`;
+var setPinfoEnvironment = function(value, allUserData){
+    if(value.startsWith('cmp_base')){
+        var pinfo_environment = allUserData.find(({name}) => name === "pinfo_environment");
+        if(pinfo_environment !== undefined){
+            pinfo_environment.value = 'prod';
```
yongwen, line 128: in the teletraan-integ, i.e., https://deploy-integ.pinadmin.com/  which host the production teletraan envs, their pinfo env need to be "dev" to be able to connect to the integ agent-service.  so this need to be handled separately. 

Also, curious what is the use case of this change, any slack / jira ticket for the issue/case encountered?  as a best practice for a PR, you could also describe the use case, as well as testing done in the description. Thanks.

yongwen (changes requested):

deploy-board/deploy_board/templates/configs/new_capacity_adv.html
Cjpilbdev: the requirement is to use "prod" as the default value for cmp_base clusters.  However, "dev" can still be used, but it has to be modify by the user.

rfleur01: This was the ticket https://jira.pinadmin.com/browse/CDP-6704. It seems like at the very least we could have a warning if a user changes it to dev on a production cluster and the opposite for integ.

PR #1565, opened 2024-04-05
add sidecar navigation buttons and show status

rwxzhu:
update the sidecar section of the host details so that:
- sidecars have navigation buttons at the top of the sidecar section which scroll to the relevant deploy details.
- relevant sidecars are now indicated with a star instead of info colour
- sidecars with status `SUCCESS` show a green navigation button
- sidecars with status `TOO_MANY_RETRY, SCRIPT_TIMEOUT, AGENT_FAILED, AGENT_FAILED, SCRIPT_FAILED` show a red navigation button

### Manual Testing

tested on dev1
Screenshot 2024-04-05 at 11 01 10 AM


ntascii (approved):
nice :)

PR #1585, opened 2024-04-12
validate not trusted build urls and marked as private in non-dev systems

Cjpilbdev:
if build url is not in trusted list, it should automatically be marked as private by Teletraan and the deploy-agent should reject the build for non dev systems

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
@@ -384,7 +384,7 @@ public class DeployHandler implements DeployHandlerInterface{
         5. no private build for sox env
         6. only sox build can be deployed for sox env
     */
-    private void validateBuild(EnvironBean envBean, String buildId) throws Exception {
+    protected void validateBuild(EnvironBean envBean, String buildId) throws Exception {
         if(StringUtils.isEmpty(buildId)) {
             throw new WebApplicationException("Build id can not be empty.", Response.Status.BAD_REQUEST);
         }
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
@@ -442,8 +442,8 @@ public class DeployHandler implements DeployHandlerInterface{
     }
 
     private boolean isPrivateBuild(BuildBean buildBean) {
-        return buildBean.getScm_branch() != null
-                && buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH);
+        return buildBean.getScm_branch() == null
+                || buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH);
     }
 
     public String deploy(EnvironBean envBean, String buildId, String desc, String deliveryType, String operator) throws Exception {
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
@@ -395,11 +395,10 @@ public class DeployHandler implements DeployHandlerInterface{
             throw new DeployInternalException("Build name (%s) does not match stage config (%s).",
                     buildBean.getBuild_name(), envBean.getBuild_name());
         }
-        
+
         if(envBean.getStage_type() != EnvType.DEV &&
                 buildAllowlist != null && !buildAllowlist.trusted(buildBean.getArtifact_url())
-                && buildBean.getScm_branch() != null
-                && !buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH)) {
+                && isPrivateBuild(buildBean)){
             buildBean.setScm_branch(PRIVATE_BUILD_SCM_BRANCH);
         }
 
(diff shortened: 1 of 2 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -395,6 +395,13 @@ private void validateBuild(EnvironBean envBean, String buildId) throws Exception
             throw new DeployInternalException("Build name (%s) does not match stage config (%s).",
                     buildBean.getBuild_name(), envBean.getBuild_name());
         }
+        
+        if(envBean.getStage_type() != EnvType.DEV &&
+                buildAllowlist != null && !buildAllowlist.trusted(buildBean.getArtifact_url())
+                &&  !buildBean.getScm_branch().equalsIgnoreCase("private")) {
```
rwxzhu, line 401:
```suggestion
                && !buildBean.getScm_branch().equalsIgnoreCase("private")) {
```
nit: remove extra space

tylerwowen: you want the opposite to prevent NPE.
` "private".equalsIgnoreCase(buildBean.getScm_branch())`

tylerwowen: Then it's the best to create a constant for `"private"`.

tylerwowen: I founds that pinboard uses `private_build` instead of `private`. It means they will bypass this check. We need to investigate more.

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -395,6 +395,14 @@ private void validateBuild(EnvironBean envBean, String buildId) throws Exception
...
+        
+        if(envBean.getStage_type() != EnvType.DEV &&
+                buildAllowlist != null && !buildAllowlist.trusted(buildBean.getArtifact_url())
+                && buildBean.getScm_branch() != null
+                && !buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH)) {
+            buildBean.setScm_branch(PRIVATE_BUILD_SCM_BRANCH);
+        }
+
         // only allow a non-private deploy if the build is from a trusted artifact url
         if(envBean.getEnsure_trusted_build() && !buildBean.getScm_branch().equals("private") &&
```
tylerwowen, line 407: Let's make a more complete change.
1. Create a new method `isPrivateBuild(BuildBean)` and returns `buildBean.getScm_branch() != null
                && !buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH)`
2. Replace checks like `!buildBean.getScm_branch().equals("private")` to the new method. 

So we have a consistent way of telling if a build is a private build.

tylerwowen: I'd like to see test coverage for these changes.

deploy-service/common/src/test/java/com/pinterest/deployservice/handler/DeployHandlerTest.java
tylerwowen, line 1: Format this file and add license.

deploy-service/common/src/test/java/com/pinterest/deployservice/handler/DeployHandlerTest.java
```diff
@@ -0,0 +1,205 @@
...
+        try {
+            deployHandler.validateBuild(envBean, "");
+            fail(String.format(FAIL_MESSAGE,
+                    ERROR_EMPTY_BUILD_ID
+            ));
+        } catch (WebApplicationException ex){
+            assertTrue(ex.getMessage().contains(
+                    ERROR_EMPTY_BUILD_ID
+            ));
+        }
```
tylerwowen, line 87: Consider this pattern
```java
Throwable throwable =  assertThrows(WebApplicationException.class, () -> xxx);
assertTrue(ex.getMessage().contains(
                    ERROR_EMPTY_BUILD_ID
            ));
```

tylerwowen (changes requested):

tylerwowen (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/DeployHandler.java
```diff
@@ -384,47 +393,59 @@ public void disableAutoPromote(EnvironBean envBean, String operator, boolean for
...
             throw new WebApplicationException(
-                    "This stage requires SOX builds. The build must be from a sox-compliant source. Contact your sox administrators.",
+                    ERROR_STAGE_REQUIRES_SOX_BUILD_COMPLIANT_SOURCE,
                     Response.Status.BAD_REQUEST);
         }
     }
 
+    private boolean isPrivateBuild(BuildBean buildBean) {
+        return buildBean.getScm_branch() != null
+                && buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH);
```
tylerwowen, line 446: I think we should mark null buildBean.getScm_branch() as private build
```
return buildBean.getScm_branch() == null
                || buildBean.getScm_branch().toLowerCase().startsWith(PRIVATE_BUILD_SCM_BRANCH);
```

tylerwowen (approved):

PR #1600, opened 2024-04-25
Create a new API to mark/unmark a pindeploy pipeline on a Teletraan env/stage

liyaqin1:


liyaqin1:
After testing on dev1, when I enable a stage to be a Keel pipeline managed one, the following entry is added to the database:
```
mysql> select * from pindeploy;
+------------------------+--------------+----------+
| env_id                 | is_pindeploy | pipeline |
+------------------------+--------------+----------+
| f3LLmLiaTnODsyMwHF_S3g |            1 | xyz      |
+------------------------+--------------+----------+
1 row in set (0.01 sec)
```
The database also reflects the change after I disable it:
```
mysql> select * from pindeploy;
+------------------------+--------------+----------+
| env_id                 | is_pindeploy | pipeline |
+------------------------+--------------+----------+
| f3LLmLiaTnODsyMwHF_S3g |            0 | abc      |
+------------------------+--------------+----------+
1 row in set (0.00 sec)

```

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBPindeployDAOImpl.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.commons.dbutils.handlers.BeanHandler;
+
+/* count table cache to store # of actively deploying agents and # of existing agents */
+public class DBPindeployDAOImpl implements PindeployDAO {
+    private static final String GET_PINDEPLOY =
+        "SELECT * FROM pindeploy WHERE env_id=?";
+    private static final String DELETE_PINDEPLOY =
+        "DELETE * FROM pindeploy WHERE env_id=?";
+    private static final String INSERT_OR_UPDATE_PINDEPLOY =
+        "INSERT INTO pindeploy SET %s ON DUPLICATE KEY UPDATE %s";
```
mitchellngoodman, line 33: We had a recent security risk for a query using %s in pindeploy, fix was to use ?, can see the commit here https://github.com/pinternal/optimus/commit/af9794a2e6c8ad2ba2305a24b9f92df9bbe69b82

mitchellngoodman: That might not be relevant since it looks like the clause can't be set to just anything

mitchellngoodman (approved):
LGTM, but do you think it needs to use the IS-PINTEREST config flag?

liyaqin1:
> LGTM, but do you think it needs to use the IS-PINTEREST config flag?

I doubt it, as there are no Pinterest configurations present. Additionally, there is a Pinterest-related API such as @Path("/is-sox/{booleanValue}") in the same file, EnvStages.java.

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBPindeployDAOImpl.java
liyaqin1: Never mind, I updated it to enhance security. ^^

mitchellngoodman (approved):

PR #1604, opened 2024-04-29
Allow Stage Type change from non-prod to prod

rfleur01:
## Background:
Stage Type is immutable in Teletraan, which makes it difficult for users to update their Stage Types if they made a mistake. Furthermore, if there is a mismatch between the stage type in Teletraan and the stage type in PinDeploy.

## Test Plan

<details><summary>Save with DEFAULT (PROD)</summary>
<p>
image
</p>

</details> 

<details><summary>DEFAULT to PROD</summary>
<p>
image
</p>
</details> 

<details><summary>DEFAULT (PROD) to PROD</summary>
<p>
image
image
</p>
</details> 

<details><summary>PROD to NON-PROD</summary>
<p>
image
</p>
</details> 

<details><summary>DEFAULT to NON-PROD</summary>
<p>
image
</p>
</details> 

<details><summary>NON-PROD to NON-PROD</summary>
<p>
image
</p>
</details> 

<details><summary>NON-PROD to PROD</summary>
<p>
image
</p>
</details> 

<details><summary>Set to DEFAULT</summary>
<p>
image
</p>
</details>

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
@@ -22,6 +22,9 @@ import org.apache.commons.lang.StringEscapeUtils;
 import org.hibernate.validator.constraints.Range;
 
 import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
 import javax.validation.constraints.NotEmpty;
 
 /**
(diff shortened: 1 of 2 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/bean/EnvironBean.java
```diff
@@ -228,10 +231,35 @@ public void validate() throws IllegalArgumentException {
         if (this.chatroom != null && !this.chatroom.matches(chatRegex)) {
             throw new IllegalArgumentException(String.format("Chatroom must match regex %s", chatRegex));
         }
+    }
+
+    public void stageTypeValidate(EnvironBean origBean) throws Exception {
```
yongwen, line 236: i wonder if it might be better to move this check with "origBean" out to the envStages.java file as it is a very specific business logic. leaving "this.stage_type == EnvType.DEFAULT" check here is fine as it is part of the standard data model validation,i.e, no origbean info is needed.

rfleur01: I updated it so that all the business logic is in EnvStages

yongwen (approved):

PR #1605, opened 2024-04-29
Add an API to get pindeploy related stuff according to the environment id

liyaqin1:


liyaqin1:
Tested on dev1, looks good:
Screenshot 2024-04-29 at 4 01 14 PM

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
```diff
@@ -97,6 +101,19 @@ public EnvironBean get(
         return Utils.getEnvStage(environDAO, envName, stageName);
     }
 
+    @GET
+    @Path("/keelPipeline")
```
mitchellngoodman, line 105: Could it make sense to use `/pindeployPipeline` in the url to be consistent when the rest of the terminology used here?

mitchellngoodman (dismissed):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
liyaqin1: yes, updated, thank you

mitchellngoodman (approved):

PR #1606, opened 2024-04-29
upgraded library json-path-upgrade

Cjpilbdev:
upgrade library because of vulnerabilities.

### DIFF
--- a/deploy-service/common/pom.xml
+++ b/deploy-service/common/pom.xml
@@ -110,16 +110,6 @@
             <version>2.8.0</version>
         </dependency>
         <dependency>
-            <groupId>com.jayway.jsonpath</groupId>
-            <artifactId>json-path</artifactId>
-            <version>2.9.0</version>
-        </dependency>
-        <dependency>
-            <groupId>net.minidev</groupId>
-            <artifactId>json-smart</artifactId>
-            <version>2.5.1</version>
-        </dependency>
-        <dependency>
             <groupId>com.slack.api</groupId>
             <artifactId>slack-api-client</artifactId>
             <version>1.11.0</version>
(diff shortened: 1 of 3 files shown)


deploy-service/common/pom.xml
```diff
@@ -112,7 +112,12 @@
         <dependency>
             <groupId>com.jayway.jsonpath</groupId>
             <artifactId>json-path</artifactId>
-            <version>0.9.1</version>
+            <version>2.9.0</version>
+        </dependency>
+        <dependency>
+            <groupId>net.minidev</groupId>
+            <artifactId>json-smart</artifactId>
+            <version>2.5.1</version>
```
gzpcho, line 120: Any breaking changes that might impact existing functionality?

tylerwowen:
I was thinking of switching to jackson for json processing but that requires code change. If we need code change anyways, then we should just reimplement with jackson and remove the dependency.

deploy-service/common/src/main/java/com/pinterest/deployservice/common/DBConfigReader.java
```diff
@@ -81,10 +83,22 @@ public Integer getPort(String replicaSetName) {
     }
 
     public String getUsername(String role) {
-        return JsonPath.read(credJson, String.format("$.%s.users[?(@.enabled == true)][0].username", role));
+        JSONArray jsonArray = JsonPath.read(credJson, String.format("$.%s.users[?(@.enabled == true)]", role));
+        if(jsonArray.size()>0){
+            Map hashMap =  (Map) jsonArray.get(0);
```
gzpcho, line 88: Should we not initialize a new `HashMap` with the `new` keyword? What's the structure of `jsonArray.get(0)`?

deploy-service/common/pom.xml
Cjpilbdev: this is only used for getting the database configuration from json  in DBConfigReader

tylerwowen (commented):
LGTM.
Can you run `mvn spotless:apply` and only commit the changes to `deploy-service/common/src/main/java/com/pinterest/deployservice/common/DBConfigReader.java`?

PR #1608, opened 2024-04-30
minor updated libraries

Cjpilbdev:
- Upgrade minor libraries:
-  skipped org.hibernate:hibernate-validator already in 6.2.5.Final 
- commons-beanutils not found 

image

image

### DIFF
--- a/deploy-service/common/pom.xml
+++ b/deploy-service/common/pom.xml
@@ -112,22 +112,7 @@
         <dependency>
             <groupId>com.slack.api</groupId>
             <artifactId>slack-api-client</artifactId>
-            <version>1.11.0</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.squareup.okhttp3</groupId>
-                    <artifactId>okhttp</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>com.google.code.gson</groupId>
-                    <artifactId>gson</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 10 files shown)


deploy-service/common/pom.xml
```diff
@@ -154,6 +169,17 @@
...
             <version>1.19.7</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-compress</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
```
tylerwowen, line 179: It was removed https://github.com/pinterest/teletraan/pull/1579/files

deploy-service/common/pom.xml
```diff
@@ -118,6 +118,21 @@
...
+                <exclusion>
+                    <groupId>com.google.code.gson</groupId>
+                    <artifactId>gson</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>4.12.0</version>
```
tylerwowen, line 135: I think dependencyManagement is a better approach then pinning transient dependency.

tylerwowen: Or even better, can we upgrade slack-api-client?

Cjpilbdev: updated

tylerwowen (approved):

PR #1610, opened 2024-05-01
Address an AuthZ flaw

tylerwowen:
Previously Teletraan looks for the request body for AuthZ resource identifier. However it has a security flaw for some endpoints. This PR address the issue by using the identifier from request path parameter for those affected resources.

## Details of the issue
Some Teletraan endpoints follows a pattern that requires the resource identifier in the path parameter, while the request body also contains the same ID field. The ID in the path parameter is used for later processing. Therefore it's possible for a request to contain 2 resource IDs and the one used for authorization can be different from the resource being changed. 

## Changes
First of all some interface change. Previously the annotation `ResourceAuthZInfo.type` expects the type of the extracted resource. Now it's the type of the input resource. This will reduce some confusion because it matches the method it annotates. 

When appropriate, a path resource extractor replaces the previously used body resource extractor. New implementations of resource extractors are added.

The most change happen in the [EnvStageBodyExtractor.java](https://github.com/pinterest/teletraan/pull/1610/files#diff-a49e40979c111009fe3e8ecb19415406d092e63d99033e04077178dd4ec45f81). This is the result of the interface change. The implementation are split into individual extractors. 

## Tests and validation
Close to 100% test coverage for this change.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBEnvironDAOImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBEnvironDAOImpl.java
@@ -28,6 +28,7 @@ import org.apache.commons.lang.StringUtils;
 
 import java.util.Collection;
 import java.util.HashSet;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
@@ -47,6 +48,8 @@ public class DBEnvironDAOImpl implements EnvironDAO {
         "UPDATE environs SET external_id=? WHERE env_name=? AND stage_name=?";
     private static final String GET_ENV_BY_ID =
         "SELECT * FROM environs WHERE env_id=?";
+    private static final String GET_ENV_BY_DEPLOY_ID =
+        "SELECT e.* FROM environs e INNER JOIN deploys d ON e.env_id = d.env_id WHERE d.deploy_id=?";
     private static final String GET_ENV_BY_NAME =
         "SELECT * FROM environs WHERE env_name=?";
     private static final String GET_ENV_BY_STAGE =
@@ -180,8 +183,14 @@ public class DBEnvironDAOImpl implements EnvironDAO {
 
     @Override
     public EnvironBean getById(String envId) throws Exception {
-        ResultSetHandler<EnvironBean> h = new BeanHandler<EnvironBean>(EnvironBean.class);
+        ResultSetHandler<EnvironBean> h = new BeanHandler<>(EnvironBean.class);
         return new QueryRunner(dataSource).query(GET_ENV_BY_ID, h, envId);
+    }
+
+    @Override
+    public EnvironBean getByDeployId(String deployId) throws SQLException {
+        ResultSetHandler<EnvironBean> h = new BeanHandler<>(EnvironBean.class);
+        return new QueryRunner(dataSource).query(GET_ENV_BY_DEPLOY_ID, h, deployId);
     }
 
     @Override
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/EnvStageBodyExtractor.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/EnvStageBodyExtractor.java
@@ -19,12 +19,9 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.pinterest.deployservice.bean.EnvironBean;
 import com.pinterest.teletraan.universal.security.AuthZResourceExtractor;
 import com.pinterest.teletraan.universal.security.bean.AuthZResource;
-
 import java.io.IOException;
 import java.io.InputStream;
-
 import javax.ws.rs.container.ContainerRequestContext;
-
 import org.glassfish.jersey.server.ContainerRequest;
 
 public class EnvStageBodyExtractor implements AuthZResourceExtractor {
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/EnvStageBodyExtractor.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/EnvStageBodyExtractor.java
@@ -18,7 +18,6 @@ package com.pinterest.teletraan.security;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.pinterest.deployservice.ServiceContext;
 import com.pinterest.deployservice.bean.AgentBean;
-import com.pinterest.deployservice.bean.DeployBean;
 import com.pinterest.deployservice.bean.EnvironBean;
 import com.pinterest.deployservice.bean.HotfixBean;
 import com.pinterest.deployservice.dao.EnvironDAO;
@@ -71,19 +70,6 @@ public class EnvStageBodyExtractor implements AuthZResourceExtractor {
             }
         }
 
-        if (DeployBean.class.equals(beanClass) || tryAll) {
-            try {
-                return extractDeployResource(inputStream);
-            } catch (Exception e) {
-                if (!tryAll) {
-                    if (e instanceof NotFoundException) {
-                        throw (NotFoundException) e;
-                    }
-                    throw new BeanClassExtractionException(beanClass, e);
-                }
-            }
-        }
-
         if (HotfixBean.class.equals(beanClass) || tryAll) {
             try {
                 return extractHotfixResource(inputStream);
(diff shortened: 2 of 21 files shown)


deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/security/bean/AuthZResource.java
```diff
@@ -77,16 +77,22 @@ public enum Type {
...
         IAM_ROLE,
         /** For build related resources. */
-        BUILD
+        BUILD,
+        /** For deploy related resources. */
+        DEPLOY,
+        /** For hotfix related resources. */
+        HOTFIX,
+        /** For Agent related resources. */
+        AGENT,
```
yongwen, line 95: seems like i can not find the AGENT related extractor.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/EnvStageBodyExtractor.java
```diff
@@ -16,151 +16,26 @@
...
-        } catch (Exception e) {
-            throw new ExtractionException("Failed to get hotfix bean", e);
-        }
-        if (hotfixBean == null) {
-            throw new NotFoundException(String.format("Hotfix(%s) not found", hotfixId));
-        }
-        return new AuthZResource(hotfixBean.getEnv_name(), "");
-    }
-
-    private AuthZResource extractAgentResource(InputStream inputStream) throws IOException, ExtractionException, NotFoundException {
```
yongwen, line 151: i did not find the corresponding extrator class for agent. should we keep this logic around

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/TeletraanAuthZResourceExtractorFactory.java
```diff
@@ -23,51 +23,58 @@ public class TeletraanAuthZResourceExtractorFactory implements AuthZResourceExtr
 
     private static final AuthZResourceExtractor BUILD_BODY_EXTRACTOR = new BuildBodyExtractor();
     private static final AuthZResourceExtractor ENV_PATH_EXTRACTOR = new EnvPathExtractor();
-    private static final AuthZResourceExtractor ENV_STAGE_PATH_EXTRACTOR = new EnvStagePathExtractor();
-    private final AuthZResourceExtractor BUILD_PATH_EXTRACTOR;
-    private final AuthZResourceExtractor ENV_STAGE_BODY_EXTRACTOR;
+    private static final AuthZResourceExtractor ENV_STAGE_BODY_EXTRACTOR =
+            new EnvStageBodyExtractor();
+    private static final AuthZResourceExtractor ENV_STAGE_PATH_EXTRACTOR =
+            new EnvStagePathExtractor();
```
rfleur01, line 29: What is the difference between these and 
```
  buildPathExtractor = new BuildPathExtractor(serviceContext);
  deployPathExtractor = new DeployPathExtractor(serviceContext);
  hotfixPathExtractor = new HotfixPathExtractor(serviceContext);
```

deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/security/bean/AuthZResource.java
tylerwowen: Yes it's in a separate PR https://github.com/pinterest/teletraan/pull/1489.
Beside, there are values to be consumed by Rodimus.

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/EnvStageBodyExtractor.java
tylerwowen: I will reimplement in https://github.com/pinterest/teletraan/pull/1489

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/TeletraanAuthZResourceExtractorFactory.java
tylerwowen: These are simple classes that don't need DBDAO in their constructor.

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBEnvironDAOImpl.java
```diff
@@ -47,6 +48,8 @@ public class DBEnvironDAOImpl implements EnvironDAO {
         "UPDATE environs SET external_id=? WHERE env_name=? AND stage_name=?";
     private static final String GET_ENV_BY_ID =
         "SELECT * FROM environs WHERE env_id=?";
+    private static final String GET_ENV_BY_DEPLOY_ID =
+        "SELECT e.* FROM environs e INNER JOIN deploys d ON e.env_id = d.env_id WHERE d.deploy_id=?";
```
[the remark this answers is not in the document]
tylerwowen: Thanks for being detail oriented. I compared the 2, since the key used are primary, we won't have performance impact. 

## When the deploy exists

### Current

```mysql
explain SELECT e.* FROM environs e INNER JOIN deploys d ON e.env_id = d.env_id WHERE d.deploy_id="0--4nnYXQMi7bXh-mbzVpA";
+----+-------------+-------+------------+-------+------------------------+---------+---------+-------+------+----------+-------+
| id | select_type | table | partitions | type  | possible_keys          | key     | key_len | ref   | rows | filtered | Extra |
+----+-------------+-------+------------+-------+------------------------+---------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | d     | NULL       | const | PRIMARY,deploy_env_idx | PRIMARY | 24      | const |    1 |   100.00 | NULL  |
|  1 | SIMPLE      | e     | NULL       | const | PRIMARY                | PRIMARY | 24      | const |    1 |   100.00 | NULL  |
+----+-------------+-------+------------+-------+------------------------+---------+---------+-------+------+----------+-------+
```

### Proposed
```mysql
explain SELECT e.* FROM environs e where e.env_id = (select d.env_id from deploys d WHERE d.deploy_id="0--4nnYXQMi7bXh-mbzVpA");
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
| id | select_type | table | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | Extra |
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
|  1 | PRIMARY     | e     | NULL       | const | PRIMARY       | PRIMARY | 24      | const |    1 |   100.00 | NULL  |
|  2 | SUBQUERY    | d     | NULL       | const | PRIMARY       | PRIMARY | 24      | const |    1 |   100.00 | NULL  |
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
```


## When the deploy doesn't exist

### Current
```mysql
explain SELECT e.* FROM environs e INNER JOIN deploys d ON e.env_id = d.env_id WHERE d.deploy_id="123"; 
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+--------------------------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra                          |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+--------------------------------+
|  1 | SIMPLE      | NULL  | NULL       | NULL | NULL          | NULL | NULL    | NULL | NULL |     NULL | no matching row in const table |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+--------------------------------+
```
### Proposed
```mysql
explain SELECT e.* FROM environs e where e.env_id = (select d.env_id from deploys d WHERE d.deploy_id="123");
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+--------------------------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra                          |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+--------------------------------+
|  1 | PRIMARY     | NULL  | NULL       | NULL | NULL          | NULL | NULL    | NULL | NULL |     NULL | no matching row in const table |
|  2 | SUBQUERY    | NULL  | NULL       | NULL | NULL          | NULL | NULL    | NULL | NULL |     NULL | no matching row in const table |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+--------------------------------+
2 rows in set (0.00 sec)
```

yongwen (approved):

PR #1622, opened 2024-05-21
update Container Status to Health Check Status since some services are not containerized

liyaqin1:


liyaqin1:
The current UI statement is not correct since some service is not containerized:
Screenshot 2024-05-21 at 4 02 06 PM

gzpcho (approved):

PR #1625, opened 2024-05-23
Improve layout of host panel in env deploy progress page

gzpcho:
## Summary

Environments with thousands of hosts can make of the elements in the host panel hard to see or use (having to scroll past thousands of hosts). This PR will address this issue by reorganizing the layout so that the smaller sections like "New Hosts" live in a collapsible above/before the long host list.

## Test Plan

<details>
<summary>before</summary>

image

</details>

<details>
<summary>after</summary>

image

image

</details>


deploy-board/deploy_board/templates/environs/env_landing.html
```diff
@@ -513,7 +513,7 @@ <h4 class="panel-title pull-left pointer-cursor">
         </script>
         <div class="collapse in panel-body">
             <div id="activeDeployId">
-                {% include "deploys/deploy_progress.tmpl" %}
+                {% include "deploys/deploy_progress.html" %}
```
tylerwowen, line 516: Why the rename? It's not really html.

gzpcho: Found on the Django docs:

> A template is a text file. It can generate any text-based format (HTML, XML, CSV, etc.). A template contains variables, which get replaced with values when the template is evaluated, and tags, which control the logic of the template.

https://docs.djangoproject.com/en/5.0/ref/templates/language/#templates

gzpcho: tmpl files (Django extension installed):

image

html files:

image

tylerwowen: SGTM

tylerwowen (approved):

PR #1630, opened 2024-05-28
Update UI for user and group management

tylerwowen:
For Pinterest devs, we show a message for the moved functionality to reduce confusions. Script token UI stays intact.

Screenshot 2024-05-28 at 4 27 13 PM

### DIFF
--- a/deploy-board/deploy_board/templates/users/users_moved_message.html
+++ b/deploy-board/deploy_board/templates/users/users_moved_message.html
@@ -1,0 +1,14 @@
+<div class="panel-heading clearfix">
+    <h4 class="panel-title pull-left">
+        User and group management have moved
+    </h4>
+</div>
+
+<div class="panel-body">
+    <p>
+        User and group management have moved to Pastis based ACLs. Please follow the <a href="https://pinch/teletraan-pastis-acl">instructions</a> to manage users and groups.
+    </p>
+    <p>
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/templates/users/users_moved_message.html
```diff
@@ -0,0 +1,14 @@
+<div class="panel-heading clearfix">
+    <h4 class="panel-title pull-left">
+        User and group management have moved
+    </h4>
+</div>
+
+<div class="panel-body">
+    <p>
+        User and group management have moved to Pastis based ACLs. Please follow the <a href="https://pinch/teletraan-pastis-acl">instructions</a> to manage users and groups.
```
osoriano, line 9: I thought there was a process of adding hardcoded urls to [`settings.py`](https://github.com/pinterest/teletraan/blob/master/deploy-board/deploy_board/settings.py#L305), but I see other urls like this in the codebase too so LGTM

osoriano (approved):

deploy-board/deploy_board/templates/users/users_moved_message.html
tylerwowen: This is Pinterest only and doesn't disclose the real URL, and we can change the target in pinch, so using a config from settings.py doesn't have extra benefits.

PR #1632, opened 2024-05-29
Fix `quickstart.py` local script

gzpcho:
This script is used to publish a build to a local teletraan environment.

With Teletraan local environment up:
```
cd deploy-sentinel/
make
```


osoriano (approved):
:+1:

y2587wan (approved):

PR #1634, opened 2024-05-30
U24 support for deploy-agent

osoriano:
Add support for Ubuntu 24 to deploy-agent. This includes adding
support for Python 3.12.

This requires updating some packages (e.g. `requests`, `gevent`, `boto`). For
`boto`, an `s3_client` class was added to support new (`boto3`) and old
(`boto`) packages.

Testing done:
- Unit tests passed
- Installed the deploy-agent package on an Ubuntu 24 host
  - Deploy agent ran, but failed due to missing `facter` command
- Installed the deploy-agent package on an Ubuntu 20 host
  - The deploy agent worked as expected. Was able to install new
    Teletraan builds and restart a Teletraan stage


gzpcho (dismissed):
👍

.github/workflows/python.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
```
yongwen, line 26: @gzpcho do we still have 3.7 use case?

deploy-agent/tox.ini
```diff
@@ -14,4 +14,4 @@ commands=py.test --cov=deployd {posargs}
 python =
     3.6: py36
```
yongwen, line 15: earlier, we have py37, py38? should they be consistent here

.github/workflows/python.yml
osoriano: Updated to `3.6` for consistency. Also specified the Ubuntu version since latest didn't support 3.6

deploy-agent/tox.ini
osoriano: Updated the versions in the GitHub workflow for consistency

gzpcho:
Should also update the bazel requirements:
https://github.com/pinterest/teletraan/blob/master/deploy-agent/thirdparty/requirements.txt#L1-L11

osoriano:
See https://github.com/pinterest/teletraan/pull/1653 instead

I went with a different approach supporting python 3.8 and 3.12 which is more consistent with other components

PR #1636, opened 2024-05-30
Enforce pre-commit hooks on pull requests

gzpcho:
The pre-commit action will only run on files changed by the PR.

failure: https://github.com/pinterest/teletraan/actions/runs/9309522517/job/25625109192
success: https://github.com/pinterest/teletraan/actions/runs/9309531761/job/25625137907


osoriano (approved):
LGTM, there are not too many lint errors right now, so I could follow up with a PR to address it

PR #1637, opened 2024-05-31
Bug fix: For a javascript ajax call, set the return code instead of default 200 in error view exception handler

liyaqin1:
Customer reported a bug: https://jira.pinadmin.com/browse/CDP-8039

### DIFF
--- a/deploy-board/deploy_board/webapp/error_views.py
+++ b/deploy-board/deploy_board/webapp/error_views.py
@@ -41,10 +41,12 @@ class ExceptionHandlerMiddleware:
         if request.is_ajax():
             ajax_vars = {'success': False, 'error': str(exception)}
             ret = 500
-            if isinstance(exception, NotAuthorizedException):
-                ret = 403
-            elif isinstance(exception, FailedAuthenticationException):
+            if isinstance(exception, FailedAuthenticationException):
                 ret = 401
+            elif isinstance(exception, NotAuthorizedException):
+                ret = 403
+            elif isinstance(exception, NotFoundException):
+                ret = 404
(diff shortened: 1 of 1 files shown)


tylerwowen (dismissed):

deploy-board/deploy_board/webapp/error_views.py
```diff
@@ -40,7 +40,12 @@ def process_exception(self, request, exception):
         # Error is displayed as a fragment over related feature area
         if request.is_ajax():
             ajax_vars = {'success': False, 'error': str(exception)}
-            return HttpResponse(json.dumps(ajax_vars), content_type='application/javascript')
+            ret = 500
+            if isinstance(exception, NotAuthorizedException):
+                ret = 403
+            elif isinstance(exception, FailedAuthenticationException):
+                ret = 401
+            return HttpResponse(json.dumps(ajax_vars), status=ret, content_type='application/javascript')
```
tylerwowen, line 48: It'd be nice if we can handle more exception types defined in deploy-board/deploy_board/webapp/helpers/exceptions.py

liyaqin1: there are only 3 exceptions in this file. So, I can add another 404 case.

gzpcho:
Would be good to see if we can address the TODO on this fie:

https://github.com/pinterest/teletraan/blob/d9675f8646eb42dda767ddcc1049a7dc226cb2b4/deploy-board/deploy_board/webapp/error_views.py#L26-L29

tylerwowen (approved):

PR #1641, opened 2024-06-03
 added host protection status 

Cjpilbdev:


osoriano:
The lint error should be fixed. Can you rebase and retry?

PR #1643, opened 2024-06-03
Address deploy-board lint errors

osoriano:
`ruff` was recently introduced to the precommit checks.

This PR addresses the existing checks for deploy-board:

* Run `ruff check --fix`
* Run `ruff check --fix --unsafe-fixes`
  * Verified the fixes were safe
* Use explicit exceptions
* Remove unused variables and methods
* Remove trailing ws


gzpcho (commented):
are there any lint issues on the html files? any we should fix here or exclude from checking? otherwise, the pre-commit check will fail on those

gzpcho (approved):

PR #1651, opened 2024-06-24
Update Teletraan auth schema

tylerwowen:
Enum updsates

### DIFF
--- a/docs/auth/teletraan_auth.yaml
+++ b/docs/auth/teletraan_auth.yaml
@@ -78,13 +78,14 @@ definitions:
           - GROUP
           - SYSTEM
           - PLACEMENTS
-          - BASE_IMAGES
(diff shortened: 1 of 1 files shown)


docs/auth/teletraan_auth.yaml
```diff
@@ -78,13 +78,14 @@ definitions:
           - GROUP
           - SYSTEM
           - PLACEMENTS
```
osoriano, line 80: Where is `PLACEMENTS` used? I wasn't able to find it from a quick search

osoriano (approved):

docs/auth/teletraan_auth.yaml
tylerwowen: It's not used at the moment, and we might use it in the future. This is a list of allowed values.

PR #1652, opened 2024-06-25
added info tooltip for pinfo_environment

Cjpilbdev:
image
image


rfleur01 (changes requested):
Is it possible to move the tooltip to the right of the text box so that it is in line with the text box?

Also, lets update the language to be something like: "In Teletraan production, please use set the pinfo_environment to "prod" for cmp_base clusters"

Cjpilbdev:
@rfleur01 
image

image

rfleur01 (approved):

PR #1666, opened 2024-07-15
Remove origin prefix from branch name before publishing to EB

knguyen100000010:
Some legacy CI jobs still add prefix "origin/" to branch name. Remove this prefix before publishing.

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/events/EventBridgePublisher.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/events/EventBridgePublisher.java
@@ -29,12 +29,13 @@ public class EventBridgePublisher implements BuildEventPublisher {
 
   @Override
   public void publish(BuildBean buildBean, String action) {
+    final String originalBranch = buildBean.getScm_branch();
 
     // Some legacy CI jobs still use remote-tracking branch (with prefix "origin/" added to branch name).
     // Remove this prefix before publishing.
-    if (StringUtils.startsWithIgnoreCase(buildBean.getScm_branch(), ORIGIN_PREFIX) && !StringUtils.equalsIgnoreCase(buildBean.getScm_branch(), ORIGIN_PREFIX)) {
-      final String correctedBranch = buildBean.getScm_branch().substring(ORIGIN_PREFIX.length());
-      buildBean.setScm_branch(correctedBranch);
+    if (StringUtils.startsWithIgnoreCase(originalBranch, ORIGIN_PREFIX) && !StringUtils.equalsIgnoreCase(originalBranch, ORIGIN_PREFIX)) {
+      final String localBranch = buildBean.getScm_branch().substring(ORIGIN_PREFIX.length());
+      buildBean.setScm_branch(localBranch);
     }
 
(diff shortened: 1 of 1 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/events/EventBridgePublisher.java
```diff
@@ -29,11 +29,12 @@ public EventBridgePublisher(EventBridgeAsyncClient eventBridgeAsyncClient, Strin
...
   public void publish(BuildBean buildBean, String action) {
 
-    // The branch name is actually "master" in this case. Some legacy CI jobs still use "origin/master",
-    // which is a nonexistent branch. Make this correction so consumers of this event can process
-    // branch info properly.
-    if (StringUtils.equalsIgnoreCase("origin/master", buildBean.getScm_branch())) {
-      buildBean.setScm_branch("master");
+    // Some legacy CI jobs still use remote-tracking branch (with prefix "origin/" added to branch name).
+    // Remove this prefix before publishing.
+    final String originPrefix = "origin/";
```
mitchellngoodman, line 34: Seems like a constant that could be pulled out of the method

deploy-service/common/src/main/java/com/pinterest/deployservice/events/EventBridgePublisher.java
```diff
@@ -29,11 +29,12 @@ public EventBridgePublisher(EventBridgeAsyncClient eventBridgeAsyncClient, Strin
...
+    if (StringUtils.startsWithIgnoreCase(buildBean.getScm_branch(), originPrefix) && !StringUtils.equalsIgnoreCase(buildBean.getScm_branch(), originPrefix)) {
+      final String correctedBranch = buildBean.getScm_branch().substring(originPrefix.length());
+      buildBean.setScm_branch(correctedBranch);
```
mitchellngoodman, line 37: Will mutating the input buildBean object have any other effects? Like where is it used elsewhere after this function and will that change anything (like [here](https://github.com/pinterest/teletraan/blob/master/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/Builds.java#L241))

Some might say input to methods should not be mutated.

rfleur01 (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/events/EventBridgePublisher.java
```diff
@@ -53,6 +55,9 @@ public void publish(BuildBean buildBean, String action) {
     } catch (Exception e) {
       logger.error("Failed to publish event to Event Bridge: {}", entry, e);
     }
+
+    // set branch name back to its original value in case it's expected.
+    buildBean.setScm_branch(originalBranch);
```
mitchellngoodman, line 60: Best practice would generally be to make a copy instead of modifying the input object. In the case that multiple threads are referencing the object there could be a race condition. And even if it doesn't matter today maybe it could in the future.

Not blocking though.

mitchellngoodman (approved):

PR #1676, opened 2024-07-26
Change a host termination error to forbidden

tylerwowen:
The actual change is just `NotAllowedException` -> `ForbiddenException`. Forbidden (403) makes more sense than the "method not allowed" error.


osoriano (approved):
:+1:

PR #1677, opened 2024-07-27
make sure there is running hosts when marking deploy state as succeeding

liyaqin1:
Don't set the deployment state as succeeding if there is no running hosts.


liyaqin1:
Thanks Tyler for diligently investigating the root cause of the issue reported by the customer on the Teletraan Platform. Here is the analysis of the root cause: https://pinterest.slack.com/archives/CBLQ8DR09/p1722549044166659?thread_ts=1721239921.238729&cid=CBLQ8DR09

PR #1684, opened 2024-08-03
Format all java code

tylerwowen:
Ran. `mvn spotless:apply -DspotlessSetLicenseHeaderYearsFromGitHistory=true`.
Inspected all changed files and made adjustment to some doc strings. 

Note that the SQL table in some of the DB beans were all removed, because they were not well maintained.


osoriano (approved):
I reviewed this by formatting the code locally. I saw some change in the comment strings but no code changes. LGTM

PR #1688, opened 2024-08-07
Add a new AuthZ resource type

tylerwowen:
This new resource type represents the config `provision_prefix` in the cluster configuration. It's similar to IAM_ROLE. Clusters with this configuration will have hostnames with this configuration as the prefix.


osoriano (commented):
Could you add a description for the use case for the auth resource type?

osoriano (approved):

PR #1689, opened 2024-08-07
Fix broken link

hobinyoon:


osoriano (approved):
:+1:

hobinyoon:
Seems like the file had an extra line at the end and broke pre-commit. Will follow up with another PR quickly

PR #1693, opened 2024-08-12
More consistent calls for updating capacity

osoriano:
There are two calls that can be used to add capacity. One uses [`POST`](https://github.com/pinterest/teletraan/blob/master/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvCapacities.java#L145). The other is a [`PUT`](https://github.com/pinterest/teletraan/blob/master/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvCapacities.java#L101)

Update the auth for these two calls to be more consistent

# Testing
* Added unit tests
* Will deploy to staging

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvCapacities.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvCapacities.java
@@ -15,6 +15,7 @@
  */
 package com.pinterest.teletraan.resource;
 
+import com.google.common.collect.ImmutableList;
 import com.pinterest.deployservice.bean.EnvironBean;
 import com.pinterest.deployservice.bean.TeletraanPrincipalRole;
 import com.pinterest.deployservice.common.Constants;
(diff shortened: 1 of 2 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvCapacities.java
```diff
@@ -157,9 +158,12 @@ public void add(
             @NotEmpty String name,
             @Context SecurityContext sc)
             throws Exception {
+        name = name.replace("\"", "");
```
tylerwowen, line 161: Should we do this to all endpoints? Or we might have inconsistent behaviors.

tylerwowen: NVM

tylerwowen (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvCapacities.java
osoriano: This line was just moved from below, so kept for backward compatibility

PR #1696, opened 2024-08-14
Emit launch latency

tylerwowen:
## TAS
Previously launch latency was emitted by a Rodimus worker. This PR implements the same measure in the TAS. 

Launch latency is defined as the duration from host launch to the the complete of the first deployment. It measures for all environments deployed on new hosts. The configuration **Launch grace period** is the user set threshold for this latency and it's used for AgentJanitor and Rodimus health check.

Another fix is that I found all first deploy metrics were success. So I updated the condition when the `first_deploy` flag should be turned off. 

## UI
While updating the group status page to include the launch latency, I realized I forgot that the original plan was to create a dashboard. So I removed the links added in #1694, created a dashboard including the changes in a new Teletraan user dashboard.

Fixed launch failure rate graph and updated some metrics calculations.

image

### TAS
1. Deploy this PR to TAS dev1
2. Launch a new host in tyler/test
    - [x] Launch latency should be emitted 
    - [x] First deploy counter should increment with success=true.
4. Deploy a bad build to tyler/test
5. Launch a new host
    - [x] Launch latency should be emitted
    - [x] First deploy counter should increment with success=false.

### UI
1. Deploy this PR to deploy-board dev1
2. Visit group status page /groups/tyler-test/
    - [x] verify the link is updated and working

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
@@ -492,7 +492,7 @@ public class GoalAnalyst {
                 return hosts.get(0).getCreate_date();
             }
         } catch (Exception ex) {
-            LOG.info("Failed to get host with id {}", agent.getHost_id(), ex);
+            LOG.warn("Failed to get host with id {}", agent.getHost_id(), ex);
         }
 
         return null;
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
@@ -435,7 +435,9 @@ public class GoalAnalyst {
             updateBean.setStart_date(agent.getStart_date());
         }
 
-        if (report.getDeployStage() == DeployStage.SERVING_BUILD && updateBean.getFirst_deploy()) {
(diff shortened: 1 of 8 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
```diff
@@ -457,13 +466,38 @@ private void emitMetrics(AgentBean updateBean) {
...
     }
 
+    private Long estimateHostStartTime(AgentBean agent) {
+        try {
+            List<HostBean> hosts = hostDAO.getHostsByHostId(agent.getHost_id());
+            if (!hosts.isEmpty()) {
+                return hosts.get(0).getCreate_date();
+            }
+        } catch (Exception ex) {
+            LOG.info("Failed to get host with id {}", agent.getHost_id(), ex);
```
osoriano, line 495: Could be a warn similar to [previous change](https://github.com/pinterest/teletraan/blob/master/deploy-service/common/src/main/java/com/pinterest/deployservice/handler/PingHandler.java#L242) so that the level matches, but may not matter much since it seems unexpected

osoriano (approved):

deploy-service/common/src/main/java/com/pinterest/deployservice/handler/GoalAnalyst.java
tylerwowen: Good catch.

osoriano (approved):

PR #1707, opened 2024-09-03
added stage-type badges to teletraan environments page

bhnord:
canary: 
* canary badge

control:
* control badge

dev:
* dev badge

latest:
* latest badge

production:
* production badge

staging: 
* staging badge

### DIFF
--- a/deploy-board/deploy_board/templates/environs/env_landing.html
+++ b/deploy-board/deploy_board/templates/environs/env_landing.html
@@ -372,6 +372,9 @@ $('#privateBldUploadBtnId button').click(function () {
                 </div>
                 {% endif %}
             {% endif %}
+            <h6 class="stage-type-badge">
+                {{env.stageType|default:"NO STAGE TYPE"}}
+            </h6>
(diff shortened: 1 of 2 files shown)


deploy-board/deploy_board/templates/environs/env_landing.html
```diff
@@ -372,6 +372,9 @@ <h4 class="panel-title pull-left pointer-cursor">
                 </div>
                 {% endif %}
             {% endif %}
+            <h6 class="stage-type-badge">
+                {{env.stageType|default:"NO STAGE TYPE"}}
```
bhnord, line 376: added NO STAGE TYPE as stageType is an optional field from the API according to the docs

osoriano: I'm thinking we could show a warning if the stage type is missing, to encourage users to set it. The warning could be similar to the one in the screenshot.

However, I think we could discuss this separately, so this change looks good to me :+1:

osoriano (approved):

PR #1709, opened 2024-09-04
 Hide 'Backup host types in order: None' text if the feature is disabled

liyaqin1:
Reasoning: Not many users will have this enabled

### DIFF
--- a/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
+++ b/deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
@@ -579,7 +579,7 @@ Vue.component('backup-hosttypes', {
     template: '<div class="form-group" v-if="isVisible">\
     <div class="col-xs-2"></div>\
     <div class="col-xs-8">\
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/static/js/components/clusterconfigcomponents.js
```diff
@@ -576,13 +576,18 @@ Vue.component('remaining-capacity', {
...
     <div class="col-xs-2"></div>\
     <div class="col-xs-8">\
         <span class="col-xs-8" style="padding:0;">Backup Host Types In Order: {{backuphosttypes}}</span>\
     </div>\
     </div>',
-    props: ['backuphosttypes']
+    props: ['backuphosttypes'],
+    computed: {
+        isVisible: function () {
+            return this.backuphosttypes === 'None' ? false : true
```
osoriano, line 588: Seems this could be simplified
```suggestion
            return this.backuphosttypes !== 'None'
```

liyaqin1: sounds good, thank you

PR #1727, opened 2024-10-04
toggle private build when stage type changes

rwxzhu:
toggle off private builds once the stage type is changed from dev

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
@@ -132,6 +132,8 @@ public class EnvStages {
 
         if (environBean.getStage_type() == EnvType.DEV) {
             environBean.setAllow_private_build(true);
+        } else if (origBean.getStage_type() == EnvType.DEV) {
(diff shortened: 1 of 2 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvStages.java
```diff
@@ -132,6 +132,8 @@ public void update(
 
         if (environBean.getStage_type() == EnvType.DEV) {
             environBean.setAllow_private_build(true);
+        } else {
```
tylerwowen, line 135: What if stage type isn't changed but still provided in the request?

rwxzhu: Like DEV - > DEV or would it be NON-DEV -> NON-DEV?

If it's the 2nd one I have updated it to only update for changes where it goes from DEV -> NON-DEV.

tylerwowen: NON-DEV -> NON-DEV

tylerwowen (approved):

PR #1731, opened 2024-10-16
Set `isManagedResource` to false on any cluster updates

gzpcho:
### `isManagedResource` is persisted as `false` when updating cluster with backend change

https://www.loom.com/share/c89443b35aae4c36bb0aef70831b810b

### `isManagedResource` is persisted as `false` when creating cluster with backend change

https://www.loom.com/share/bfa3315843bb40f3894f94a3e00f8e81

### `isManagedResource` is persisted as `false` when updating cluster capacity with backend change

https://www.loom.com/share/bd240f05641f4198ac17f5d4be638a94?sid=3e69f69e-9698-49ba-a02e-605c3b712503

### Cluster update still works without backend change

https://www.loom.com/share/a6158ed5feb44c8c8e077744ee33209f?sid=95f4700f-14a2-48cf-9c70-045b710a13da

### Cluster creation still works without backend change

https://www.loom.com/share/18ca52987d754dff889d105187fd3d67

### Cluster capacity update still works without backend change

https://www.loom.com/share/74d9ef456e2f403ca5622757d0b70401?sid=0d631b28-8a9f-4e35-8f22-861d454bf53e

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -1006,7 +1006,6 @@
                            delete clusterInfo['configs']['cmp_group']
                        }
                     }
-                    clusterInfo['isManagedResource'] = false
 
                     if (this.validateInput(clusterInfo, this.imageNameValue)) {
                         //Send request
--- a/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
+++ b/deploy-board/deploy_board/templates/clusters/cluster_configuration.html
@@ -1006,6 +1006,7 @@
                            delete clusterInfo['configs']['cmp_group']
                        }
                     }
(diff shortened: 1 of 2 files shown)


robbintt:
For the cases:
* isManagedResource is persisted as false when updating cluster with backend change
* isManagedResource is persisted as false when updating cluster capacity with backend change

Should it not just pass the previous value when updating?

I know that this can only be updated if it is false, but is it a bit easier to reason about if it only sets it on "create" and otherwise persists whatever value previously existed?

I am not 100% certain on the best logic, but just thinking about how to reason through the system.

The actual set to "false" for the 2 cases update / update cluster would come from the form contents rather than hardcoded into the update function?

Is there an additional reason to hardcode "false" in update? Thanks!

robbintt (approved):

PR #1732, opened 2024-10-17
Add checkbox in cluster forms for when the cluster is managed

gzpcho:
image

image

image

image

image

image

### DIFF
--- a/deploy-board/deploy_board/templates/clusters/cluster_capacity.tmpl
+++ b/deploy-board/deploy_board/templates/clusters/cluster_capacity.tmpl
@@ -61,8 +61,8 @@
         const banner = document.getElementById('resourceManagedBanner');
 
         checkbox.addEventListener('change', function() {
-            fieldset.disabled = checkbox.checked;
-            submitBtn.disabled = checkbox.checked;
+            fieldset.disabled = !!checkbox.checked;
+            submitBtn.disabled = !!checkbox.checked;
             banner.style.display = checkbox.checked ? 'none' : 'block';
         });
     });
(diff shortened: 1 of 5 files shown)


robbintt:
Could the checkbox go into a div appropriate with for the rest of the form contents, and maybe go in a vertical form width 
 stack above the submit button so it is much more prominent?

robbintt:
Also I think the checkbox should have a description like "Disable Mangement", then the long language would be a tooltip or a italicized small font description on the following line?

robbintt:
I think it would better not to ship the form partially disabled --

What do you think about:
1. ask the team if there is a straightforwards vue way to disable
- (we can sacrifice the active approach to allowing changes to the form, e.g. the checkbox would need to be submitted via the button, then the form reloads and it is editable.)
2. hide any fields that can't be disabled, so they are invisible, then solve the hiding as a followup after the project is ready.

gzpcho:
> Could the checkbox go into a div appropriate with for the rest of the form contents, and maybe go in a vertical form width stack above the submit button so it is much more prominent?

> Also I think the checkbox should have a description like "Disable Mangement", then the long language would be a tooltip or a italicized small font description on the following line?

Addressed on [368d6ec](https://github.com/pinterest/teletraan/pull/1732/commits/368d6ec7a9768df0c7333dfb17cdbc68e112730e)

gzpcho:
Was able to completely disable the form on [c84bb33](https://github.com/pinterest/teletraan/pull/1732/commits/c84bb332c6129413b2f8334b52a7956961ce833e)

robbintt (approved):

PR #1736, opened 2024-10-30
Return 404 if no current deploy exists

tylerwowen:
The current implementation let `getDeploySafely` handle `deployId` being null and throws an exception that will eventually be mapped to 500. To improve the usability of this API, explicitly throw 404 if `deployId` is blank.


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/resource/EnvDeploys.java
```diff
@@ -95,6 +96,10 @@ public DeployBean get(
                     String stageName)
             throws Exception {
         EnvironBean envBean = Utils.getEnvStage(environDAO, envName, stageName);
+        if (StringUtils.isNotBlank(envBean.getDeploy_id())) {
```
liyaqin1, line 99: Shouldn't it be isBlank instead of isNotBlank?

liyaqin1 (approved):

PR #1741, opened 2024-11-06
Refactor Http Client implementation 

tylerwowen:
Refactor the implementation using java.net with OkHttp.

The current HTTP client has various issues. Instead of pathing the existing implementations, it's better to use another client library. OkHTTPClient is one of the most popular one and it's used by PinDeploy and Spinnaker. 

## Improvements
1. Observability for outgoing requests.
2. 1 shared connection pool managed by OkHttp instead of many ad-hoc connections.
3. Exponential backup retry on certain failures.
4. Logging with configurable level and sensitive header redaction.
4. Moved to universal so Rodimus can use the same implementation

## Tests and validations
Thanks to the added observability, it's very easy to know and monitor if this refactoring is working in addition to the unit tests.

### Smoke tests
Deploy both service and worker to dev1 and ensure these are working
- Github integration - commit data fetch
- Rodimus integration -  Ec2 tags can load
- Changefeed - events are published to changefeed
- Webhook - webhooks can be invoked

#### Basic
Screenshot 2024-11-07 at 15 34 48

#### Body and header
Screenshot 2024-11-07 at 15 35 46

### Metrics
Screenshot 2024-11-07 at 15 31 55
Screenshot 2024-11-07 at 15 31 42


deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/http/RetryInterceptor.java
```diff
@@ -0,0 +1,69 @@
...
+                if (response.isSuccessful() || !shouldRetry(response)) {
+                    return response;
+                }
+            } catch (IOException e) {
+                lastException = e;
+            }
+
+            try {
+                long backoff = (long) Math.pow(2, i) * retryInterval;
+                TimeUnit.MILLISECONDS.sleep(backoff);
```
osoriano, line 51: Seems the `sleep` can be avoided in the last iteration. It may be a rare case though

deploy-service/common/src/main/java/com/pinterest/deployservice/common/Jenkins.java
```diff
@@ -26,8 +27,7 @@
 /** Wrapper for Jenkins API calls */
 public class Jenkins {
     private static final Logger LOG = LoggerFactory.getLogger(Jenkins.class);
-    private static final int RETRIES = 3;
-    private HTTPClient httpClient = new HTTPClient();
+    private static final HttpClient httpClient = HttpClient.builder().maxRetries(3).build();
```
osoriano, line 30: Could clean up the `.maxRetries(3)` since that is the same as the default

osoriano (approved):

osoriano (approved):

PR #1743, opened 2024-11-06
CDP-8544: Added an endpoint to get active host ids

vitalii-honchar:
# Context

Rodimus should have a possibility to fallback to check running hosts in the Teletraan service. This PR adds a new endpoint which returns active (running) hosts from the Teletraan database.

# Tests

## Test Script
```Bash
#!/bin/bash

TOKEN=''

curl -i -X POST --location 'https://<teletraan_url>/v1/hosts/active' \
    --header 'Content-Type: application/json' \
    --header "Authorization: token $TOKEN" \
    --data '["i-00ed0c81d932f4038", "i-045152ef8256fcb18"]'
```

### Active Hosts

```
["i-00ed0c81d932f4038","i-045152ef8256fcb18"]
```

### Not Active Hosts

```
[]
```

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
@@ -28,6 +28,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.commons.dbcp.BasicDataSource;
 import org.apache.commons.dbutils.QueryRunner;
 import org.apache.commons.dbutils.ResultSetHandler;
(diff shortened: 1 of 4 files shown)


deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
```diff
@@ -55,6 +56,10 @@ public class DBHostDAOImpl implements HostDAO {
             "SELECT COUNT(host_id) FROM hosts WHERE group_name=?";
     private static final String GET_ALL_HOSTS_BY_GROUP =
             "SELECT * FROM hosts WHERE group_name=? AND state!='TERMINATING'";
+
+    private static final String GET_ACTIVE_HOST_IDS_BY_HOST_IDS =
+            "SELECT DISTINCT(host_id) FROM hosts WHERE host_id IN (%s) AND state NOT IN ('TERMINATING', 'TERMINATED')";
```
tylerwowen, line 61: I think this should be just `ACTIVE`.

tylerwowen (commented):
LGTM
Please run `mvn spotless:apply`

deploy-service/common/src/main/java/com/pinterest/deployservice/db/DBHostDAOImpl.java
vitalii-honchar: Thanks, changed.

tylerwowen (approved):

PR #1746, opened 2024-11-08
Improve job pool configurations

tylerwowen:
Due to misconfiguration, the job pool used in Teletraan has only one thread. Similar issue fixed in https://github.com/pinternal/rodimus/pull/110

1. `ExecutorService` used for thread pool has only 1 thread. Changed the work queue type to fix. The configuration is the same as the `regular` pool in the Rodimus. 
2. `ExecutorService` instance is not managed, so it's not properly shutdown. Use Dropwizard for lifecycle management.
3.  Bind it to Micrometer's registry so we get metrics. 

## Considerations 
I didn't have the whole change in Rodimus ported because it was not needed. There are only 3 types of tasks involved in Teletraan, ChangeFeed, NotifyJob and Webhook. They will work fine with just the regular pool configuration.

## Test plan
This change is a subset of changes we made to Rodimus and it's been working well. For sanity tests, I deployed a build with this change and verified that change feed events are published and metrics are emitted.

Screenshot 2024-11-08 at 14 39 52
Screenshot 2024-11-08 at 14 39 43

### DIFF
--- a/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
+++ b/deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
@@ -64,16 +64,22 @@ import com.pinterest.teletraan.worker.HotfixStateTransitioner;
 import com.pinterest.teletraan.worker.MetricsEmitter;
 import com.pinterest.teletraan.worker.SimpleAgentJanitor;
 import com.pinterest.teletraan.worker.StateTransitioner;
+
+import io.dropwizard.setup.Environment;
+import io.dropwizard.util.Duration;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
 import java.util.TreeMap;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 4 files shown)


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
```diff
@@ -218,20 +222,20 @@ public static TeletraanServiceContext setupContext(TeletraanServiceConfiguration
...
+                        .lifecycle()
+                        .executorService(jobPoolName)
+                        .minThreads(poolSize * 4)
+                        .maxThreads(poolSize * 8)
+                        .keepAliveTime(Duration.seconds(30))
+                        .workQueue(new ArrayBlockingQueue<>(poolSize * 5000, false))
+                        .shutdownTime(Duration.seconds(30))
+                        .rejectedExecutionHandler(new AbortPolicy())
+                        .build();
+        new ExecutorServiceMetrics(jobPool, jobPoolName, null).bindTo(Metrics.globalRegistry);
```
osoriano, line 238: I see the linked PR uses `InstrumentedExecutorService` instead. Would be nice to be consistent, but we could also update in the future if needed

osoriano (approved):

deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/ConfigHelper.java
tylerwowen: Good catch. This is intentional because `InstrumentedExecutorService` is part of dropwizard and metrics are emitted via dropwizard-metrics, which is being replaced by micrometer.

PR #1760, opened 2024-12-04
CDP-8328: deploy-service: add normandie and knox statuses to PingRequest and hosts_and_agents table

Tweedeldee:
# Description
Add normandie status and knox status to ping requests 
[CDP-8328](https://jira.pinadmin.com/browse/CDP-8328)

# Development status
* Normandie status is defined by the presence of a valid normandie's certificate.
* Knox status is defined by knox deamon running successfully 
* Both fields are emitted in ping requests and stored in `hosts_and_agents` SQL table

## Associated PRs
* deploy-agent changes: https://github.com/pinterest/teletraan/pull/1759
* deploy-service: output normandie and knox status in getHosts https://github.com/pinterest/teletraan/pull/1763

## Missing before pushing:
* Database changes in Prod

## Previous version of this PR
https://github.com/pinterest/teletraan/pull/1745

PR #1763, opened 2024-12-06
CDP-8328: Add normandie status and knox statuses to Hosts.get() EnvHosts.getHostByHostName()

Tweedeldee:
# Description
In deploy-board, Normandie and Knox statuses are to be added to `hosts/host_details.html`.
`hosts/host_details.html` is rendered in two places, using a list of hosts coming from two different methods

`hosts = hosts_helper.get_hosts_by_name(request, hostname)` :https://github.com/pinterest/teletraan/blob/b0f6c6aab3084b4d7b7f4448f2fac210df10f96f/deploy-board/deploy_board/webapp/host_views.py#L150

`hosts = environ_hosts_helper.get_host_by_env_and_hostname(request, name, stage, hostname)` https://github.com/pinterest/teletraan/blob/b0f6c6aab3084b4d7b7f4448f2fac210df10f96f/deploy-board/deploy_board/webapp/host_views.py#L203

This means that both `get_hosts_by_name` and `get_host_by_env_and_hostname` needs to return normandie and knox statuses.

This is done by:
* creating a new POJO class, HostBeanWithStatuses, which inherits HostBean and include the two new fields
* update the relevant DAO methods to return HostBeanWithStatuses
* update the associated SQL queries to retrieve the two fields from `hosts_and_agents` table

## Associated PRs
* deploy-service write changes: https://github.com/pinterest/teletraan/pull/1760

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/HostBean.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/bean/HostBean.java
@@ -15,14 +15,16 @@
  */
 package com.pinterest.deployservice.bean;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
 import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
 
-@Builder
+@SuperBuilder
 @NoArgsConstructor
 @AllArgsConstructor
 @Data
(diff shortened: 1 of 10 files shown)


Tweedeldee:
## DEV1 proof of work:
https://deploy-dev1.pinadmin.com/env/helloworlddummyservice/dev/host/engprod-teletraan-service-dev1-0ab9009c
image

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/HostBean.java
```diff
@@ -153,9 +89,4 @@ public SetClause genSetClause() {
                     + "last_update=VALUES(last_update),"
                     + "state=VALUES(state),"
                     + "can_retire=VALUES(can_retire)";
-
-    @Override
```
Tweedeldee, line 157: As discussed offline, this toString() might be used for non-trivial logic, and it's quite hard to identify all toString() usages since the method can be implicitly called. 
Reverting the change to keep the existing behavior.

deploy-service/common/src/main/java/com/pinterest/deployservice/bean/HostBeanWithStatuses.java
```diff
@@ -0,0 +1,69 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
+public class HostBeanWithStatuses {
+    @JsonProperty("hostName")
+    private String host_name;
```
tylerwowen, line 30: since this is external facing, it can just be
```java
private String hostName;
```
The private field doesn't have to match the DB column name.

tylerwowen (approved):

PR #1764, opened 2024-12-10
Map errors to exceptions

tylerwowen:
Map WebApplicationException to more detailed subclasses, this is helpful for Rodimus adoption of this HttpClient implementation. Additionally, making `callTimeout` configurable and replacing existing `readTimeout` and `connectTimeout`. It's needed by Rodimus and it's easier to control 1 then 2. 

> The call timeout spans the entire call: resolving DNS, connecting, writing the request body, server processing, and reading the response body. If the call requires redirects or retries all must complete within one timeout period.

Lastly, there is a leak in the constructor. It's now moved to the static init block.

### DIFF
--- a/deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/http/HttpClient.java
+++ b/deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/http/HttpClient.java
@@ -53,6 +53,12 @@ public class HttpClient {
     private static final OkHttpClient sharedOkHttpClient = new OkHttpClient();
     private static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(20);
 
+    static {
+        observationRegistry
+                .observationConfig()
+                .observationHandler(new DefaultMeterObservationHandler(Metrics.globalRegistry));
+    }
+
     @Getter private final OkHttpClient okHttpClient;
 
     @Builder(buildMethodName = "buildInternal")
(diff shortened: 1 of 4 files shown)


deploy-service/universal/src/main/java/com/pinterest/teletraan/universal/http/HttpClient.java
```diff
@@ -182,13 +187,37 @@ String makeCall(Request request) throws IOException {
...
+                responseCode,
+                responseBody);
+
+        switch (responseCode) {
+            case 400:
+                throw new BadRequestException(responseBody);
+            case 401:
+                throw new ForbiddenException(responseBody);
+            case 403:
+                throw new NotAuthorizedException(responseBody);
```
osoriano, line 210: Seems the exceptions here should be flipped
* [`ForbiddenException`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/ForbiddenException.html) - 403
* [`NotAuthorizedException`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/NotAuthorizedException.html) - 401 Unauthorized

tylerwowen: Great, fixed

osoriano (approved):

osoriano (approved):

PR #1777, opened 2025-02-06
Format all java code

osoriano:
Formatted all the java code twice, since after the first format command, the history is updated and it needs another format.
Verified that running the formatter again does not produce an additional changes.

Ran:
```
mvn spotless:apply -DspotlessSetLicenseHeaderYearsFromGitHistory=true
git commit -am "Format all java code"
mvn spotless:apply -DspotlessSetLicenseHeaderYearsFromGitHistory=true
git commit -am "Format all java code, again"
```


osoriano:
I was able to work around this by removing the DspotlessSetLicenseHeaderYearsFromGitHistory from the formatter command

PR #1799, opened 2025-04-03
Smaller Teletraan tar ball

tylerwowen:
Since migrating to docker based, it's no longer needed to include any jars in the tar ball.

Also, we only need teletraan* directories from internal teletraanconfigs.

Note: will need to update jenkins job to pick up the smaller tar ball. The original is still needed for docker image build.


osoriano (approved):
Nice improvement

PR #1806, opened 2025-04-15
Update default arm64 image

osoriano:
The current default arm64 image sent in the UI does not match what arm users typically use. Update this default

### DIFF
--- a/deploy-board/deploy_board/settings.py
+++ b/deploy-board/deploy_board/settings.py
@@ -394,7 +394,7 @@ if IS_PINTEREST:
 
     # Pinterest Default AMI image name
     DEFAULT_CMP_IMAGE = "cmp_base-ebs-18.04"
-    DEFAULT_CMP_ARM_IMAGE = "cmp_base_arm64"
+    DEFAULT_CMP_ARM_IMAGE = "cmp_base_u20_arm64"
 
     # Pinterest Default setting whether to use launch template or not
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/settings.py
```diff
@@ -394,7 +394,7 @@
 
     # Pinterest Default AMI image name
     DEFAULT_CMP_IMAGE = "cmp_base-ebs-18.04"
-    DEFAULT_CMP_ARM_IMAGE = "cmp_base_arm64"
+    DEFAULT_CMP_ARM_IMAGE = "cmp_base_u20_arm64"
```
gzpcho, line 397: I wonder if it could make sense to use the Ubuntu 24 image as the default - `cmp_base_u24_arm64`

gzpcho (approved):

deploy-board/deploy_board/settings.py
osoriano: I'm hoping we could update both (amd/arm) of these real soon, but currently `cmp_base_u24_arm64` adoption is very low

osoriano: Maybe after a bit more u24 validation, we could update both the `arm` and `amd` default images here

osoriano: Going to merge since the current `cmp_base_arm64` doesn't even exist and causes an unexpected default AMI selection

PR #1808, opened 2025-04-15
Update link in groups page

osoriano:
## Summary

Previously, the link would point to the particular environment.

This can be misleading because the group can belong to a particular stage. Update the link to point to the stage

## Screenshots

**Before**

before

**After**

after


gzpcho (approved):
Thank you!

PR #1826, opened 2025-05-29
Add deploy info metrics

osoriano:
## Summary

Add deployment info metric data that could be used for ACA. These are sent at the end of the deploy-agent run

## Testing

* Added unit tests
* Ran on a local setup and verified metrics were visible
* Ran on a dev cluster, verified metrics were visible and able to be used to segment canary/non-canary metrics
* Ran a performance test on a large cluster (3k+ nodes). Ensured metrics could still be sent through the socket quickly (<1s)

## Rollout Plan

Although, there shouldn't be any performance issues due to the large amounts of requests the metrics system can handle, it will still be rolled out gradually since this affects many hosts

### DIFF
--- a/deploy-agent/deployd/agent.py
+++ b/deploy-agent/deployd/agent.py
@@ -19,6 +19,7 @@ import logging
 import os
 import sys
 from random import randrange
+import socket
 import time
 import traceback
 
@@ -47,6 +48,9 @@ from deployd.common.types import (
 from deployd import __version__, IS_PINTEREST, MAIN_LOGGER
 from deployd.types.deploy_goal import DeployGoal
 from deployd.types.ping_response import PingResponse
+
+DEPLOY_INFO_METRIC_NAME = "deploy.info"
+DEPLOY_INFO_METRIC_VALUE = 1
 
 log: logging.Logger = logging.getLogger(name=MAIN_LOGGER)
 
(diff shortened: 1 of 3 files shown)


yongwen (approved):

deploy-agent/deployd/agent.py
```diff
@@ -550,6 +559,61 @@ def _update_deploy_alias(self, deploy_goal) -> None:
...
+                log.info(
+                    f"Skip deploy info metric for {env_name} due to missing commit"
+                )
+                continue
+
+            put_stmts.append(
+                f"put {DEPLOY_INFO_METRIC_NAME} "
+                f"{epoch_in_seconds} {DEPLOY_INFO_METRIC_VALUE} "
+                f"source=teletraan "
+                f"artifact={env_name} "
```
yongwen, line 597: i think it is better to use artifact = build_name. it is more meaningful and can be used to differenciate different deploy on the same host, and to idenfiy the commit sha it belongs to.

deploy-agent/deployd/agent.py
```diff
@@ -550,6 +559,61 @@ def _update_deploy_alias(self, deploy_goal) -> None:
                 )
             )
 
+    def _send_deploy_info_metrics(self) -> None:
```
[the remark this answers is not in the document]
osoriano: The tsd service is open source. Here is the documentation for the `put` metric format: https://opentsdb.net/docs/build/html/api_telnet/put.html#request

deploy-agent/deployd/agent.py
osoriano: Thanks, updated. Going to deploy to a test cluster

yongwen (approved):

deploy-agent/deployd/agent.py
yongwen: thanks

PR #1795, opened 2025-04-01
Support Teletraan and Buildkite integration through CIPlatformManagerProxy

jasmineqin:
### Summary
[ER-12903](https://jira.pinadmin.com/browse/ER-12903)
I plan to make the changes in three different phases.

#### Phase 1
This PR. Essentially, when a Hotfix is created from Teletraan, both Buildkite and Jenkins `<repo>-hotfix-job` will be triggered, Buildkite `<repo>-hotfix-job` will cherry pick and create a branch, similar to the hotfix branch. The primary goal of this phase is to shadow run the `<repo>-hotfix-job` and make sure the PinCI cherry-pick process work as expected, since we have migrated away from the `perform-hotfix` CLI.

This PR includes changes:
- Adding Buildkite class.
- Introducing CIPlatformManagerProxy as the proxy to both Buildkite and Jenkins for triggering, querying builds etc.
- The [corresponding teletraanconfigs PR](https://github.com/pinternal/teletraanconfigs/pull/137/files#diff-bcb1cc4cfb84d81fbac12cd2258aa44f96a2ced003636453d32d4c5d8eb693bf) includes changes to turn Jenkins and Buildkite as list of CIPlatforms under `ci`.
- To maintain backward compatibility, leverage constructor overloading for the Jenkins class.
- To maintain backward compatibility, leverage `defaultImpl` inside `@JsonTypeInfo` in the CIPlatformFactory class.
- To support both CIPlatformManagerProxy and Jenkins individually, setup Jenkins and Buildkite accordingly in the HotfixStateTransitioner class
- [HotfixStateTransitioner.java](https://github.com/pinterest/teletraan/pull/1795/files#diff-2fb2a900cd1c42a9d496fe9f1cd310242963e88e3d9506032cf3ac977a2ce77f) during `HotfixState.INITIAL` state, it will trigger both Buildkite and Jenkins `<repo>-hotfix-job` if the job exists.

#### Phase 2
Also triggers Buildkite for the `<repo>-private-build` if the job exists. This will allow Buildkite to build tarball, the tarball will not be published until we are ready to switch over to Buildkite than Jenkins

#### Phase 3
For both `<repo>-hotfix-job` and `<repo>-private-build`, if they exist on Buildkite, then only Buildkite pipelines are triggered, Jenkins' will not be. This phase will last a while until all hotfix pipelines are migrated off Jenkins.

### Testing plan with a list of CI platforms in teletraanconfigs (when Buildkite pipeline exists)

[...]

PR #1827, opened 2025-05-30
Fully support Jenkins -> Buildkite migration for hotfix pipelines 

jasmineqin:
### Summary

After the [initial change in Teletraan](https://github.com/pinterest/teletraan/pull/1795) and the [TeletraanConfigs change](https://github.com/pinternal/teletraanconfigs/commit/0d407a563c794fa0e996a3d36abee133a0a80e66#diff-34a2d843812933f25ca64bbd37fa7459c79d2c445a431f27faab0d927f19492c), once this change is merged, Buildkite will be recognized as a higher priority CI platform than Jenkins when pipelines get triggered, i.e. covers the `Phase 2` and `Phase 3` that are mentioned in [this change](https://github.com/pinterest/teletraan/pull/1795).

> Phase 2
> Also triggers Buildkite for the <repo>-private-build if the job exists. This will allow Buildkite to build tarball, the tarball will not be published until we are ready to switch over to Buildkite than Jenkins
> 
> Phase 3
> For both <repo>-hotfix-job and <repo>-private-build, if they exist on Buildkite, then only Buildkite pipelines are triggered, Jenkins' will not be. This phase will last a while until all hotfix pipelines are migrated off Jenkins.

### Test plan

- Rollback scenario
Set priority of `Jenkins` to be `1` and `Buildkite` to be `2`.
Builds are triggering as expected.

- Hotfix pipelines (`*-hotfix-job` & `*-private-build`) half migrated scenario
Each job will be validated if they exist on Buildkite first, if yes the Buildkite pipeline will be triggered, otherwise it will fallback to Jenkins to trigger. Builds are triggering as expected.

- When the Buildkite pipeline enter a state, such as [scheduled](https://buildkite.com/docs/pipelines/configure/defining-steps#job-states), which doesn't exist in Jenkins, the hotfix should not enter a failed state or throw NPE but continue polling for the next round.  

Some **non-pinboard** testing examples https://deploy-integ.pinadmin.com/env/envoy/prod-aws-us-east-1c/hotfix/DqB-1Z8ySveHUHtc8T3AsQ
Screenshot 2025-05-30 at 1 44 51 PM

Some **pinboard** testing examples:

https://deploy-integ.pinadmin.com/env/pinboardjasmine/pindboard-test/hotfix/VIrMsSKzSxWdlEqIGkOC7w
https://deploy-integ.pinadmin.com/env/pinboardjasmine/pindboard-test/hotfix/OrCc5v4jQO6EUWHL0Xd8Vw


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/worker/HotfixStateTransitioner.java
```diff
@@ -53,7 +54,7 @@ public class HotfixStateTransitioner implements Runnable {
     private Counter errorBudgetSuccess;
     private Counter errorBudgetFailure;
     // TODO make this configurable
-    private static final int HOTFIX_JOB_DURATION_TIMEOUT = 180;
+    private static final int HOTFIX_JOB_DURATION_TIMEOUT = 15; // TODO: go back to 180
```
jasmineqin, line 57: TODO: revert back after all comments resolved, this is for testing with less noisy logs in kibana

deploy-service/common/src/main/java/com/pinterest/deployservice/ci/Jenkins.java
```diff
@@ -180,9 +182,14 @@ public boolean jobExist(String pipeline) throws IOException {
...
             String ret = httpClient.get(url, null, null);
             JsonObject json = (JsonObject) JsonParser.parseString(ret);
-            return json != null;
-        } catch (IOException e) {
-            LOG.error("Failed to get job info from jenkins", e);
+            if (json == null || json.isJsonNull()) {
+                return false;
+            } else {
+                return true;
+            }
```
nubster74, line 189:
```suggestion
            return json != null && !json.isJsonNull();
```

nubster74 (commented):
Just to be sure: this being merged won't affect actual production, right? Because we probably shouldn't have BK taking priority until the pipelines are in place.

deploy-service/common/src/main/java/com/pinterest/deployservice/ci/Jenkins.java
jasmineqin: curious when will this final return statement get reached?

gzpcho (dismissed):

rfleur01 (approved):

PR #1837, opened 2025-06-19
add a new auth type, SOX_PROPERTY, for isSOX api to pass through

carlosyang-per:
In [this ER](https://jira.pinadmin.com/browse/ER-11482), we need to give SOX client the permission to access isSOX API, which is currently only accessible by Teletraan's system admin. 
To avoid grand the SOX team to be Teletraan system admins, our solution is add a new auth type, SOX_PROPERTY, and modify pastis acl.

- jira: https://jira.pinadmin.com/browse/CDP-9352
- ER: https://jira.pinadmin.com/browse/ER-11482


deploy-service/teletraanservice/src/main/java/com/pinterest/teletraan/security/TeletraanAuthZResourceExtractorFactory.java
```diff
@@ -57,6 +57,9 @@ public AuthZResourceExtractor create(ResourceAuthZInfo authZInfo) {
                         return hotfixPathExtractor;
                     case HOST:
                         return hostPathExtractor;
+                    case SOX_PROPERTY:
```
tylerwowen, line 60: We can group this with `ENV_STAGE`
```java
case ENV_STAGE:
case SOX_PROPERTY:
  retrun ...
```

tylerwowen (commented):
LGTM

tylerwowen (approved):

PR #1944, opened 2026-05-05
Add indication for stuck cluster replacement

knguyen100000010:
Add indication for stuck cluster replacement. When cluster replacement is not making progress at all (stuck at 0%) for over 30 minutes, a warning is displayed. This is pure front end change with a 30-minutes heuristic. Additional coordination with backend APIs is needed if more precise detection is desired in the future. This approach is recommended to keep things simple.

Screenshot 2026-05-06 at 12 34 27 PM
Screenshot 2026-05-06 at 12 34 52 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,6 @@ pip install pre-commit
 pre-commit install
 ```
 
-
 ### Documentation
 
 [Check out our wiki!](https://github.com/pinterest/teletraan/wiki)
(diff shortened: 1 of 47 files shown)


osoriano (approved):
Agree that the more precise detection could be added in the future if needed. Thanks for the improvements here!

PR #1952, opened 2026-05-29
Update T2 onboarding banner

osoriano:
Update the messaging for the tier2 service migration

### DIFF
--- a/deploy-board/deploy_board/templates/deploys/warning_no_tier3.tmpl
+++ b/deploy-board/deploy_board/templates/deploys/warning_no_tier3.tmpl
@@ -1,8 +1,8 @@
 <div id="noTier3Alert" class="alert alert-danger" role="alert" align="center">
   <a href="#" class="close" data-dismiss="alert">&times;</a>
-  <i><b style="font-size:24px;">Tier 3 new service onboarding to Teletraan is now disabled.</b></i>
+  <i><b style="font-size:24px;">Tier 3 and Tier 2 new service onboarding to Teletraan is now disabled.</b></i>
   <p>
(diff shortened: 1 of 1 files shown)


deploy-board/deploy_board/templates/deploys/warning_no_tier3.tmpl
```diff
@@ -1,21 +1,18 @@
 <div id="noTier3Alert" class="alert alert-danger" role="alert" align="center">
   <a href="#" class="close" data-dismiss="alert">&times;</a>
-  <i><b style="font-size:24px;">Tier 3 new service onboarding to Teletraan is now disabled.</b></i>
+  <i><b style="font-size:24px;">Tier 3 and Tier 2 new service onboarding to Teletraan is now disabled.</b></i>
   <p>
-     Please launch new Tier 3 services in PinCompute. PinCompute onboarding documentation may be found
+     Please launch new Tier 3 and 2 services in PinCompute. PinCompute onboarding documentation may be found
```
carlosyang-per, line 5: nit: May be "Tier 3 and `Tier` 2", to consistent with others?
Also, do we have stateless Tier 3 and Tier 2 services?

carlosyang-per (approved):

deploy-board/deploy_board/templates/deploys/warning_no_tier3.tmpl
osoriano: This text was from the requested ticket. Can you add a comment there so that the requestor is aware of the updates?

osoriano: Planning to merge since user is asking for it already, but we can follow up, ty!

PR #1959, opened 2026-07-10
Derive env useEntitlements from entitlement kill switch + rollout config

prashantbytesyntax:
## Summary

Makes the capacity UI's "entitlement enforced" state (capacity fields greyed out + "View Entitlement" button, driven by `EnvironBean.use_entitlements`) reflect a live two-layer control instead of a static DB column, so the entitlement rollout and its incident kill switch take effect without a redeploy or DB write.

A new `EntitlementEnforcementProvider` (deploy-service `common`) computes the effective state for a cluster (`<env>-<stage>`) from two on-box managed-data inputs, read on each request (fail-safe: any missing/corrupt input → "not enforced", never throws):

- **Layer 1 — kill switch:** the `entitlement_enforcement_teletraan` decider (`/var/config/config.manageddata.admin.decider`) must be at 100.
- **Layer 2 — onboarding/rollout list:** the cluster id must be in the `entitlements/onboarded_teletraan` managed-data list (`/var/config/config.manageddata.entitlements.onboarded_teletraan`), delivered next to the decider.

`EnvStages`/`Environs` call `applyUseEntitlements(...)` on read so the value the deploy-board renders always reflects the current decider + rollout list. Flipping the decider to 0 (or removing a cluster from the list) instantly re-enables manual capacity editing.

The onboarding source uses the classic managed-data list framework (same framework/delivery as the decider), not PinConf, so it reuses Teletraan's existing managed-data integration.

## Test plan

- [ ] Unit: `EntitlementEnforcementProviderTest` — enforced only when decider==100 AND cluster onboarded; fail-safe on missing/corrupt decider/list.
- [ ] Unit: `EnvStagesTest` — `use_entitlements` overridden from the live gate; kill switch off re-enables editing.
- [ ] Manual (staging): onboarded cluster shows greyed capacity + "View Entitlement" with decider at 100; becomes editable when decider flipped to 0.

Jira: (add ticket)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/deploy-service/common/src/main/java/com/pinterest/deployservice/common/EntitlementEnforcementProvider.java
+++ b/deploy-service/common/src/main/java/com/pinterest/deployservice/common/EntitlementEnforcementProvider.java
@@ -1,0 +1,179 @@
+/**
+ * Copyright (c) 2016-2024 Pinterest, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


osoriano (dismissed):

prashantbytesyntax:
Superseded by #1960, which is the same change from a branch on pinterest/teletraan (no fork). Continuing review there.

